PR #119, opened 2013-09-05
Add rabbitmqadmin user/pass

FabioBatSilva:
Hi,

This patch fix exchange definitions, 
Basically it will fail if you remove the guest user (```delete_guest_user : true```)

Cheers


cmurphy:
Closing due to inactivity, and since there is now a .rabbitmqadmin.conf which should solve this issue.

PR #215, opened 2014-07-24
Added ulimit support for RabbitMQ which was restricted to 4096 file descriptors

guilhermeblanco:
Likely none using this library under Debian had more than 4096 connections to RabbitMQ.

By default, this library does not control ulimit settings for RabbitMQ, fallbacking to system default which is 4096 file descriptors. This patch solves this problem under Debian and is well tested in production with over 24k connections successfully. =)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -94,4 +94,5 @@ class rabbitmq::params {
   $environment_variables      = {}
   $config_variables           = {}
   $config_kernel_variables    = {}
+  $ulimit                     = 4096
 }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -83,6 +83,16 @@ class rabbitmq::config {
     notify  => Class['rabbitmq::service'],
   }
 
+  if $::osfamily == 'Debian' {
+    file { '/etc/default/rabbitmq-server':
+      ensure => file,
+      content => template('rabbitmq/default.erb'),
+      mode    => 0644,
+      owner   => '0',
+      group   => '0',
+      notify  => Class['rabbitmq::service'],
+    }
+  }
 
   if $config_cluster {
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -48,6 +48,7 @@ class rabbitmq(
   $version                    = $rabbitmq::params::version,
   $wipe_db_on_cookie_change   = $rabbitmq::params::wipe_db_on_cookie_change,
   $cluster_partition_handling = $rabbitmq::params::cluster_partition_handling,
+  $ulimit                     = $rabbitmq::params::ulimit,
   $environment_variables      = $rabbitmq::params::environment_variables,
   $config_variables           = $rabbitmq::params::config_variables,
   $config_kernel_variables    = $rabbitmq::params::config_kernel_variables,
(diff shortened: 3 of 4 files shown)


deviantony:
+1 for this PR.

manifests/params.pp
```diff
@@ -94,4 +94,5 @@
   $environment_variables      = {}
   $config_variables           = {}
   $config_kernel_variables    = {}
+  $ulimit                     = 4096
```
michaelklishin: RabbitMQ team member here.

The default is `1024` on most distributions and I suspect our docs use it as an example value. `4096`, `16384`, or even `65536` are sensible values since `ulimit -n` is scoped per OS user.

I'd recommend `16384`.

cmurphy: @michaelklishin thanks for lending your expertise!

If 1024 is the default on most distributions then I'd still think that the default ought to be the same here, with the idea that puppet's default behavior should be to not change anything about the system unless told to. Is 1024 a completely insensible default?

michaelklishin: @cmurphy that was a great default 15 years ago. Totally inadequate today. Everybody and their grandma changes it to be 50K+ for production machines hosting data stores, messaging brokers, proxies, etc.

michaelklishin: @cmurphy with this default you can't have more than (oversimplifying) 1K (connections + queues). Should you go above that limit, you either stop accepting new connections (as RabbitMQ does) or fail to write to disk (some tools are not prepared to handle failures like this) or simply shut down (e.g. Cassandra nodes typically shut down when they detect inability to do local disk I/O). It is often inadequate even for development and QA environments: many types of tests result in using a few hundreds of connections and a few thousands of queues.

By bumping it to 16K, you trade an insignificant amount of memory for a much lower chance of system unavailability or developer frustration.

cmurphy:
@guilhermeblanco can you change the parameter name to $file_limit as @igalic suggested and change the default value to 16384 as @michaelklishin suggested? Then rebase it against master to fix the travis error and make it mergeable.

guilhermeblanco:
@cmurphy I don't have too much time for this right now, but I can try during the week.
TBH, it'll be somehow harder to test as I don't even work for the same company and don't use this module anymore.

deviantony:
I've merged this PR with the updated suggestions here: https://github.com/deviantony/puppetlabs-rabbitmq/tree/file-limit

Would you want me to create a new PR?

cmurphy:
@deviantony please do, thanks!

michaelklishin:
By coincidence our team is working on a doc guide on tuning kernel resources and limits. If you use `ulimit` from `rabbitmq-env.conf` or similar, then by default the highest value you can set is `4096` (the default hard limit). To go above that you have to modify `limits.conf` which is a can of worms on Debian and Ubuntu:

 * Upstart [ignores it](http://lzone.de/Debian%20Ubuntu%20ulimit%20Check%20List)
 * Upstart then re-invents [limit controls](http://bryanmarty.com/2012/02/10/setting-nofile-limit-upstart/)

`limits.conf` changes only take effect on login, too. So most likely the users'd have to combine this attribute with another module that manages `limits.conf`.

deviantony:
@cmurphy Done #313

cmurphy:
Merged as #313.

PR #226, opened 2014-08-11
Features/gentoo support

optiz0r:
Adds the tweaks needed to enable this module to be deployed on Gentoo-type distributions. Rabbitmq::Params needs a Gentoo block to avoid the compilation fail, and config files/dirs need to be owned by the package-created rabbitmq user to allow rabbitmq-plugin to make changes.


cmurphy:
I'm not comfortable rebasing this and merging it myself because the question of which user should own /etc/rabbitmq was not addressed. I am closing this for now, but if someone with expertise would like to offer an explanation and perhaps a new PR please do.

PR #227, opened 2014-08-11
Support passwords containing spaces

optiz0r:
A trivial change to quote the password passed through to rabbitmqadmin so that it may contain spaces, which are otherwise passed through verbatim and break the invocation.


jonnytpuppet:
Hi There

I'm closing this PR due to inactivity and failed tests. Please re-submit the PR to address these concerns if interested.

Thanks!

PR #237, opened 2014-09-23
Make the config class slightly more strict variables compliant

solarkennedy:
I run this module under the strict variables mode.

This change makes it work under strict variables mode in the case where the cookie doesn't exist yet.

I tried running the tests under strict variables mode, and 135 of them failed :( So it's not quite ready to be *test* under that mode, but at least it will *run*.


cmurphy:
We no longer use the $::rabbitmq_erlang_cookie fact, so this change is no necessary.

PR #279, opened 2014-12-17
Make rabbitmqctl provider to wait for rabbitmq cluster

bogdando:
W/o this fix, then RabbitMQ put in a cluster and managed
by Pacemaker, rabbitmqctl could fail its requests due
to cluster is not ready yet.

The solution is:
1) Introduce a base Rabbitmq_common provider class with wait_online method
and a retry decorator for rabbitmqctl commands to ensure they will retry
instead of fail then cluster is not ready.
3) Derive rabbitmqctl provider classes from the former one to make
them able to wait for RabbitMQ reasy as well.
4) Update rspecs

Co-authors: Dmitry Ilyin <dilyin@mirantis.com>
Closes-bug: #MODULES-1452

### DIFF
--- a/lib/puppet/provider/rabbitmq_common.rb
+++ b/lib/puppet/provider/rabbitmq_common.rb
@@ -1,0 +1,47 @@
+class Puppet::Provider::Rabbitmq_common < Puppet::Provider
+  initvars
+  commands :rabbitmqctl => 'rabbitmqctl'
+
+  # Wait 'count*step' seconds while RabbitMQ is ready (able to list its users&channels)
+  # Make 'count' retries with 'step' delay between retries.
+  # Limit each query time by 'timeout'
+  def self.wait_for_online(count=30, step=6, timeout=10)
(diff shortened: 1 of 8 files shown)


lib/puppet/provider/rabbitmq_common.rb
```diff
@@ -0,0 +1,44 @@
+class Puppet::Provider::Rabbitmq_common < Puppet::Provider
```
bogdando: Ypu are right, the rabbitmq_common should be reimplemented as a part of Puppet::Provider::Rabbitmqctl.

bogdando:
I will reopen PR once all the fixes will be done

cmurphy:
Feel free to keep it open while you're working on it so that others can see that it's being addressed.

bogdando:
New PR is here https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/292

PR #282, opened 2014-12-18
Replace erlang cookie fact with type and provider

cmurphy:
Alternate proposal to #278


cmurphy:
@lucab http://puppetlabs.com/security/cve/cve-2014-9568

PR #289, opened 2015-01-02
Do not enforce the source value for yum provider

bogdando:
W/o this patch, source for package is always being constructed
as an Internet link in case of RedHat/SUSE.

This is a problem as we may want to use no source in case of
yum provider.

The solution is to not construct the real_package_source for yum
provider and use it as is.

Closes-bug: #MODULES-1631

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -121,7 +121,9 @@ class rabbitmq(
   # $package_source needs to override the constructed value in params.pp
   if $package_source { # $package_source was specified by user so use that one
     $real_package_source = $package_source
-  } else { #  package_source was not specified, so construct it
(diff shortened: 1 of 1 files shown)


cmurphy:
If the provider is yum then the source parameter for the package resource seems to be ignored, as far as I can tell. Is there a specific issue you're running into caused by specifying the source parameter?

bogdando:
Yes, there is a case for OS type, and for RedHat / SUSE the source will be evaluated forcibly and ingoring the package provider. So, the patch is about to fix it

aamerik:
I'm also running this issue, where the package is always fetched via the http link from yum.  This is a welcomed patch!

PR #291, opened 2015-01-03
Added ssl_versions parameter

jtopjian:
This parameter defines which ssl versions to enable in RabbitMQ.
By default, all versions reported by Erlang are enabled. This should
be the same as if this setting was not added in the first place.

### DIFF
--- a/templates/rabbitmq.config.erb
+++ b/templates/rabbitmq.config.erb
@@ -1,6 +1,9 @@
 % This file managed by Puppet
 % Template Path: <%= @module_name %>/templates/rabbitmq.config
 [
+<% if @ssl -%>
+  {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
+<% end -%>
   {rabbit, [
 <% if @ldap_auth -%>
     {auth_backends, [rabbit_auth_backend_internal, rabbit_auth_backend_ldap]},
@@ -21,7 +24,8 @@
                     {certfile,"<%= @ssl_cert %>"},
                     {keyfile,"<%= @ssl_key %>"},
                     {verify,<%= @ssl_verify %>},
-                    {fail_if_no_peer_cert,<%= @ssl_fail_if_no_peer_cert %>}]},
+                    {fail_if_no_peer_cert,<%= @ssl_fail_if_no_peer_cert %>},
+                    {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]}]},
 <%- end -%>
 <% if @config_variables -%>
 <%- @config_variables.keys.sort.each do |key| -%>
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ all features against earlier versions.
 * rabbitmq configuration file.
 * rabbitmq service.
 
-###Beginning with rabbitmq  
+###Beginning with rabbitmq
 
 
 ```puppet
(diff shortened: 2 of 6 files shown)


cmurphy:
According to https://www.rabbitmq.com/ssl.html SSLv3 is disabled as of 3.4.0, so having it on by default is not acceptable. It also notes that TLSv1.0 ought to be disabled as well, though I'm not sure what to do about that.

I would prefer to have the ssl_versions parameter default to undef, then only try to manage that configuration if the user sets it explicitly. That way the default for whatever version they are using will be used.

templates/rabbitmq.config.erb
```diff
@@ -1,6 +1,9 @@
 % This file managed by Puppet
 % Template Path: <%= @module_name %>/templates/rabbitmq.config
 [
+<% if @ssl -%>
+  {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
```
cmurphy: There is already an `if @ssl` section on line 21, this should be within that.

As mentioned in my other comment I think `@ssl_versions` should default to `undef` and have this line wrapped in `if @ssl_versions`. I'm seeking a second opinion now.

jtopjian:
Thank you for your comments.

I agree about sslv3 and originally had it removed by default. I added it back because I noticed that Ubuntu 14.04 contains rabbitmq 3.2 by default. With that in mind, I was trying to be conservative and preserve existing features -- even if they're insecure. I'm happy to remove it, though.

I also agree that tlsv1 should be removed, too, but for some reason, rabbit still accepts tlsv1 connections. I haven't dug into this yet.

With all of that in mind, I agree with you that just having `ssl_versions` as `undef` is the easiest way to go. This way the user explicitly chooses what versions they want, and if they don't pick a version, the config falls back to whatever rabbitmq supports.

Though setting `ssl_versions` to `undef` causes other changes:

First, the validation of `ssl_versions` in `init.pp` will fail because `undef` is not an array. I can think of two things to do here: either set `ssl_versions` to an empty array by default, change the type of `ssl_versions` to something else (such as a comma-separated string), or some other option. What do you think is the best approach?

Next, the template gets a little more complicated. You are correct that there is already an `if @ssl` on line 22, but I think the new conditional on line 4 needs to stay. In addition, it needs to be expanded to check _both_ if `@ssl` is true and if `@ssl_versions` is not `undef`, `[]`, etc.

As well, a nested conditional in the main `ssl_listeners` block is needed, too.

Here's what I've come up with: https://gist.github.com/jtopjian/7423b6bc27709eeb0eef

Note the odd comma, bracket, and brace placements. It's not incorrect... just a little weird.

Now that I'm reviewing all of this, I remembered that the reason I had default values in `ssl_versions` was in order to circumvent the extra template logic. I admit this was a poor decision.

I have no problems making these changes as well as any further changes to implement this correctly (including possibly a conditional in `init.pp` to check to make sure `ssl_versions` is set _only if_ `ssl` is also set). I'd be very grateful for some input on the best way to do this.

Thanks,
Joe

cmurphy:
> First, the validation of ssl_versions in init.pp will fail because undef is not an array

You could wrap the validation in `if $ssl_version`, essentially allowing the value to be `undef` or an array.

> Next, the template gets a little more complicated. 

I think you could still put the new code within the original `if @ssl`, then within that, have a nested `if @ssl_versions` (assuming we go with above suggestion to make the default `undef`) rather than using a separate block with an `and`. Example: https://gist.github.com/cmurphy/76dfe6febee85de5b676

> including possibly a conditional in init.pp to check to make sure ssl_versions is set only if ssl is also set

I don't think this is wholly necessary but it also wouldn't hurt. Whatever you think is best.

jtopjian:
Ah! I like what you did in the template. 

I'll go ahead and rework this and re-submit a new pull request off of the current master.

Thank you for your input.

PR #292, opened 2015-01-05
Make rabbitmq providers to retry the commands

bogdando:
W/o this fix, then RabbitMQ put in a cluster and managed,
for example, by Pacemaker, rabbitmqctl/rabbitmqadmin commands
could fail due to cluster is not ready yet.

The solution is:
1) Add a run_with_retries method to the existing Rabbitmqctl class
   as a retry decorator to ensure the commands will retry instead
   of fail then the cluster is not ready.
2) Derive the providers from the former one to make
   them able to retry all of the list_* commands and, as a result,
   to wait for RabbitMQ ready as well.
3) Update rspecs. Add missing rspec for rabbitmq_plugin

Co-authors:
  Dmitry Ilyin <dilyin@mirantis.com>,
  Colleen Murphy <colleen@puppetlabs.com>

Closes-bug: #MODULES-1452

### DIFF
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -1,4 +1,5 @@
-Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins) do
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'rabbitmqctl'))
+Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, :parent => Puppet::Provider::Rabbitmqctl) do
 
   if Puppet::PUPPETVERSION.to_f < 3
     if Facter.value(:osfamily) == 'RedHat'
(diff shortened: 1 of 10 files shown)


bogdando:
colleen@puppetlabs.com https://github.com/cmurphy

lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
```diff
@@ -1,4 +1,5 @@
-Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins) do
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'rabbitmqctl'))
```
cmurphy: This could just be `require puppet/provider/rabbitmqctl` instead.

cmurphy:
Works for me, I think the `require File.expand_path...` is a little ugly, but it's already in the rabbitmq_policy provider and I've seen issues with rspec-puppet not being able to load the path properly if done the prettier way.

PR #296, opened 2015-01-16
Added ssl_versions parameter

jtopjian:
This commit adds the ssl_versions parameter. This allows users to
choose which versions of SSL that RabbitMQ should accept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,7 +352,9 @@ rabbitmq.config `fail_if_no_peer_cert` setting.
 
 ####`ssl_versions`
 
-Choose which SSL versions to enable. Example: `['tlsv1.2', 'tlsv1.1']`
+Choose which SSL versions to enable. Example: `['tlsv1.2', 'tlsv1.1']`.
+
+Note that it is recommended to disable `sslv3` and `tlsv1` to prevent against POODLE and BEAST attacks. Please see the [RabbitMQ SSL](https://www.rabbitmq.com/ssl.html) documentation for more information.
 
 ####`stomp_port`
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ all features against earlier versions.
 * rabbitmq configuration file.
 * rabbitmq service.
 
-###Beginning with rabbitmq  
+###Beginning with rabbitmq
 
 
 ```puppet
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -350,6 +350,10 @@ rabbitmq.config SSL verify setting.
 
 rabbitmq.config `fail_if_no_peer_cert` setting.
 
+####`ssl_versions`
+
+Choose which SSL versions to enable. Example: `['tlsv1.2', 'tlsv1.1']`
```
cmurphy: Could you also add a link to the rabbitmq SSL guide (https://www.rabbitmq.com/ssl.html) and note the recommendation to turn off SSLv3 and TLSv1.0?

cmurphy:
Looks great! One small docs request.

jtopjian:
I've amended the README -- let me know how it reads.

cmurphy:
Looks great, thanks!

dalees:
Hi, this feature has an issue in that the functionally added is ignored by RabbitMQ (or Erlang?).

Tested with: 
$openssl s_client -connect localhost:5671 -ssl3
$openssl s_client -connect localhost:5671 -tls1
$openssl s_client -connect localhost:5671 -tls1_1
$openssl s_client -connect localhost:5671 -tls1_2
(note the output line "Protocol  :")

The actual result is that tls1, tls1.1 and tls1.2 are still enabled, even when specifying only [ 'tls1.1', 'tls1.2'].

The syntax of the rabbitmq.config file in the change should actually add a line to ssl_options ",{versions, ['tlsv1.1', 'tlsv1.2']}". Compare syntax with https://www.rabbitmq.com/ssl.html#disabling-tls-versions
I suggest it does the same to ssl_opts of the rabbitmq_management block also.

I've also noted that the line "{ssl, [{versions, ['tlsv1.1', 'tlsv1.2']}]}" (which is outside the rabbit config block in the docs) doesn't have any effect on ssl protocol support.
Though this is included in the docs(https://www.rabbitmq.com/ssl.html), and does no harm being there.

Perhaps it is my versions of erlang(1:17.4) or rabbitmq-server (3.4.3-1), can you confirm the above tests work for you with the current patch?

jtopjian:
I have seen the same result regarding tlsv1 still being enabled even though it is not specified in the config. I saw this while I was working on this patch. There's some prior discussion in #291 which was a rough draft of this patch.

I'm not exactly sure how to disable tlsv1 entirely given I've followed everything in the RabbitMQ docs and haven't looked into this further yet. Suggestions are definitely welcome.

dalees:
Thanks for confirming, I'd not read all of #291. I've got this working and disabling tls1.0 with the above noted changes to your patch.

I'll be looking to submit these for review tomorrow (GMT).

dalees:
Created pull request #301 with these changes.

PR #297, opened 2015-01-16
Fix rabbitmq-env.conf parameters

bogdando:
W/o this patch, env parameters from rabbitmq-env.conf
will always override the ones specified at the init file of
rabbitmq-server.

This is an issue as we may want them to differ
for separated rabbitmq instances running at the same host.

The soution is to remove RABBIT(MQ)_ prefixes for all
parameters in rabbitmq-env.conf as it described at the
man pages.

Closes-bug: #MODULES-1673

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ To use RabbitMQ Environment Variables, use the parameters `environment_variables
 class { 'rabbitmq':
   port              => '5672',
   environment_variables   => {
-    'RABBITMQ_NODENAME'     => 'node01',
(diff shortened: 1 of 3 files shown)


hunner:
@bogdando Thanks for this! Could you please update the unit tests for the new variables as well so that they pass? You can view the errors by following the Travis CI link on the PR.

hunner:
Oh, and could you update the README?

bogdando:
Will do

PR #299, opened 2015-01-26
Add retries for rabbitmqadmin curl command

bogdando:
W/o this patch, Rabbitmq::Install::Rabbitmqadmin
sometimes fails to install the rabbitmqadmin script if
there are some connectivity (which is non-transient) errors.

The solution is to add retries for exec command of dependency
staging module as curl allows retries only for transient errors
and fails for connectivity errors. Note, we can do it since the
dependency patch for staging module was merged
https://github.com/nanliu/puppet-staging/pull/52

Related-bug: #MODULES-1650

### DIFF
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -18,6 +18,8 @@ class rabbitmq::install::rabbitmqadmin {
     curl_option => '-k --noproxy localhost --retry 30 --retry-delay 6',
     timeout     => '180',
     wget_option => '--no-proxy',
+    tries       => 30,
+    try_sleep   => 6,
     require     => [
       Class['rabbitmq::service'],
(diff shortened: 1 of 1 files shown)


cmurphy:
I don't understand why the curl_option flags are insufficient? What is the difference, from the module's point of view, between a transient error and a connectivity error? If the connectivity error isn't "transient" then retrying wouldn't help, I would think.

bogdando:
curl and wget can handle transient errors by retries. But if connection refused, it would fail ignoring retry flags. Suggested retries for staging::file will retry exec curl/wget operation itself, so when and if connection alive, it would successfully retry the operation.

cmurphy:
This now requires a different minimum version of the staging module, so I don't think we can merge this until we're ready for a new major release.

hunner:
Closed. See ticket comment.

PR #301, opened 2015-01-29
Bugfix for ssl_versions in rabbitmq.config 

dalees:
RabbitMQ was not respecting this setting. The format of the configuration file was incorrect. This patch updates to bring this in line with documentation.

  The format reference example is given in https://www.rabbitmq.com/ssl.html#disabling-tls-versions
  Also add version list for rabbitmq_management config ssl_opts.

### DIFF
--- a/templates/rabbitmq.config.erb
+++ b/templates/rabbitmq.config.erb
@@ -1,6 +1,9 @@
 % This file managed by Puppet
 % Template Path: <%= @module_name %>/templates/rabbitmq.config
 [
+<%- if @ssl and @ssl_versions -%>
+  {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
+<%- end -%>
   {rabbit, [
 <% if @ldap_auth -%>
     {auth_backends, [rabbit_auth_backend_internal, rabbit_auth_backend_ldap]},
@@ -16,9 +19,6 @@
     {tcp_listeners, []},
 <%- end -%>
 <%- if @ssl -%>
-    <%- if @ssl_versions -%>
-      {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
-    <%- end -%>
     {ssl_listeners, [<%= @ssl_port %>]},
     {ssl_options, [<%- if @ssl_cacert != 'UNSET' -%>{cacertfile,"<%= @ssl_cacert %>"},<%- end -%>
                     {certfile,"<%= @ssl_cert %>"},
@@ -26,7 +26,7 @@
                     {verify,<%= @ssl_verify %>},
                     {fail_if_no_peer_cert,<%= @ssl_fail_if_no_peer_cert %>}
                     <%- if @ssl_versions -%>
-                      ,{ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]}
+                      ,{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}
                     <% end -%>]},
 <%- end -%>
 <% if @config_variables -%>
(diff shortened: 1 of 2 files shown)


templates/rabbitmq.config.erb
```diff
@@ -1,6 +1,9 @@
 % This file managed by Puppet
 % Template Path: <%= @module_name %>/templates/rabbitmq.config
 [
+<%- if @ssl and @ssl_versions -%>
+  {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
+<%- end -%>
```
jtopjian: Does the placement of this block make a difference with the behavior of rabbit? My original pull request had this block placed here as well, but we moved it further down in order to combine conditional statements.

templates/rabbitmq.config.erb
```diff
@@ -16,18 +19,16 @@
...
-                    <%- if @ssl_versions -%>
-                      ,{ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]}
-                    <% end -%>]},
+    {ssl_options, [<%- if @ssl_cacert != 'UNSET' -%>{cacertfile,"<%= @ssl_cacert %>"},
+                   <% end %>{certfile,"<%= @ssl_cert %>"},
+                   {keyfile,"<%= @ssl_key %>"},
+                   {verify,<%= @ssl_verify %>},
+                   {fail_if_no_peer_cert,<%= @ssl_fail_if_no_peer_cert %>}
+                   <%- if @ssl_versions -%>
+                   ,{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}
```
jtopjian: I really have no idea how I missed this multiple times. Thanks for catching it.

templates/rabbitmq.config.erb
dalees: No, because it doesn't actually do anything functional in my tests.
The docs place it outside the rabbit config scope, so that is where it should be unless the docs are wrong.

jtopjian:
Looks like you'll need to update the rspec tests as well. Line 521 of `spec/classes/rabbitmq_spec.rb` should be the only modification needed.

dalees:
Just saw the rspec failures, thanks. Will get these updated as soon as I get these running myself.

templates/rabbitmq.config.erb
jtopjian: I agree with this.

templates/rabbitmq.config.erb
```diff
@@ -16,18 +19,16 @@
...
-    {ssl_options, [<%- if @ssl_cacert != 'UNSET' -%>{cacertfile,"<%= @ssl_cacert %>"},<%- end -%>
-                    {certfile,"<%= @ssl_cert %>"},
-                    {keyfile,"<%= @ssl_key %>"},
-                    {verify,<%= @ssl_verify %>},
-                    {fail_if_no_peer_cert,<%= @ssl_fail_if_no_peer_cert %>}
-                    <%- if @ssl_versions -%>
-                      ,{ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]}
-                    <% end -%>]},
+    {ssl_options, [<%- if @ssl_cacert != 'UNSET' -%>{cacertfile,"<%= @ssl_cacert %>"},
+                   <% end %>{certfile,"<%= @ssl_cert %>"},
```
cmurphy: Why did you move the `<%- end -%>` here?

dalees: Only for prettiness in the output config when no ssl_cacert is specified, though it sacrifices some readability in the template code.
Which should we prefer? I've seen both in this template.

cmurphy: We could keep it where it was but change `<%- end -%>` to `<% end %>`, then the template looks okay and the output is prettier.

dalees: On discussion with my colleague we decided these whitespace changes shouldn't be in this pull request at all. I've got some more changes to come later so I'll add these separately.

cmurphy:
Thanks @dalees and @jtopjian!

PR #303, opened 2015-02-03
Make $rabbitmq::params::package_apt_pin default to undef

raphink:
rabbitmq::repo::apt tests whether this value is true,
which fails with the future parser if default is the
empty string.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -40,7 +40,7 @@ class rabbitmq::params {
   #install
   $admin_enable               = true
   $management_port            = '15672'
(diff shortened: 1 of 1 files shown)


raphink:
@cmurphy seems you missed this one when you were reviewing PRs with @mcanevet

cmurphy:
@raphink sorry about that, thanks!

mhaskel:
@raphink @cmurphy won't this cause all sorts of failures due to #290? Looks like `init.pp` will fail unless `package_apt_pin` is specified due to the `validate_re()`

PR #304, opened 2015-02-04
Fix unscoped template variables

raphink:
LDAP related variables in rabbitmq.config.erb rely
on dynamic scoping, which is discontinued in the future parser.

They need to be exposed in rabbitmq::config in order to be
accessible in the template.


raphink:
This fix is blocking for Puppet 4 migration.

@cmurphy @mhaskel can you check it please?

PR #306, opened 2015-02-06
Do not stop RabbitMQ before creating the Erlang Cookie

EmilienM:
When boostraping a cluster, Puppet will:
* create the cookie and manage permissions
* if there is a change on this resource (create or update), thanks to
  the notify in rabbitmq::config, RabbitMQ service will be restarted by
  Puppet.

Given that, we should not manage the service in the provider itself,
since it's already done in the manifest.

This patch aims to delete the 'puppet resource' command in the
rabbitmq_erlang_cookie provider that ensured RabbitMQ service was
stopped before the cookie file management.

This patch has been tested on a new environment and an existing one, to be sure RabbitMQ is updated with the right cookie, on both use-cases.

### DIFF
--- a/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
+++ b/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
@@ -15,7 +15,6 @@ Puppet::Type.type(:rabbitmq_erlang_cookie).provide(:ruby) do
 
   def content=(value)
     if resource[:force] == :true # Danger!
-      puppet('resource', 'service', resource[:service_name], 'ensure=stopped')
       FileUtils.rm_rf('/var/lib/rabbitmq/mnesia')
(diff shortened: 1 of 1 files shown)


sbadia:
This patch is related to [PDB-1223](https://tickets.puppetlabs.com/browse/PDB-1223)

cmurphy:
If you run the acceptance tests, you can see this causes serious problems with the rabbitmq-server. If the cookie changes while the server is still running, it becomes unable to connect to itself and therefore unable to restart itself.

An alternative to using the puppet resource command would be stopping the service via the init system. This means dealing with all supported init systems and all possible service names, so using the puppet resource command is much simpler.

jonnytpuppet:
Closing due to inactivity and the concerns raised by @cmurphy.

Feel free to open a new PR

Thanks!

PR #308, opened 2015-02-10
Fixed up rabbitmq.conf file layout

daniel-gadd:
Fixed layout of ssl_options section to make sure they are on separate
lines, making it easier to read when the file is cat'ed.


daniel-gadd:
Seems to failing due to rspec - Syntactically its correct and compiles catelogs correctly in puppet

cmurphy:
I think merging #307 caused a conflict here, mind rebasing?

daniel-gadd:
Pushed up rebased commits off #307

PR #309, opened 2015-02-10
Add Support for OpenBSD and fix rabbitmq_exchange/rabbitmqadmin.rb

buzzdeee:
This is a two commits pull request, the majority is in order to add support for OpenBSD.

The other smaller part, that I ran into after getting the module to work, is to fix lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin.rb.

Both explained below.

To support OpenBSD, in the manifests/params.pp, add some more variables, esp. for the rabbitmq_home directory, the rabbitmq system user and group, and additionally make the $plugin_dir OS/Distribution dependent. Then use the new variables throughout the other manifests, where the hardcoded versions were used before.
Further in manifests/init.pp, the dependency of Class['::rabbitmq::install::rabbitmqadmin'] -> Rabbitmq_exchange<| |> is moved up into the if $admin_enable part, since it will throw an error, if rabbitmq admin is not enabled.
In manifests/install/rabbitmqadmin.pp, the group of the rabbitmqadmin file was changed from 'root' -> '0'. 
On *BSD, there is not group root, the equivalent group with GID 0 is called wheel.
Other option would be to have the group name added as an OS dependent parameter.

The problem with the rabbitmq_exchange was, that on successive runs, it always re-created the configured exchanges every time. The parse_command method in the provider was flawed. The provider was restructured, to use rabbitmqctl -q, and with that, the whole parse_command method could and was dropped.
The spec testing for that was updated to reflect the changes.

This development was done on a current OpenBSD 5.7-beta, using rabbitmq-3.4.2.

comments, feedback, or if you like just merge it, welcome ;)

thanks,
Sebastian

### DIFF
--- a/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
+++ b/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
@@ -16,12 +16,12 @@ Puppet::Type.type(:rabbitmq_erlang_cookie).provide(:ruby) do
   def content=(value)
     if resource[:force] == :true # Danger!
       puppet('resource', 'service', resource[:service_name], 'ensure=stopped')
-      FileUtils.rm_rf('/var/lib/rabbitmq/mnesia')
+      FileUtils.rm_rf(resource[:rabbitmq_home] + File::PATH_SEPARATOR + 'mnesia')
       File.open(resource[:path], 'w') do |cookie|
         cookie.chmod(0400)
         cookie.write(value)
       end
-      FileUtils.chown('rabbitmq', 'rabbitmq', resource[:path])
+      FileUtils.chown(resource[:rabbitmq_user], resource[:rabbitmq_group], resource[:path])
     else
       fail("The current erlang cookie needs to change. In order to do this the RabbitMQ database needs to be wiped.  Please set force => true to allow this to happen automatically.")
     end
--- a/README.md
+++ b/README.md
@@ -381,6 +381,18 @@ repository otherwise.
 
 Boolean to determine if we should DESTROY AND DELETE the RabbitMQ database.
 
+####`rabbitmq_user`
+
+String: OS dependent, default defined in param.pp. The system user the rabbitmq daemon runs as.
+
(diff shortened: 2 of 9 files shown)


igalic:
i *like* this change.

lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
```diff
@@ -16,12 +16,12 @@ def exists?
   def content=(value)
     if resource[:force] == :true # Danger!
       puppet('resource', 'service', resource[:service_name], 'ensure=stopped')
-      FileUtils.rm_rf('/var/lib/rabbitmq/mnesia')
+      FileUtils.rm_rf(resource[:rabbitmq_home] + File::PATH_SEPARATOR + 'mnesia')
```
cmurphy: This results in a failure in the acceptance tests:

```
Error: /Stage[main]/Rabbitmq::Config/Rabbitmq_erlang_cookie[/var/lib/rabbitmq/.erlang.cookie]/content: change from UFZMNVXTKIVWOFKSPHHB to TESTCOOKIE failed: Invalid parameter rabbitmq_home(:rabbitmq_home)
```

This is because rabbitmq_home isn't a parameter for the rabbitmq_erlang_cookie type. Same with rabbitmq_user and rabbitmq_group below.

You should be able to add them as parameters in lib/puppet/type/rabbitmq_erlang_cookie.rb.

buzzdeee:
@cmurphy thanks, changes added, as I hope that's what you intended. Travis-CI doesn't run the acceptance tests, so haven't seen them failing.

While there, I added some short notes to README.md, mentioning the rabbitmq_user, rabbitmq_group and rabbitmq_home variables.

Further rebased to your master branch.

cmurphy:
Awesome, thanks!

We unfortunately don't have CI for the acceptance tests, but you can run them locally by following the beaker-rspec docs here: https://github.com/puppetlabs/beaker/wiki/How-to-Write-a-Beaker-Test-for-a-Module

PR #310, opened 2015-02-11
Cannot use default localhost when SSL is enabled.

swissslinky:


cmurphy:
This change provides a way to change the parameter, but it isn't specifically solving the problem "Cannot use default localhost when SSL is enabled." Could you add logic to do the right thing when SSL is enabled? Could you also enhance your commit message by describing in more detail what the problem you're running into is, and how this fixes it?

A couple of prs before this just landed, so this now needs a rebase.

jonnytpuppet:
Closing due to inactivity.

Feel free to open another PR

Thanks!

PR #311, opened 2015-02-18
Fix for incorrect apt repo key value.

lnxusr:
The key 'F7B8CEA6056E8E56' is incorrect which results in the module adding this key on every run. Correct key is: '056E8E56' as taken from http://www.rabbitmq.com/signatures.html

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -6,7 +6,7 @@ class rabbitmq::repo::apt(
   $release     = 'testing',
   $repos       = 'main',
   $include_src = false,
-  $key         = 'F7B8CEA6056E8E56',
+  $key         = '056E8E56',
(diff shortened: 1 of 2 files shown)


cmurphy:
I don't think the key is incorrect. You can see the complete value if you query apt-key:

```
# apt-key adv --list-keys --with-colons --fingerprint | grep -i rabbit
pub:-:1024:17:F7B8CEA6056E8E56:2007-07-06:::-:RabbitMQ Release Signing Key <info@rabbitmq.com>::scESC:
```

This is more likely a bug in the apt module. What version of the apt module are you using? If you don't mind filing a ticket on https://tickets.puppetlabs.com with more details we can try to get this fixed.

lnxusr:
I agree, the key is not incorrect. I have filed a bug with some more details as per request.

Bug report: https://tickets.puppetlabs.com/browse/MODULES-1786

PR #305, opened 2015-02-06
Make LDAP section more configurable

voroniys:
Add a possibility for more flexible configuration of LDAP auth

PR #312, opened 2015-02-23
Make LDAP section more configurable

cmurphy:
Rebase of #305 plus added documentation


mhaskel:
@cmurphy could you update unit tests for the new parameters?

PR #315, opened 2015-03-02
Remove values validation in policy definition

tobyhs:
This fixes https://tickets.puppetlabs.com/browse/MODULES-1807 because something like ha-params should be an integer, not a string.

### DIFF
--- a/spec/acceptance/policy_spec.rb
+++ b/spec/acceptance/policy_spec.rb
@@ -21,14 +21,16 @@ describe 'rabbitmq policy on a vhost:' do
         ensure => present,
       } ->
 
-      rabbitmq_policy { 'ha-all@myhost':
-        pattern    => '.*',
-        priority   => 0,
+      rabbitmq_policy { 'ha-exactly@myhost':
+        pattern    => 'ha_.*',
+        priority   => 1,
         applyto    => 'all',
-        definition => {
-          'ha-mode'      => 'all',
-          'ha-sync-mode' => 'automatic',
-        },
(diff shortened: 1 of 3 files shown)


spec/acceptance/policy_spec.rb
```diff
@@ -21,14 +21,16 @@ class { '::rabbitmq':
...
-        priority   => 0,
+      rabbitmq_policy { 'ha-exactly@myhost':
+        pattern    => 'ha_.*',
+        priority   => 1,
         applyto    => 'all',
-        definition => {
-          'ha-mode'      => 'all',
-          'ha-sync-mode' => 'automatic',
-        },
+        # Workaround for Puppet interpreting numbers as strings
```
hunner: This will only apply to puppet < 4, because as of 4.0.0 integers will not be integers.

Should we instead munge strings of `^\d+$` to integers?

tobyhs: @hunner

I'm unsure about munging certain strings to integers if Puppet 4.0.0 will fix it.  What if RabbitMQ decides to introduce a policy definition where the value can match `^\d+$` but is supposed to be a string?  Then we'll reach a similar issue to what we have now (now, if I specify an integer for ha-params, it fails because of the validation; if I specify a string, then it fails because `rabbitmqctl set_policy` fails.

jonnytpuppet:
Hi There

I'm closing this PR due to inactivity as well as this branch diverging lots from master. Please re-submit your PR if you would still like this functionality included, with the concerns addressed.

Thanks!

PR #316, opened 2015-03-05
Fix file limit for future parser

aldavud:
The validate_re() function takes a string as first argument, so passing in an integer is invalid.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -94,7 +94,7 @@ class rabbitmq(
   validate_re($stomp_port, '\d+')
   validate_bool($wipe_db_on_cookie_change)
   validate_bool($tcp_keepalive)
-  validate_re($file_limit, '\d+')
+  validate_re($file_limit, '^(\d+|-1|unlimited|infinity)$')
   # Validate service parameters.
   validate_re($service_ensure, '^(running|stopped)$')
   validate_bool($service_manage)
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -116,5 +116,5 @@ class rabbitmq::params {
   $environment_variables      = {}
   $config_variables           = {}
   $config_kernel_variables    = {}
-  $file_limit                 = 16384
+  $file_limit                 = '16384'
 }


igalic:
that feels… wrong…
i'm sure that @hlindberg would be (or was) against adding another validate_number() or validate_integer() to puppetlabs-stdlib, but this just feels wrong.

aldavud:
So consequently the correct solutions is to set file_limit to be a string as I'm suggesting in this PR. This is also consistent with how the ports and all other parameter with numeric values are configured.
In the end this is about setting a value in a file, so passing it in as a string containing a numeric value seems correct.

daenney:
Actually, it makes total sense that you can't pass an integer to `validate_re`. You can't use regexes on anything but strings so the behaviour is pretty correct.

The nice way to solve this in Puppet right now is to leave it as an integer and validate it as such:

```puppet
unless is_integer($param) {
  fail('$param must be an integer')
}
```

aldavud:
Again, I think setting this as a string containing a double is the correct thing to do here.

daenney:
A double is a double, not a string containing a double and should be validates as being a double. This isn't (supposed to be) a stringly typed language.

A good argument would be "because any item in limits can also take the values of -1, unlimited or infinity except for priority and nice" as pointed out by `man limits.conf`.

hlindberg:
I agree with @daenney

CyBeRoni:
> This isn't (supposed to be) a strongly typed language.

That is the direction we're going though. https://github.com/puppetlabs/puppet-specifications/blob/master/language/types_values_variables.md

daenney:
No we're not. If anything we're moving away from it. We now actually have a language with a Type system, that respects and enforces the types given. See the Puppet Types section alone.

CyBeRoni:
I see, I misinterpreted 'stringly' as a typo for 'strongly'.

aldavud:
So according to the conversation above I changed the following:

* apart from a positive integer  the values -1, infinity and unlimited should be allowed as valid parameters
* consequently file_limit has to be checked via validate_re and the default can't be an integer

hlindberg:
How is infinity different from unlimited? (We use 'unlimited' for things like environment timeout). If there is no difference between what they do I suggest dropping 'infinity' since all that can be achieved is 'no limit imposed' vs 'to the end of time' (which hopefully will be long after the machine has turned to silicon dust).

aldavud:
From my understanding they are functionally the same. Citing **man 5 limits.conf**

```
All items support the values -1, unlimited or infinity indicating no limit, except for priority and nice.
```

But I think going with exactly what **ulimit** actually supports as the better approach than arbitrarily choosing a subset.

hlindberg:
Yes agree, add both since the thing being modeled has both values.

mhaskel:
closed by #319

PR #317, opened 2015-03-06
Removed apt module dependency

gildub:
Apt is not needed on other systems like Red Hat familly of Oses


jonnytpuppet:
Thanks for your contribution.

However, won't this break support for Debian/Ubuntu OSes (as they do indeed require apt?)

Thanks

cmurphy:
puppetlabs/apt is a hard dependency of this module because of the rabbitmq::repo::apt class. Agreed that RedHat etc doesn't need it, but it doesn't hurt anything to have the module installed. Removing it would break Debian OS's.

PR #318, opened 2015-03-09
Revert "Make $rabbitmq::params::package_apt_pin default to undef"

mhaskel:
Reverts puppetlabs/puppetlabs-rabbitmq#303


mhaskel:
After testing, this commit actually breaks functionality with future parser unless package_apt_pin is specified.

PR #319, opened 2015-03-10
Fixes for file_limit

mhaskel:
There were no docs or tests, and the validation failed if an integer was
passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,11 @@ RabbitMQ Environment Variables in rabbitmq_env.config
 The erlang cookie to use for clustering - must be the same between all nodes.
 This value has no default and must be set explicitly if using clustering.
 
-###`key_content`
+####`file_limit`
+
+Set rabbitmq file ulimit. Defaults to 16384. Only available on systems with `$::osfamily == 'Debian'`
+
+####`key_content`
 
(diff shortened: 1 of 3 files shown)


mhaskel:
Alternate implementation to #316, plus tests and docs

nibalizer:
:+1:

README.md
```diff
@@ -222,7 +222,11 @@ RabbitMQ Environment Variables in rabbitmq_env.config
 The erlang cookie to use for clustering - must be the same between all nodes.
 This value has no default and must be set explicitly if using clustering.
 
-###`key_content`
+####`file_limit`
+
+Set rabbitmq file ulimit. Defaults to 16384. Only available on systems with `$::osfamily == 'Debian'`
```
igalic: why the limitation to debuntu?

mhaskel: Because that's what was implemented? [This](https://github.com/puppetlabs/puppetlabs-rabbitmq/blob/15e2421048a3495e60482837910441a1248bf6f0/manifests/config.pp#L101-L110) is the only place `file_limit` is used. I didn't know enough about rabbitmq on other oses to change that, I just documented and tested what i found.

igalic: *nod* thank you for the clarification!

PR #321, opened 2015-03-10
Fix for fixes for file_limit

aldavud:
The fix was not correct, and was missing test for invalid values.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -101,9 +101,7 @@ class rabbitmq(
   }
   validate_bool($wipe_db_on_cookie_change)
   validate_bool($tcp_keepalive)
-  if ! is_integer($file_limit) {
-    validate_re($file_limit, '^(unlimited|infinity)$', '$file_limit must be an integer, \'unlimited\', or \'infinity\'.')
-  }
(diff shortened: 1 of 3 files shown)


aldavud:
Fixing #319. Great community work on helping me improve my patch...

jonnytpuppet:
Hi There

Thanks for your contribution.

It would be appreciated if you could please provide some details on what was wrong with the original implementation, and how your changes have fixed it.

Thanks!

aldavud:
Hi Jonny

The current solution checks if the number passed in is an integer via the stdlib is_integer function. The issue is that is_integer also accepts integers below zero, for example -42, which in turn is not a valid argument for the limits.conf file. You can also compare this with the test I added.

bmjen:
@aldavud Thanks for the contribution and the explanation. This does seem like a nice validation update to have. Can you please rebase?

aldavud:
@bmjen there you go.

bmjen:
Hi @aldavud, your unit tests for `file_limit` are failing now.

aldavud:
Hi @bmjen thanks for the heads up. I did a far too simple rebase and missed that there were now more contexts for other distros.

All is well now :)

PR #322, opened 2015-03-11
MODULES-1847 - Fixed password verification support

centromere:
The API for authentication has changed in RabbitMQ 3.5.0:

https://github.com/rabbitmq/rabbitmq-server/commit/feb6d85ccb7d18e9762680e69a3b94c0439ebc9e#diff-5a86f1b120c375298765cfc8ff4f8a7cL90

### DIFF
--- a/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
+++ b/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
@@ -45,7 +45,7 @@ Puppet::Type.type(:rabbitmq_user).provide(:rabbitmqctl, :parent => Puppet::Provi
 
 
   def check_password
-    response = rabbitmqctl('eval', 'rabbit_auth_backend_internal:check_user_login(<<"' + resource[:name] + '">>, [{password, <<"' + resource[:password] +'">>}]).')
+    response = rabbitmqctl('eval', 'rabbit_auth_backend_internal:user_login_authentication(<<"' + resource[:name] + '">>, [{password, <<"' + resource[:password] +'">>}]).')
     if response.include? 'invalid credentials'
         false
     else


terrorobe:
I opened a ticket at puppetlabs for this issue: https://tickets.puppetlabs.com/browse/MODULES-1847

cyberious:
Please add test for both cases.

ddeboer:
:+1:

waptang:
Is there a reason to continue to use `rabbit_auth_backend_internal:check_user_login`? It looks like a better option would be to use `rabbit_access_control:check_user_login`, which has been available since v2.3.0, and could open up the possibility of wiring in different authentication plugins.

michaelklishin:
You should use `rabbit_access_control:check_user_login/2`, yes.

nibalizer:
#332 has a partial implementation that uses `rabbit_access_control:check_user_login`

tamaskozak:
I would strongly recommend the standard api route too as @michaelklishin and @waptang recommends. Internals change without notice, APIs tend to be more long living things. 

The patch in #332 seems to be backward compatible as well. 

# dpkg -l|grep rabbit
ii  rabbitmq-server                   3.3.0-1                           AMQP server written in Erlang
# /usr/sbin/rabbitmqctl eval 'rabbit_access_control:check_user_pass_login(list_to_binary("user"), list_to_binary("ooo")).'
{refused,"user '~s' - invalid credentials",[<<"user">>]}
...done.

michaelklishin:
It's not about `rabbit_auth_backend_internal:check_user_login/2` going away. It probably won't. The main reason is that you only use internal authentication backend, which won't work for environments that use LDAP (for example) for authentication.

`rabbit_access_control:check_user_login/2` has been around for more than 4 years.

tamaskozak:
My point is using the api calls advocated by the rabbitmq guys produces a better quality code, the less "hacks" we have to use the better. LDAP and others are an added benefit.

nibalizer:
Neither PR has tests though.

michaelklishin:
LDAP support is a lot more important than nice Puppet module code to quite a few users.

nibalizer:
If a site is using LDAP for user authentication to rabbit, why would they need the rabbitmq_user type?

cmurphy:
It sounds like there is no downside to choosing #332 over this, except for lack of tests, given that `rabbit_access_control:check_user_pass_login` works on both new and old versions of rabbitmq? Is there any argument for using `rabbit_auth_backend_internal` with the version conditional?

tobiashochguertel:
I tested "centromere:patch-1" branch with rabbitmq-server-3.5.0-1 (Ubuntu 14.04) - HTH

tamaskozak:
I would gladly provide tests for #332 but my ruby knowledge is very limited so a helping hand would be appreciated if we need those tests for the future.

cmurphy:
Closing in favor of #332.

PR #323, opened 2015-03-15
MODULES-1856 - Fix to allow bindings and queues to be created on non-def...

jaxim:
MODULES-1856 - Fix to allow bindings and queues to be created
when non-default management port is being used by rabbitmq.

Without this fix the rabbitmq_binding & rabbitmq_queue providers are unable to work when
rabbitmq is running on a non-default management port. This fix fixes the problem by passing in
the /etc/rabbitmq/rabbitmqadmin.conf to the call to rabbitmqadmin so that the port can be read from the conf file.

### DIFF
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -90,6 +90,8 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
       vhost_opt,
       "--user=#{resource[:user]}",
       "--password=#{resource[:password]}",
+      '-c',
+      '/etc/rabbitmq/rabbitmqadmin.conf',
       "source=#{name}",
       "destination=#{destination}",
       "arguments=#{arguments.to_json}",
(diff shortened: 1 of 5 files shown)


jaxim:
Travis build failed on my commit due to using the latest versions of rspec-core (3.2.2) & rspec-puppet (2.0.1). Submitted PR #324 to fix the test issues.

jonnytpuppet:
Hi There

Thanks for submission. Before I can merge this, it would be appreciated if you could provide any information on if you feel this change would cause any other unexpected changes, as this is a pretty wholesale change.

Also, when not including the configuration file in the command, where does it get it's configuration from? Does it just use some built-in default values?

Thanks!

jaxim:
Hi @jonnytpuppet If the config file is not passed to the command then it looks to see if ~/.rabbitmqadmin.conf exists. Below is a partial snippet from the commands help:

```
[vagrant@rabbitmqnode01 ~]$ rabbitmqadmin --help
Usage
=====
  rabbitmqadmin [options] subcommand

Options
=======
--help, -h              show this help message and exit
--config=CONFIG, -c CONFIG
                        configuration file [default: ~/.rabbitmqadmin.conf]
--node=NODE, -N NODE    node described in the configuration file [default:
                        'default' only if configuration file is specified]
--host=HOST, -H HOST    connect to host HOST [default: localhost]
--port=PORT, -P PORT    connect to port PORT [default: 15672]
```

You can also see from this that the default port is 15672 unless passed in. I went with passing in the configuration file instead of the port to keep it consistent with the rabbitmq_exchange provider code as this also passes in the config file.

I added two beaker tests. One using a default management port and one with a non-default and both scenarios created the queues and bindings ok.

mhaskel:
Thanks @jaxim !

PR #324, opened 2015-03-15
MODULES-1857 - Fixing test failures

jaxim:
The Gemfile is now pulling in rspec-core 3.2.2 & rspec-puppet 2.0.1.
This fix has updated the tests to ensure they pass using these versions.

### DIFF
--- a/spec/acceptance/zz281_spec.rb
+++ b/spec/acceptance/zz281_spec.rb
@@ -164,9 +164,9 @@ describe 'rabbitmq class with 2.8.1:' do
     describe file('/var/lib/rabbitmq/rabbitmqadmin') do
       it { should be_file }
     end
-
-    describe command('rabbitmqadmin --help') do
-      it { should return_exit_status 0 }
+    
+    describe command('/usr/local/bin/rabbitmqadmin --help') do
+      its(:exit_status) { should eq 0 }
     end
 
   end
@@ -204,8 +204,8 @@ describe 'rabbitmq class with 2.8.1:' do
       it { should be_file }
     end
 
-    describe command('rabbitmqadmin --help') do
-      it { should return_exit_status 0 }
+    describe command('/usr/local/bin/rabbitmqadmin --help') do
+      its(:exit_status) { should eq 0 }
     end
 
   end
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -6,7 +6,7 @@ describe 'rabbitmq' do
     let(:facts) {{ :osfamily => 'Unsupported' }}
 
     it 'we fail' do
-      expect { subject }.to raise_error(/not supported on an Unsupported/)
+      expect { catalogue }.to raise_error(Puppet::Error, /not supported on an Unsupported/)
     end
   end
 
@@ -163,7 +163,7 @@ describe 'rabbitmq' do
     context 'with file_limit => foo' do
       let(:params) {{ :file_limit => 'foo' }}
       it 'does not compile' do
-        expect{subject}.to raise_error(/\$file_limit must be an integer, 'unlimited', or 'infinity'/)
+        expect { catalogue }.to raise_error(Puppet::Error, /\$file_limit must be an integer, 'unlimited', or 'infinity'/)
       end
     end
   end
@@ -315,7 +315,7 @@ describe 'rabbitmq' do
 
         describe 'with defaults' do
           it 'fails' do
-            expect{subject}.to raise_error(/^You must set the \$erlang_cookie value/)
+            expect { catalogue }.to raise_error(Puppet::Error, /^You must set the \$erlang_cookie value/)
           end
         end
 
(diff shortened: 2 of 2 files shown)


tphoney:
Any particular reason why you changed from expect to should ?
http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/

jaxim:
Hi @tphoney I tried the following:
__expect { subject }.to raise_error(Puppet::Error, /not supported on an Unsupported/)__
but the test still failed:
1) rabbitmq on unsupported distributions we fail
     Failure/Error: expect { subject }.to raise_error(Puppet::Error, /not supported on an Unsupported/)
       expected Puppet::Error with message matching /not supported on an Unsupported/ but nothing was raised
     # ./spec/classes/rabbitmq_spec.rb:9:in `block (3 levels) in <top (required)>'

After reading these this morning:
https://github.com/rodjek/rspec-puppet/issues/257
https://github.com/rodjek/rspec-puppet/issues/173

I have now got the test working with the following syntax:
__expect { should compile }.to raise_error(RSpec::Expectations::ExpectationNotMetError, /not supported on an Unsupported/)__

Is this syntax ok to use? If so I will make the change

Note: I am new to the world of rspec

tphoney:
the expect syntax is preferred. Thanks for the effort you are putting in.

jaxim:
Have committed the tests I amended to use the expect syntax

spec/classes/rabbitmq_spec.rb
```diff
@@ -6,7 +6,7 @@
     let(:facts) {{ :osfamily => 'Unsupported' }}
 
     it 'we fail' do
-      expect { subject }.to raise_error(/not supported on an Unsupported/)
+      expect { is_expected.to compile }.to raise_error(RSpec::Expectations::ExpectationNotMetError, /not supported on an Unsupported/)
```
cmurphy: Could this be changed to just match the regex and not the error type? As in

```
expect { is_expected.to compile }.to raise_error(/not supported on an Unsupported/)
```

That way if they change how the error is detected again this will still work. Checking that the test framework raises an rspec error instead of checking that the catalog raises a puppet error seems strange to me.

cmurphy:
@tphoney if the should syntax still works in rspec 3 and allows us to test for Puppet::Error instead of RSpec::Expectations::ExpectationNotMetError I would have a preference for that. @hunner @mhaskel what do you think?

spec/acceptance/zz281_spec.rb
```diff
@@ -164,9 +164,9 @@ class { 'erlang': epel_enable => true}
     describe file('/var/lib/rabbitmq/rabbitmqadmin') do
       it { should be_file }
     end
-
-    describe command('rabbitmqadmin --help') do
-      it { should return_exit_status 0 }
+    
+    describe command('/usr/local/bin/rabbitmqadmin --help') do
```
cmurphy: Was the path not working properly? This part seemed to work for me. If it's not working it might be indicative of a bug in the module and not the test.

jaxim: Hi @cmurphy When I ran the beaker test prior to the change I made and looked on the centos-64-x64 VM, /usr/local/bin was not on the PATH for the root user.

I had a quick look in the module to see where rabbitmqadmin would be called and if we take the [rabbitmq_exchange provider] (https://github.com/puppetlabs/puppetlabs-rabbitmq/blob/master/lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin.rb) as an example,  this has rabbitmqadmin with the full path:

```ruby
if Puppet::PUPPETVERSION.to_f < 3
    commands :rabbitmqctl   => 'rabbitmqctl'
    commands :rabbitmqadmin => '/usr/local/bin/rabbitmqadmin'
  else
    has_command(:rabbitmqctl, 'rabbitmqctl') do
      environment :HOME => "/tmp"
    end
    has_command(:rabbitmqadmin, '/usr/local/bin/rabbitmqadmin') do
      environment :HOME => "/tmp"
    end
  end
```

In my environment (CentOS 6.5) where I am installing rabbitmq via puppet I made a local change to the rabbitmq_exchange provider code and removed the path from the above code. When I ran puppet it failed and was unable to create my test exchange. Reverting back to including the full path and the puppet run was successful.

spec/classes/rabbitmq_spec.rb
jaxim: Hi @cmurphy tested this morning with matching the regex and not the error type and that was ok. I think this would be neater.

hunner: :+1: because catching `RSpec::Expectations::ExpectationNotMetError` is gross.

cmurphy: Actually, I think you have the right idea below, by using `catalogue` instead of `subject`. The original line with `subject` replaced by `catalogue` seems to work, and it even seems to work for checking the type of error, if we want. So either

```
expect { catalogue }.to raise_error(/not supported on an Unsupported/)
```

or 

```
expect { catalogue }.to raise_error(Puppet::Error, /not supported on an Unsupported/)
```

are optimal solutions, I think.

jaxim: Hi @cmurphy Thanks for the comment. I have committed and went with option two.

cmurphy:
@jaxim awesome. One last thing, could you [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) these three commits into one?

jaxim:
@cmurphy Hopefully I have done the squash correctly. I have learnt something new with that :smile:

cmurphy:
@jaxim thanks for working on this, we're going to follow the same approach for our other modules.

PR #326, opened 2015-03-18
Add new line in ssl options

eimirae:
Rabbitmq does not start the management service properly (the symptom on my system was not reading the ssl-private cert) if this extra carriage return is not in the template, which causes extra spacing for ssl options between ssl_cacert and ssl_cert.


igalic:
@eimirae can you look into whether the travis failures are related to your change?

DavidS:
Resubmitted after rebase in #339 .

PR #327, opened 2015-03-21
set_policy: convert known parameters to integers

awelzel:
When using set_policy to define the 'expires' policy, or setting 'ha-params'
when using 'ha-mode' 'exactly', the rabbitmqctl provider passes a stringified
numeric value to the `rabbitmqctl set_policy' invocation. In turn, this one
chokes on the values and throws the following error:

err: ...
set_policy -p vhost --priority 0 --apply-to queues expire_queues ^rpc_.* {"expires":"1800000"}' returned 2:
Setting policy "expire_queues" for pattern "^rtcp_.*" to "{\"expires\":\"1800000\"}" with priority "0"
...
<<"1800000">> is not a valid queue expiry

This is with puppet 2.7.

I'm not 100% sure why this happens - it seems the a hash is passed all-strings
down to the type/provider in 2.7? Even if in the manifest file, a numeric
value is used. Hum?

 82     rabbitmq_policy { 'expire_queues@vhost':
 83         pattern => '^rpc_.*',
 84         applyto => 'queues',
 85         definition => {
 86             'expires' => 1800000,
 87         }
 88     }

As a work-around, this commit adds two special cases to explicity convert
the 'ha-params' field in case of ha-mode 'exactly' as well as converting
the 'expires' setting when provided.

### DIFF
--- a/lib/puppet/type/rabbitmq_policy.rb
+++ b/lib/puppet/type/rabbitmq_policy.rb
@@ -43,6 +43,9 @@ Puppet::Type.newtype(:rabbitmq_policy) do
     validate do |value|
       resource.validate_definition(value)
     end
+    munge do |value|
+      resource.munge_definition(value)
+    end
   end
 
   newproperty(:priority) do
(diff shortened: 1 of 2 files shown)


awelzel:
This might be related to #315 - not sure what people think.

lib/puppet/type/rabbitmq_policy.rb
```diff
@@ -72,5 +72,21 @@ def validate_definition(definition)
         raise ArgumentError, "Invalid definition"
       end
     end
+    # If ha-mode is 'exactly', convert ha-params to integer
+    if definition['ha-mode'] == 'exactly'
+      ha_params = definition['ha-params']
+      unless ha_params.to_i.to_s == ha_params
+        raise ArgumentError, "Invalid ha-params '#{ha_params}' for ha-mode 'exactly'"
+      end
+      definition['ha-params'] = ha_params.to_i
```
hunner: Values should not be modified from the validate methods. Puppet provides a `munge` method for attributes to perform this kind of conversion.

awelzel:
@hunner , I moved the conversion to munge. Thanks for the feedback!

EmilienM:
I'm facing the same bug and I tested this code, it works for me. Thanks @awelzel !

awelzel:
@hunner , there are some more candidates for policies that will all run into the same issue. E.g. ```max-length```, ```max-length-bytes```, ```message-ttl```...

Would it make sense to add them all for munging?

PR #328, opened 2015-03-23
Added parameters to exchanges

vvincze:
Added capability to use 'durable', 'internal' and 'auto_delete' options with exchanges


nibalizer:
Can you sqash these commits into one or two commits?

vvincze:
Commits have been squashed

PR #329, opened 2015-03-23
Don't create rabbitmqadmin.conf if disabled

dvorak:
If the admin service isn't enabled then we shouldn't create the config
file.

This caused an issue for us because any changes to any resources in rabbitmq::config will cause rabbit to be restarted.  We had intentionally disabled admin when moving to a newer version of the module, because we were trying to avoid restarting rabbit at the time.  Due to this file being created, we had rabbit restarts due to a feature we had explicitly disabled.


nibalizer:
:+1:

PR #332, opened 2015-03-24
Fix check_password function and use the standard rabbitmq api call instead

tamaskozak:
Old one was failing with:

Error: Execution of '/usr/sbin/rabbitmqctl eval rabbit_auth_backend_internal:check_user_login(<<"root">>, [{password, <<"pass">>}]).' returned 2: Error: {undef,
           [{rabbit_auth_backend_internal,check_user_login,
                [<<"root">>,[{password,<<"pass">>}]],
                []},
            {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,569}]},
            {rpc,'-handle_call_call/6-fun-0-',5,
                [{file,"rpc.erl"},{line,205}]}]}

rabbitmq-server                     3.5.0-1

### DIFF
--- a/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
+++ b/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
@@ -45,8 +45,8 @@ Puppet::Type.type(:rabbitmq_user).provide(:rabbitmqctl, :parent => Puppet::Provi
 
 
   def check_password
-    response = rabbitmqctl('eval', 'rabbit_auth_backend_internal:check_user_login(<<"' + resource[:name] + '">>, [{password, <<"' + resource[:password] +'">>}]).')
-    if response.include? 'invalid credentials'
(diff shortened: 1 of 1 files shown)


nibalizer:
This is a duplicate of #322, though the implementation here uses the suggestions found in code review. I like that the implementation in #322 is backwards compatible with older versions of rabbit. Perhaps collaboration is possible?

cmurphy:
I think this is the preferred approach over #322, hoping to get more feedback there. Since tests aren't failing and no new functionality was added I think this PR is complete, just waiting for more community feedback.

errygg:
I am currently using this fix with no issues on Debian 7.

dalees:
I have also tested and am using this fix as it seems cleaner.
It is noted not to support <2.3.0 in the other PR, but module readme states tested on >3.0 so should be no problem.
Wouldn't be hard to add support for <2.3.0, but can't imagine it's required. It would be a regression if anyone is on a 4 year old rabbitmq.

khaefeli:
thank you guys. 
works perfect on debian jessie.

PR #333, opened 2015-03-25
Add support for SSL Cipher lists

dalees:
This patch adds optional array parameter 'ssl_ciphers' which go into rabbitmq's config if SSL is enabled and limit the ciphers to the specified list.

Documentation is erlang as these parameters are passed through: http://www.erlang.org/doc/man/ssl.html
Supported ciphers can be listed with: rabbitmqctl eval 'ssl:cipher_suites().'
Functionality can be tested with something like Cipherscan: https://github.com/jvehent/cipherscan.git

PR #335, opened 2015-04-01
Fixing test failures due to regex matching string start.

dalees:
This will resolve test failures on master so #333 can be merged.


dalees:
@jonnytpuppet  - builds fixed!

PR #336, opened 2015-04-03
Added full fingerprint to be compliant with puppetlabs-apt v1.8

mvisonneau:
This avoids warnings


cyberious:
Can you please squash in one passing commit.

cmurphy:
Can you also bump the minimum apt dependency to 1.8.0 in the metadata.json, as that was then this support was added?

mvisonneau:
There you go !

Zlo:
The long fingerprint should work fine with puppetlabs-apt v1.0, it is just that v1.8 started warning about short fingerprints. The dependency bump is not necessary IMHO.

PR #338, opened 2015-04-15
provider is not a metaparameter

scarby:
provider is not a meta parameter and not defined as a parameter in rabbitmq_plugin type - this causes failed executing when installing plugins defined in init.pp. Seems to function just by removing the provider param. Either need to do this or fix the type.

Adam

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -200,10 +200,9 @@ class rabbitmq(
     include '::rabbitmq::install::rabbitmqadmin'
 
     rabbitmq_plugin { 'rabbitmq_management':
-      ensure   => present,
-      require  => Class['rabbitmq::install'],
-      notify   => Class['rabbitmq::service'],
-      provider => 'rabbitmqplugins'
+      ensure  => present,
+      require => Class['rabbitmq::install'],
+      notify  => Class['rabbitmq::service'],
     }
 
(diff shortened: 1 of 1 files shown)


igalic:
would you very much mind [squashing](http://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits)  those two commits so we don't have failing build in the history (important for git bisect)

scarby:
ok, should be better now

hunner:
:-1: 
@scarby `provider` is not a metaparameter, but that doesn't make it an invalid attribute https://docs.puppetlabs.com/references/latest/type.html#providers. Are you getting an error? What version of puppet are you using? I think there was a bug recently around this.

hunner:
Actually, `rabbitmqplugins` is the only provider, so it should be the default and the provider attribute shouldn't be needed, so :+1:, but still this is probably a red herring anyway.

scarby:
@hunner 
Puppet v3.5.1:

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter provider on Rabbitmq_plugin[rabbitmq_management] at /etc/puppet/environments/rabbitmq/modules/rabbitmq/manifests/init.pp:209 on node dvrbt01.

PR #342, opened 2015-04-22
fix to allow specifying either key_content or package_gpg_key

nnathan:
The example in the README.md is confusing and doesn't work.

1. There is no way to specify $key_content without the apt_key provider complaining (see [here](https://github.com/puppetlabs/puppetlabs-apt/blob/master/lib/puppet/type/apt_key.rb#L22-L29)).

2. Also when specifying the $packaging_gpg_key and setting to a file, you must explicitly set $key_content to undef, otherwise we get stuck in the same issue (this was originally confusing because the example was wrong).

This patch attempts to avoid any confusion and tries to alert the user if there's an invalid combination which will not work with apt_key.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,7 +1,9 @@
 fixtures:
   repositories:
     "stdlib": "git://github.com/puppetlabs/puppetlabs-stdlib.git"
-    "apt": "git://github.com/puppetlabs/puppetlabs-apt.git"
+    "apt": 
+      repo: "git://github.com/puppetlabs/puppetlabs-apt.git"
+      branch: "1.8.x"
     "staging": "git://github.com/nanliu/puppet-staging.git"
     erlang:
       repo: "https://github.com/garethr/garethr-erlang.git"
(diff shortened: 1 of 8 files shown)


bmjen:
Fails lint.

This change also removes any kind of validation for $package_gpg_key for non-Debian operating systems.

bmjen:
@nnathan Can you also please rebase from master to pull in some fixes to CI? Thanks!

nnathan:
I just made the patch to address Debian only since I'm not familiar the behaviour with other OSes.

The functionality for non-Debian OSes should remain unaffected.

bmjen:
@nnathan Can you also squash your two commits?

bmjen:
@nnathan Also, it would have been better to do a rebase instead of a merge of master.

jonnytpuppet:
Hi There

I'm closing this PR due to inactivity as well as the concerns raised by @bmjen 
Please re-submit your PR if you would still like this functionality included with the concern saddressed.

Thanks!

PR #343, opened 2015-04-22
Provided full key fingerprint to suppress warning

centromere:
Puppet gives this warning when installing RabbitMQ:

    Warning: /Apt_key[Add key: F7B8CEA6056E8E56 from Apt::Source rabbitmq]: The id should be a full fingerprint (40 characters), see README.

This patch provides the full fingerprint to prevent this warning from showing.


bmjen:
Duplicate of #336

PR #348, opened 2015-04-28
(MODULES-1987) mnesia dir is not wiped out

korekhov:
change separator from “:” to “/“ to generate proper dirname to allow
removal


lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
```diff
@@ -16,7 +16,7 @@ def exists?
   def content=(value)
     if resource[:force] == :true # Danger!
       puppet('resource', 'service', resource[:service_name], 'ensure=stopped')
-      FileUtils.rm_rf(resource[:rabbitmq_home] + File::PATH_SEPARATOR + 'mnesia')
+      FileUtils.rm_rf(resource[:rabbitmq_home] + File::SEPARATOR + 'mnesia')
```
michaelklishin: `File.join` avoids this confusion entirely.

PR #351, opened 2015-05-12
Ensure idempotency between Puppet runs

EmilienM:
Running RHEL 7.1 with 'yum' or 'rpm' provider, puppetlabs-rabbitmq is not
idempotent.
After a second run, puppet tries to change the SElinux context:
<code>/File[/etc/systemd/system/rabbitmq-server.service.d]/seltype: seltype
changed 'rabbitmq_unit_file_t' to 'systemd_unit_file_t'</code>

Since packaging already manages SElinux labels, we should not let Puppet
doing it (default behavior).
This patch aims to set selinux_ignore_defaults to True for the File
resource (/etc/systemd/system/rabbitmq-server.service.d).

Thanks to that patch, Puppet will be indempotent between all runs on
RHEL platforms.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -120,10 +120,11 @@ class rabbitmq::config {
     'RedHat': {
       if versioncmp($::operatingsystemmajrelease, '7') >= 0 {
         file { '/etc/systemd/system/rabbitmq-server.service.d':
-          ensure => directory,
-          owner  => '0',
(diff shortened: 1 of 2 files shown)


EmilienM:
First patch is to demonstrate idempotency is broken.

EmilienM:
Failed to reproduce. WIP for now.

PR #352, opened 2015-05-15
Allow to manage Erlang Cookie without config_cluster

EmilienM:
When deploying RabbitMQ in cluster with Pacemaker, we don't want to let
Puppet to manage the cluster so we configure it without clustering options.
Though we need to configure the Erlang Cookie in all RabbitMQ nodes.

This patch aims to:
* allow to configure the cookie without config_cluster boolean at True
* keeps backward compatibility, it will try to install the cookie if the
  config_cluster is at True only (so it won't wail for deployment with
defaults values).
* we try to create the cookie only if it's not set to 'undef'.
* update README about new usage
* add rspec tests to test all possible cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,8 @@ RabbitMQ Environment Variables in rabbitmq_env.config
 
 The erlang cookie to use for clustering - must be the same between all nodes.
 This value has no default and must be set explicitly if using clustering.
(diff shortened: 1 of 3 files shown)


jistr:
:+1:

PR #353, opened 2015-05-18
Added mangement_address parameter to rabbitmq_management configuration.

daniel-gadd:
Added management_address parameter to set the 'listener, ip' parameter to
specify the address the rabbitmq_management port should bind to instead
of the default all interfaces port, for example if we want localhost, or a specific address.

Also changed how the rabbitmq.conf template file determines if
the rabbitmq_management should use ssl by adding its own configuration
parameter which defaults to 'false' in case you don't want SSL listening
on our management interface.

Included rspec tests for new changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,10 @@ Password to set for the `default_user` in rabbitmq.config.
 
 Boolean to decide if we should delete the default guest user.
 
+####`enable_management_ssl`
+
+Boolean to decide if the rabbitmq_management plugin should use ssl. Defaults to `false`.
+
 ####`env_config`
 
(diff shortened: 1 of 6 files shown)


daniel-gadd:
This change could have adverse effects of users who rely on SSL for their management interface (based on current `master` version. I'll update my code to provide better logic around if SSL is enabled whether or not to follow that logic (like the current `master` version).

jonnytpuppet:
Hi There

It would be appreciated if you could re-open another PR with just the additions for management_address, as we have another PR for management ssl enable/disable.

Thanks!

PR #354, opened 2015-05-21
Updating APT source to use new apt module version

bleuchtang:


nibalizer:
I don't have a strong opinion about squashing these down. 
What /is/ the plan for rolling this incompatible change out? A long-lived branch? A user wont be able to consume the new rabbitmq module until everything in their environment is ready for new-apt right?

EmilienM:
We might want a 5.3.0 before this change. It will break all Puppet OpenStack modules otherwise, our Beaker tests use so clone master now to have a patch between 5.2.0 and master.

So please release 5.3.0, and then think about merging this kind of pull-request.

EmilienM:
a release has been done, this patch can be reviewed.

bleuchtang:
@EmilienM: Should I do something to speed the review ?

metadata.json
```diff
@@ -49,7 +49,7 @@
   ],
   "dependencies": [
     {"name":"puppetlabs/stdlib","version_requirement":">=3.0.0 <5.0.0"},
-    {"name":"puppetlabs/apt","version_requirement":">=1.8.0 <2.0.0"},
+    {"name":"puppetlabs/apt","version_requirement":">=1.8.0 <2.0.2"},
```
LeslieCarr: For this case, the version required is 2.0.0 to 2.0.2 (since the apt module syntax changed from 1.8.0)

bleuchtang: I rebased from master and take care of your comment, thanks!

bmjen:
Hi @bleuchtang Thanks for the contribution. However we've recently merged in #366 which loosens the apt dependency to use the newly released puppetlabs-apt version 2.1.0. This new version of apt should be backwards compatible.

Thanks!

PR #358, opened 2015-05-27
create resources with hiera

bleuchtang:
Add the availability to create resources directly with hiera


jonnytpuppet:
Hi There

I'm closing this as we do not like the pattern where we call heira by default. Perhaps if you're interested, you could create a separate class for this and re-submit a PR?

Thanks

ThomasLohner:
A separate class only for hiera support would make this module more complicated to use. What's the problem with this approach? create_ressources won't do anything if you don't set the params in hiera  Please explain what's the harm? I see many modules using hiera exactly like this.

PR #360, opened 2015-05-29
Fix acceptance tests

cmurphy:
Found three distinct issues causing acceptance tests to fail on Ubuntu, fixes are captured in this PR.


cmurphy:
> The result of the test was: PASS

\o/

PR #371, opened 2015-06-30
Fix "Invalid parameter include_src on Apt::Source[rabbitmq]"

Pryz:
Using the latest version of puppetlabs-apt and the latest of puppetlabs-rabbitmq I get the following : 

Invalid parameter include_src on Apt::Source[rabbitmq] at /etc/puppet/environments/testing/modules/rabbitmq/manifests/repo/apt.pp:32 on node xxxxxx

This PR fix the error.

See : https://tickets.puppetlabs.com/browse/MODULES-2177

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -21,14 +21,18 @@ class rabbitmq::repo::apt(
   }
 
   apt::source { 'rabbitmq':
-    ensure      => $ensure_source,
-    location    => $location,
-    release     => $release,
-    repos       => $repos,
-    include_src => $include_src,
-    key         => $key,
-    key_source  => $key_source,
-    key_content => $key_content,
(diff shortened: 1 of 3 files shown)


DavidS:
Thank you for your contribution. 

After this change, metadata.json should require apt >=2.0.0, since it won't work with 1.8 then.

Please squash all these commits into one and update the commit title to include "(MODULES-2177)" at the start?

Pryz:
Done :)

cmurphy:
include_src is not an invalid parameter in puppetlabs-apt 2.1.0: https://github.com/puppetlabs/puppetlabs-apt/blob/master/manifests/source.pp#L14

This patch will require a major version bump which is a big inconvenience. I request that this be reverted since the parameter is in fact valid.

PR #373, opened 2015-07-10
Fixing apt repo configuration, key should be key ID not key fingerprint

vorlock:
when used with https://github.com/puppetlabs/puppetlabs-apt


cmurphy:
:-1:

This would revert https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/336

Perhaps you could report the specific issue you are seeing at tickets.puppetlabs.com? It might be an issue with the apt module, not this module.

PR #375, opened 2015-07-28
Rabbitmq bindings

vvincze:
This modification add capability to the RabbitMQ module, to be able to define multiple routing keys for the same binding.


das-vinculum:
Hi, we use the puppet module and it solves all our problems, except the multiple routing keys. 
Is there a reason why this is not merged?

Thank you for the good work.

vvincze:
I'm still waiting for somebody to merge it. I guess it's a big change in the syntax, that's why it's not merged yet, but I'm not sure. We are using the module with this modification for a while and it's working as designed.

liwo:
I currently need this feature, too. 

As a guess to why this is not yet merged, maybe because it is not backward compatible and once they merge it every user of this module has to update his puppet code. But you could easily make it backward compatible by making the new syntax optional, e.g. count the number of @-signs in the name and if there are 3, extract the routing key from the name, otherwise use the routing_key parameter. Probably you would have to do some additional validation then. 
Maybe that would improve chances of this being merged. But that's just a guess.

bmjen:
@vvincze Thanks for the contribution. This change is definitely backwards incompatible and for that reason I won't be able to merge it as is.

However, given the description of your dilemma, I believe an adequate solution would be to implement a composite namevar for `rabbitmq_binding` type. This will keep the existing parameters intact and allow for the multiple bindings as desired.

You can find an example of implementing composite namevar here: https://github.com/puppetlabs/puppetlabs-java_ks/blob/master/lib/puppet/type/java_ks.rb#L150-L176

At the top of the file, there will be documentation explaining the need.

Thanks!
Bryan

wyardley:
I had been working on a fix for the same issue using composite namevars (which would change the interface even more, while maintaining backwards compatibility), but haven't had any luck getting it to work with prefetching yet.... would really appreciate help feedback! 
https://github.com/puppetlabs/puppetlabs-rabbitmq/compare/master...wyardley:feature-modules-3679-take2?expand=1

NickMRamirez:
This would be useful for us too. It seemed like the only thing stopping me from adding multiple `rabbitmq_binding` instances was that the title of the resource must have the form:

```
rabbitmq_binding { "${exchange_name}@${queue_name}@${vhost_name}" :
}
```

This form, which does not specify the routing key, causes "Duplicate resource" errors if I try to add more than one. Maybe the fix is just to allow adding the routing key to this title?

wyardley:
@NickMRamirez This PR exactly is proposing adding the routing key to the title. Personally, I think moving away from putting all the information in the title is the correct approach (and this approach has the downside of being completely backwards-incompatible), but thus far, I haven't had the energy or knowledge to rework my solution in #504, though I did get some really helpful guidance from @bmjen, @hunner, and others

wyardley:
I have updated #504 in a way that I think should provide this functionality.

wyardley:
Closed now that #504 is merged.

PR #377, opened 2015-07-31
MODULES-2309: Add option to not notify service on config change

jmswick:
This is to add the option to not notify the rabbitmq-service to restart after a configuration file change.

This is really only useful for HA rabbitmq clusters were you want to ensure that a controlled restart is done of all node on configuration changes.  Otherwise there is the chance of clustering issues if all nodes in the rabbitmq cluster are restarted simultaneously, as the current rabbitmq clustering configuration is highly dependent on the start order for node recovery.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,19 @@ class { 'rabbitmq':
 }
 ```
 
+Setting up a cluster that comes up on its own, but isn't restarted automatically
+on configuration changes, so that a controlled restart can be done:
+
+```puppet
+class { 'rabbitmq':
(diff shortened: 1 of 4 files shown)


hunner:
The usual way to accomplish this is to pass `service_manage => false` as this is a much-less fragile implementation that covers your use case.

jmswick:
Perfectly understandable.  However, the one drawback to 'service_manage => false' is that puppet won't start or enable the service with out other manual intervention or changing the configuration after initial setup.

PR #379, opened 2015-08-11
add --tlsv1.1 flag to curl download of rabbitmqadmin so that the http…

jpackagejasonc:
…s server doesn't reject the request with newer rabbitmq-server


hunner:
Does the rabbitmq server not auto-negotiate the security protocol? Why this flag specifically? What happens on older versions?

bmjen:
Thanks for the contribution @jpackagejasonc! Are you able to address the questions that @hunner asked?

jpackagejasonc:
I can't reproduce the failure anymore with our current setup so I think this is ok to close without merging. thanks for looking!

bmjen:
Thanks for the update @jpackagejasonc!

PR #380, opened 2015-08-17
Added heartbeat option to rabbitmq-server

daniel-gadd:
Added the rabbitmq config option `heartbeat` to provide a heartbeat
connection to nodes. By default this option isn't set in rabbitmq.conf
and its package default is 580 seconds, if the option is set, the config
will reflect this change. There are no other config changes required.

This change was made as we use the particular option in our implementation
and it wasn't in the module.

Spec tests provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,11 @@ set config_cluster to 'False' and set 'erlang_cookie'.
 Set rabbitmq file ulimit. Defaults to 16384. Only available on systems with
 `$::osfamily == 'Debian'` or `$::osfamily == 'RedHat'`.
 
+####`heartbeat`
+
+Set the heartbeat timeout interval, default is unset which uses the builtin server
+defaultsof 60 seconds. Setting this to `0` will disable heartbeats.
+
 ####`key_content`
 
 Uses content method for Debian OS family. Should be a template for apt::source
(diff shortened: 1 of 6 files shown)


michaelklishin:
This does not "provide a heartbeat connection to nodes". This is the default that's used when a client doesn't specify a different heartbeat value.

For a similar feature for inter-node communication, see [net_ticktime](http://www.rabbitmq.com/nettick.html).

daniel-gadd:
Sorry - Your're right.. Guess my mind was elsewhere when I wrote my commit message

daniel-gadd:
Any update on getting this pushed?

daniel-gadd:
I have rebased my commit with the current master branch.

hunner:
Looks good! Could you add an entry to the readme describing the new parameter, some validation, and use a default value of `undef` instead of `'UNSET'`? I know the module has unset all over the place, but that's an old/ugly pattern.

hunner:
Just checking in with you @daniel-gadd for the above-mentioned changes. Thanks!

daniel-gadd:
Sorry been away from computer for last few weeks, will make changes this week and update the PR. thx

HelenCampbell:
Closing as rebased and merged in PR #437 . Thank you!

PR #381, opened 2015-08-19
Install file limits snippet to limits.d on EL7

jeckersb:
It is possible for the RabbitMQ service to be managed by something
other than systemd on EL7.  An example is when pacemaker is managing
RabbitMQ in an HA cluster.  In this use case, the systemd limits will
have no effect, but it is still desirable to set the file descriptor
limit.


jeckersb:
Oops, let me update the tests as well.  Stay tuned.

EmilienM:
I confirm we need to use template('rabbitmq/limits.conf') in RHEL7.

PR #382, opened 2015-08-27
Prefix apt key with 0x makes puppetlabs-apt happy

matthewfischer:
puppetlabs-apt is looking for a leading 0x on keys.


hunner:
What version of apt are you using? What error is raised?

matthewfischer:
Hunner, I don't see the warning that I saw before, so let's just close this.

PR #385, opened 2015-09-03
Optional correction for default stomp ports

acowan:
The RabbitMQ Stomp Plugin documents the default listening port
to be 61613. With the SSL equivalent being 61614. We should
probably be following the de facto standard.


bmjen:
:+1:  Good catch @acowan !

PR #386, opened 2015-09-10
Add parameter to configure only the SSL port for STOMP plugin

dgmorales:
This PR adds a parameter to configure only the SSL port for STOMP plugin.

I've added related parameter doc in README.md. I didn't add tests because there are very few tests about stomp and none about ssl_stomp_port.


bmjen:
Thanks for the contribution @dgmorales!

PR #388, opened 2015-09-14
Add support to uninstall rabbitmq plugins.

acowan:
The module does not currently support removal of plugins once they have been installed. While the system will remove any added configuration, this merely means that the plugin will run in its default configuration. In order to disable the plugin properly, we need to support removal.

### DIFF
--- a/lib/puppet/parser/functions/convert_to_ensurable.rb
+++ b/lib/puppet/parser/functions/convert_to_ensurable.rb
@@ -1,0 +1,18 @@
+#
+# convert_to_ensurable.rb
+#
+# Simple converter function to map boolean values into their 'ensurable' counterparts, :present & :absent
+#
+module Puppet::Parser::Functions
+  newfunction(:convert_to_ensurable, :type => :rvalue, :doc => <<-EOS
+Simple converter function to map a boolean value into its 'ensurable' counterpart
+    EOS
+  ) do |arguments|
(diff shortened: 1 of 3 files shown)


lib/puppet/parser/functions/convert_to_ensurable.rb
```diff
@@ -0,0 +1,18 @@
+#
+# convert_to_ensurable.rb
+#
+# Simple converter function to map boolean values into their 'ensurable' counterparts, :present & :absent
+#
+module Puppet::Parser::Functions
```
EmilienM: I would rather see this kind of function in stdlib so anyone can reuse it and not put functions like this in puppetlabs-rabbitmq.

acowan: So would I and there are many variations of them about.  In fact there is a stdlib function 'bool2str' that would suffice to provide the necessary functionality (though somewhat more hacky)

But doing this would require us to drop support for stdlib < 4.10.0  This might impact end users more than adding the function directly, so I am at a loss for how to proceed.

hunner: I think I'd be fine with bumping stdlib dependency to 4.10.0. stdlib's code is pretty stable, so updating that module is a low-risk operation.

acowan:
I am closing this as I have opened an updated pull request #455

PR #391, opened 2015-09-29
Update init.pp

lveljovic:
New heartbeat variable.


HelenCampbell:
Ref #394 for closure reason

PR #393, opened 2015-09-29
Update params.pp

lveljovic:
New heartbeat variable.


HelenCampbell:
Ref #394 for closure reason

PR #394, opened 2015-09-29
Update rabbitmq.config.erb

lveljovic:
New heartbeat variable.


HelenCampbell:
Hey @lveljovic ! First thank you for all your work to add this. However we like to keep similar work all contained within the one commit within a single PR. And having said that there was another currently open PR with all the changes required for this new variable (including tests and Readme updates which we would require for a new param also). I've done some work with the other PR and plan on merging it when Travis finishes running ( #437 ). I'll close off your open PRs for this, but thank you for all your effort.

PR #400, opened 2015-10-07
Only validate $file_limit against the regexp when it is not an int

gfidente:


jistr:
Wondering if `$file_limit` -> `"$file_limit"` would do the trick too?

gfidente:
hi, judging from the ci failures I think that would also help making sure it gets validated when it's < 0, updating, thanks!

gfidente:
Closing in favour of pull request #401

PR #401, opened 2015-10-08
Allow integer values for $file_limit

jistr:
(This is another try at implementing #400 . AFAIK github doesn't allow pull request amendment by a different person so i'm sending a new pull request.)

Previously errors were thrown when $file_limit was integer, as
validate_re() function only works with strings. $file_limit
is now casted to string before passing it to validate_re().


gfidente:
I think it's fine, thanks!

jistr:
This is needed for https://review.openstack.org/#/c/232983/

PR #402, opened 2015-10-09
Update ruby.rb

gwagner:
Fixes RVM implementations.  When the path is hard set, for some reason that is beyond me, and a the path gets overwritten, then puppet and ruby can not be found.  W/O this, RVM works again


gwagner:
Anyone looking at this?  I would like to drop my fork to remove one line of bad practice from the code!

tphoney:
no longer relevant, or needs re-implemented

PR #403, opened 2015-10-13
Avoid starting rabbitmq processes as root

mattymo:
Rabbitmq-plugins must be run as root
(in order to set /etc/rabbitmq/enabled_plugins), but as
a consequence, it starts epmd and leaves it running.
This can be worked around by starting epmd as rabbitmq
user before evaluating any plugins.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -201,6 +201,15 @@ class rabbitmq(
     }
   }
 
+  # Start epmd as rabbitmq so it doesn't run as root when installing plugins 
+  exec { 'epmd_daemon':
+    command => 'epmd -daemon',
+    path    => '/bin:/sbin:/usr/bin:/usr/sbin',
+    user    => $rabbitmq_user,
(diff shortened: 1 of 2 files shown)


michaelklishin:
Not sure if this makes things any easier but it should be possible to simply run any `rabbitmq*` command line tool with the effective user you want and it will start epmd for you (well, the runtime will).

mattymo:
epmd will start if it's not running already if you run any rabbitmq* script. If you're running as root (which is required for rabbitmq-plugin script), it launches epmd. In many deployment scenarios, we need to enable plugins before starting any clustered mode.

daenney:
What's the problem with running `epmd` as root? I see no explanation for this.

bmjen:
Agree with @daenney , also don't like that it's going to be the default for all users of the rabbitmq module.

eputnam:
Closing this because of inactivity. @mattymo if this is still an issue, please feel free to re-open and rebase!

PR #405, opened 2015-10-15
Update install.pp

mnothic:
add support to none default mnesia base

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -19,5 +19,13 @@ class rabbitmq::install {
       source  => $package_source,
     }
   }
-
+  if $rabbitmq::environment_variables['MNESIA_BASE'] {
+    file { $rabbitmq::environment_variables['MNESIA_BASE']:
+      ensure =>  "directory",
+      owner => "root",
+      group => "rabbitmq",
+      mode => 775,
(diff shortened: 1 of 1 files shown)


nibalizer:
@mnothic can you squash these commits into one? Can you at the same time write a more complete commit message explaining what the change is and why you are doing it?

As for test failures, this is a simple linting problem. Double vs single quotes, should be easy to fix.

Thanks

mnothic:
the module support mnesia environment variable set but when you setup it with another path which isn't the default the other task can't find the new directory and every else fails ok so the fix is for ensure the directory mnesia if is passed in the environment_variables hash

bmjen:
Tests failing linting.

~~~
Skipping metadata validation; the metadata-json-lint gem was not found
manifests/install.pp - WARNING: unquoted file mode on line 28
The command "bundle exec rake validate && bundle exec rake lint && bundle exec rake spec SPEC_OPTS='--format documentation'" exited with 1.
~~~

bmjen:
@mnothic Have you had a chance to address the issues on this PR?

nibalizer:
I'm ok with the code/feature but need the cleanups asked for.

PR #409, opened 2015-11-04
Change cluster_node_type description in Readme.md - MODULES-2786

dalesit:
Change option description from disk to disc, as the actual parameter value needs to be "disc" or "ram"


HelenCampbell:
Thanks!

PR #412, opened 2015-11-11
MODULES-2252 - fix "Could not prefetch rabbitmq_exchange provider 'rabbitmqadmin': 757: unexpected token at 'fanout'" issue

madAndroid:
Skips federated exchanges in the output, as these have a different set of fields, and cause issues when they're parsed.


bmjen:
Thanks @madAndroid !

PR #413, opened 2015-11-12
Fix incompatability with latest puppet

quixoten:
This fixes an issue with the latest version of puppet when run from a cron job, which does not include "/opt/puppetlabs/bin" in the PATH that the cron jobs see. The other issue with the current implementation is that, in the latest puppet code at least, the custom_environment that's being set in the block given to `has_command` is not used when resolving the location of the command.

To reproduce the issue, run `PATH=/usr/bin:/bin sudo /opt/puppetlabs/bin/puppet agent --test`


raphink:
:+1: this bug makes the provider fail currently!

PR #415, opened 2015-12-02
Use yum as package provider for RedHat

stieizc:
I noticed that with `epel` enabled, `yum` can resolve the erlang dependencies of rabbitmq without the help of additional puppet erlang modules on CentOs7. Since `epel` is needed on both CentOS and RedHat, maybe its safe to use `yum` as the package provider for them instead of `rpm`.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -40,7 +40,7 @@ class rabbitmq::params {
       $package_ensure   = 'installed'
       $package_name     = 'rabbitmq-server'
       $service_name     = 'rabbitmq-server'
-      $package_provider = 'rpm'
+      $package_provider = 'yum'
       $version          = '3.1.5-1'
       $rabbitmq_user    = 'rabbitmq'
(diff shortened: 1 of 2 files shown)


bmjen:
@stieizc Spec tests need to be updated to reflect this change.
https://github.com/puppetlabs/puppetlabs-rabbitmq/blob/master/spec/classes/rabbitmq_spec.rb#L1120

stieizc:
Update the spec, and test fails for Puppet 4. Any idea? I will look into the build log when in a few days.

bmjen:
@stieizc Do you mind rebasing off master to pull in my test fixes from #416? That should fix unit tests.

stieizc:
I'm Glad to. And now all checks have passed.

bmjen:
@stieizc Sorry for not requesting this last time, but can you squash the two commits down to 1?

stieizc:
Sure.

PR #421, opened 2015-12-22
This breaks resources { 'rabbitmq_user': purge => true }

hashbrowncipher:


lib/puppet/type/rabbitmq_user.rb
```diff
@@ -76,10 +76,4 @@ def should_to_s(newvalue = @should)
 
   end
 
-  validate do
-    if self[:ensure] == :present and ! self[:password]
```
hunner: Removing this validation is not a valid solution to fixing self.instances. You can check both `self[:password].nil?` and `self.provider.password.nil?` eg: https://github.com/puppetlabs/puppetlabs-netapp/blob/master/lib/puppet/type/netapp_vserver.rb#L162

HelenCampbell:
@hashbrowncipher Thank you for your contribution, however I'm going to close this PR due to inactivity. Please feel free to open another PR with your work in it and we will gladly review. Thanks again!

PR #422, opened 2016-01-02
Ensure an apt-update after change repository so we can get upstream version

gpenaud:
I needed 3.6.0 version for rabbitmq, which was handled by custom debian repo, but no apt-update was done after. I was then stuck at 3.3.5.


jecnua:
+1

PR #423, opened 2016-01-06
Update apt::source to match with new method

EmilienM:
puppetlabs-apt recently did some changes that brings warnings when we
run puppetlabs-rabbitmq, because we need to update the way we consume
apt module.

$include_src is deprecated and will be removed in the next major
release, please use $include => { 'src' => false } instead

$key_source is deprecated and will be removed in the next major release,
please use $key => { 'source' =>
http://www.rabbitmq.com/rabbitmq-signing-key-public.asc } instead.

This patch aims to update the way we use puppetlabs-apt to match with
upstream and avoid warnings in our catalog.

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -29,10 +29,14 @@ class rabbitmq::repo::apt(
     location     => $location,
     release      => $release,
     repos        => $repos,
-    include_src  => $include_src,
-    key          => $key,
-    key_source   => $key_source,
-    key_content  => $key_content,
+    include      => {
+      'src' => $include_src
+    },
+    key          => {
+      'id'      => $key,
+      'source'  => $key_source,
+      'content' => $key_content,
+    },
     architecture => $architecture,
   }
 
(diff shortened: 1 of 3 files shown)


hunner:
Does this require a bump in metadata.json, or will it still work with apt 1.8.0?

EmilienM:
@hunner yes, but 2.x.x has been released in 2015-04. I suppose people update their manifests from time to time? Maybe I'm wrong. Feel free to merge it or not.

FWIW, Puppet OpenStack CI is running 2.2.x serie:
https://github.com/openstack/puppet-openstack-integration/blob/master/Puppetfile#L117-L119

hunner:
I want to merge this, but it would require a 6.0.0 release of rabbitmq. While we're at it, what other changes do we want to make to the module?

hunner:
Could you also increase the lower bound of puppetlabs-apt in metadata.json to >= 2.0.0? Since it won't work with 1.8.x

EmilienM:
@hunner done

dhoppe:
:+1:

eputnam:
hi @EmilienM , could you please rebase?

EmilienM:
@eputnam done.

hunner:
Merged in #497

PR #424, opened 2016-01-07
Module2940 addition auth backends variable

firefly4268:
(#MODULE-2040) -Add auth_backends Variable

The current implementation of auth_backends forces configuration of auth_backends to either the RabbitMQ default or a hardcoded configuration for LDAP. This variable allows configuration to be controlled as desired, particularly, the case of interest to myself: it will allow for a configuration where LDAP is used to authenticate and the RabbitMQ mnesia database is used for authorizations. The change is backwards compatible and has a test.

### DIFF
--- a/templates/rabbitmq.config.erb
+++ b/templates/rabbitmq.config.erb
@@ -5,7 +5,9 @@
   {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
 <%- end -%>
   {rabbit, [
-<% if @ldap_auth -%>
+<% if @auth_backends -%>
+    {auth_backends, [<%= @auth_backends.map { |v| "#{v}" }.join(', ') %>]},
+<% elsif @ldap_auth -%>
     {auth_backends, [rabbit_auth_backend_internal, rabbit_auth_backend_ldap]},
 <% end -%>
 <% if @config_cluster -%>
--- a/README.md
+++ b/README.md
@@ -149,6 +149,13 @@ class { 'rabbitmq':
 
 Boolean, if enabled sets up the management interface/plugin for RabbitMQ.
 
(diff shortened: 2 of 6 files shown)


templates/rabbitmq.config.erb
```diff
@@ -5,9 +5,12 @@
   {ssl, [{versions, [<%= @ssl_versions.sort.map { |v| "'#{v}'" }.join(', ') %>]}]},
 <%- end -%>
   {rabbit, [
-<% if @ldap_auth -%>
+<% if @ldap_auth and (@auth_backends == 'UNSET') -%>
     {auth_backends, [rabbit_auth_backend_internal, rabbit_auth_backend_ldap]},
 <% end -%>
+<% if @auth_backends != 'UNSET' -%>
+    {auth_backends, <%= @auth_backends %>},
```
DavidS: I'd prefer that to be handled in the same way as `ssl_versions` (See line 5 above). Tthe user passes in an array and the template expands this into a properly formatted configuration entry.

DavidS:
Nice one! I had a minor comment on the API design, but it looks solid to me otherwise. Please make sure that your final submission is a single squashed commit, so that the history stays clean.

See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits for how to squash. You can push with --force to this branch to replace the PR's contents.

templates/rabbitmq.config.erb
firefly4268: Hi David,
Thank you for re-considering my request :)I re-submitted using the same setup as ssl_versions, I took the liberty of cleaning up my logic in the template a bit more and adding in another test.Let me know if there's anything else you'd like done with it,
Megan (firefly4268)

DavidS:
Thanks for your contribution!

PR #211, opened 2014-07-16
Support for configuring rabbitmq_shovel plugin

jimgolfgti:
Adding support for enabling the [Shovel plugin](https://www.rabbitmq.com/shovel.html)

Configuration of static shovels is possible using the `config_shovel_statics` parameter.

```puppet
class { 'rabbitmq' :
  config_shovel         => true,
  config_shovel_statics => {
    'site_to_site' => '{sources,[{broker,"amqp://"}]},
        {destinations,[{broker,"amqp://dr.example.com"}]},
        {queue,<<"to_offsite">>}'
  },
}
```

PR #425, opened 2016-01-14
Adds configuration for rabbitmq_shovel plugin including static shovels

DavidS:


DavidS:
This is a simple rebase and conflict-resolve of #211

PR #432, opened 2016-02-03
Fix rabbitmqadmin url

kozhukalov:
When trying to get rabbitmqadmin the manifest
sends request to localhost which is wrong since
node_ip_address can differ.


bogdando:
@hunner please take a look

bogdando:
+1 from my side

alexz-kh:
+1 from my side.

PR #436, opened 2016-02-20
Allow passing architecture to `apt::source`

alexmuller:
It's useful to be able to set this parameter on `apt::source`.

I've defaulted this param to `undef` which is the same as the default in `apt::source`.


nibalizer:
Thanks!

PR #437, opened 2016-02-24
Added heartbeat option to rabbitmq-server

HelenCampbell:
Added the rabbitmq config option `heartbeat` to provide a heartbeat
connection to nodes. By default this option isnt set in rabbitmq.conf
and its package default is 580 seconds, if the option is set, the config
will reflect this change. There are no other config changes required.

Spec tests provided.


HelenCampbell:
@daniel-gadd Hey, I rebased your PR for you #380 and fixed the conflicts :) Pending Travis running we should be able to get this merged in today. Thanks for all the work you've put in to this.

PR #438, opened 2016-02-29
Update default $package_gpg_key to https

jordanconway:
It is dangerous to not use https for importing gpg keys
from web sources. This shoud be https and not http.

Signed-off-by: Jordan Conway <jconway@linuxfoundation.org>


jordanconway:
Updated with passing tests.

PR #440, opened 2016-03-09
Change $repos_ensure to default to false.

jguiditta:
The problem with it being true by default is that the majority of production
environments, and any in an isolated nextwork, will not want to get the gpg key,
especially if they have not explicitly configured the module to get the rpm from
the rabbitmq repos.  Having the gpg key pulled in by default while installing
the rpm from distro repos seems inconsistent at best, and at worst, causes an
additional fail point for deployments that are unable to contact this location.
Leaving it true by default forces the majority of production users to have to
override this setting.

### DIFF
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -12,6 +12,7 @@ describe 'rabbitmq' do
 
   context 'on Debian' do
     with_debian_facts
+    let(:params) {{ :repos_ensure => true }}
     it 'includes rabbitmq::repo::apt' do
       should contain_class('rabbitmq::repo::apt')
     end
@@ -37,7 +38,7 @@ describe 'rabbitmq' do
   end
 
   context 'on Debian' do
-    let(:params) {{ :manage_repos => true }}
+    let(:params) {{ :manage_repos => true, :repos_ensure => true }}
     with_debian_facts
 
     it 'includes rabbitmq::repo::apt' do
@@ -182,6 +183,7 @@ describe 'rabbitmq' do
 
   context 'on Redhat' do
     with_redhat_facts
+    let(:params) {{ :repos_ensure => true }}
     it 'includes rabbitmq::repo::rhel' do
       should contain_class('rabbitmq::repo::rhel')
       should contain_exec('rpm --import https://www.rabbitmq.com/rabbitmq-signing-key-public.asc')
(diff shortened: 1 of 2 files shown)


EmilienM:
@hunner any opinion on this? It sounds like the initial default value was not really a good choice, but this chance is likely now backward compatible...

hunner:
I want to do a 6.0.0 soon, so it can go in that.

spec/classes/rabbitmq_spec.rb
```diff
@@ -282,9 +284,9 @@
   end
 
   context 'on Redhat' do
-    let(:params) {{ :manage_repos => true }}
+    let(:params) {{ :manage_repos => true, :repos_ensure => true }}
     let(:facts) {{ :osfamily => 'RedHat', :operatingsystemmajrelease => '7' }}
-    it 'does import repo public key when manage_repos is true' do
+    it 'does import repo public key when manage_repos and repos_ensure are true' do
```
jguiditta: Please note this change.  If you disagree with it, then it may be needed to look again at the logic in init.pp, as that checks for manage_repos != false as the condition to include repo class.  Since this parameter is itself deprecated, it seems to me this logic should change, though it is not a requirement for this patch.

jguiditta:
Looks like this needs #451 based on output of the one version of tests that failed - https://travis-ci.org/puppetlabs/puppetlabs-rabbitmq/jobs/116965330

EmilienM:
@hunner do we plan to release 6.0.0 soon?

wyardley:
@hunner @EmilienM, sorry to hijack, but would love some feedback on https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/493, which also makes a lot of changes in this area (though still uses the same GPG key when the repo is ensured only). There's been some discussion about it in the Puppet Community Slack; would love to get some more ideas and for people more familiar with the codebase to review the changes. This also gets rid of $version, which wasn't honored / ensured for Apt / Yum anyway, and pulls out some confusing parameters that were already deprecated, like $manage_repos.

If the regular repos for Debian / Ubuntu have a recent-ish package, I'd be fine with switching to using those by default (vs. the RabbitMQ repo); we'd want to add better test coverage for when $repos_ensure is disabled.  I would be Ok with using the EPEL ones (vs. dropping i  If anyone has more ideas about whether SLES has an available package without any special configuration, that would also be helpful.

trevor-vaughan:
:+1: Always support isolated environments first. Nobody wants random repositories from the outside world magically showing up in their trust chain. Likewise, don't grab random tarballs from the Internet and install them without explicit permission.

wyardley:
Ok, defaults to false now in https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/493
I wonder if the naming is a little confusing (though I'd be hesitant to deprecate yet another variable, considering we just got rid of $manage_repos).

eputnam:
closing because #493

PR #441, opened 2016-03-10
MODULES-3148: Allow shards-per-node for rabbitmq_policy definition to be integer

davewongillies:
Change `rabbitmq_policy` type to pass the policy definition value `shards-per-node` as an integer instead of a string to `rabbitmqctl`, as required by the [rabbitmq-sharding](https://github.com/rabbitmq/rabbitmq-sharding) plugin.


davewongillies:
Is there anyone reviewing these PRs?

DavidS:
Apologies for the delay.

PR #442, opened 2016-03-15
Replace nanliu-staging by voxpupuli-staging

EmilienM:
voxpupuli is the namespace supported by Puppetlabs community, we should
use this repository instead of the fork.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,7 +2,7 @@ fixtures:
   repositories:
     "stdlib": "git://github.com/puppetlabs/puppetlabs-stdlib.git"
     "apt": "git://github.com/puppetlabs/puppetlabs-apt.git"
-    "staging": "git://github.com/nanliu/puppet-staging.git"
+    "staging": "git://github.com/voxpupuli/puppet-staging.git"
     erlang:
       repo: "https://github.com/garethr/garethr-erlang.git"
   symlinks:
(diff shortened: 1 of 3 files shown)


EmilienM:
ping @hunner

lcrisci:
```
└> puppet module install voxpupuli-staging
Notice: Preparing to install into /Users/laurentcrisci/.puppetlabs/etc/code/modules ...
Notice: Created target directory /Users/laurentcrisci/.puppetlabs/etc/code/modules
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Error: Could not install 'voxpupuli-staging' (latest)
  No releases are available from https://forgeapi.puppetlabs.com
    Does 'voxpupuli-staging' have at least one published release?
```

As far as i can see the voxpupuli-staging is not hosted on puppetforge so our dependencies management though librarian-puppet is currently broken.

I would kindly ask to revert this change until we have a proper voxpupuli-staging module hosted on puppetforge ;)

https://github.com/voxpupuli/puppet-staging/issues/84

EmilienM:
Oops, sorry for breakage. Why do you want a revert if the only blocker is to create the module in the forge? Does voxpupuli has a forge?

xiii:
The fact that module is specified here https://github.com/puppetlabs/puppetlabs-rabbitmq/blob/master/metadata.json#L53 . It expects the module to be in the forge, right ?

EmilienM:
@xiii you're right, we can revert it I don't mind.

PR #443, opened 2016-03-15
rabbitmqadmin: bring ipv6 support

EmilienM:
(1) If node_ip_address is an IPv6, we need to pass specific options to curl
so staging module can work.

When using IPv6, curl needs to be used with specific options, for
example with localhost:
```curl -g -6 [::1]```

(2) Sanitize node_ip_address if IPv6
node_ip_address is used bu rabbitmqadmin.pp and also in config.pp for
rabbitmqenv.config. This parameter should not be given with brackets,
otherwise RabbitMQ won't start.
Though when using curl, we need the add the brackets so this patch makes
sure the brackets are added to node_ip_address if it's an IPv6 so curl
will work correctly.

### DIFF
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -14,10 +14,18 @@ class rabbitmq::install::rabbitmqadmin {
   $default_pass = $rabbitmq::default_pass
   $node_ip_address = $rabbitmq::node_ip_address
 
+  if is_ipv6_address($node_ip_address) {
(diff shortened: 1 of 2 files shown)


EmilienM:
do not merge it yet, we need https://github.com/puppetlabs/puppetlabs-stdlib/pull/577 first.

EmilienM:
It requires https://github.com/puppetlabs/puppetlabs-stdlib/pull/577 to work.

PR #444, opened 2016-03-16
Rabbitmqadmin IPv6 non-working.

sathlan:
The support for IPv6 address for `rabbitmqadmin` is non-working leading
to this error:

```
Debug: Exec[/var/lib/rabbitmq/rabbitmqadmin](provider=posix): Executing 'curl -k --noproxy -g -6 [::1] --retry 30 --retry-delay 6 -f -L -o /var/lib/rabbitmq/rabbitmqadmin "https://guest:guest@[::1]:15671/cli/rabbitmqadmin"'
Debug: Executing 'curl -k --noproxy -g -6 [::1] --retry 30 --retry-delay 6 -f -L -o /var/lib/rabbitmq/rabbitmqadmin "https://guest:guest@[::1]:15671/cli/rabbitmqadmin"'
Notice: /Stage[main]/Rabbitmq::Install::Rabbitmqadmin/Staging::File[rabbitmqadmin]/Exec[/var/lib/rabbitmq/rabbitmqadmin]/returns: curl: (3) [globbing] illegal character in range specification at pos 2
Notice: /Stage[main]/Rabbitmq::Install::Rabbitmqadmin/Staging::File[rabbitmqadmin]/Exec[/var/lib/rabbitmq/rabbitmqadmin]/returns: curl: (3) [globbing] illegal character in range specification at pos 22
Error: curl -k --noproxy -g -6 [::1] --retry 30 --retry-delay 6 -f -L -o /var/lib/rabbitmq/rabbitmqadmin "https://guest:guest@[::1]:15671/cli/rabbitmqadmin" returned 3 instead of one of [0]
Error: /Stage[main]/Rabbitmq::Install::Rabbitmqadmin/Staging::File[rabbitmqadmin]/Exec[/var/lib/rabbitmq/rabbitmqadmin]/returns: change from notrun to 0 failed: curl -k --noproxy -g -6 [::1] --retry 30 --retry-delay 6 -f -L -o /var/lib/rabbitmq/rabbit
```

The problem is that the order of the parameters is all mangled in the
wrong places:
 - `--noproxy` takes the `::1` as parameter, not `-g`
 - `-g` : is ignored as it's the parameter of `--noproxy`
 - `--noproxy` takes `::1` as parameter, not `[::1]`

Finally to have the `rabbitmq_management` server listen on the ip
specified in the curl command (the `node_ip_address`), you have to have
the management server listen to it.  I added the necessary bits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,8 @@ Valid values are true or false.
 
 ####`node_ip_address`
 
-The value of NODE_IP_ADDRESS in rabbitmq_env.config
+The value of NODE_IP_ADDRESS in rabbitmq_env.config and of the
+rabbitmq_management server if it is enabled.
 
 ####`package_ensure`
 
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -15,17 +15,17 @@ class rabbitmq::install::rabbitmqadmin {
   $node_ip_address = $rabbitmq::node_ip_address
 
   if is_ipv6_address($node_ip_address) {
-    $curl_prefix  = '-k --noproxy -g -6'
+    $curl_prefix  = '-g -6'
     $sanitized_ip = join(enclose_ipv6(any2array($node_ip_address)), ',')
   } else {
-    $curl_prefix  = '-k --noproxy'
+    $curl_prefix  = ''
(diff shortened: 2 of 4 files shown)


EmilienM:
Thanks for the follow-up, I missed that in my previous patch.

PR #445, opened 2016-03-16
rabbitmq_management block is created twice.

sathlan:
Following this commit 8f75ff5cfea7945a0d6a4074184a7279a81349d4, if one
use this simple manifest:

```puppet
class { 'rabbitmq':
    admin_enable   => true,
    service_manage => true,
}
```

Then two blocks of

```
{rabbitmq_management, [
]}
```

gets created.

The problem is that only the parameters of the first block is taken into
account.

Furthermore the /first/ block is *always* created.  The template has:

```
+<% if @config_management_variables -%>,
+  {rabbitmq_management, [
+    <%= @config_management_variables.sort.map{|k,v| "{#{k}, #{v}}"}.join(",\n    ") %>
+  ]}
+<%- end -%>
```

and config_management_variables is set to `{}` with is not `false`.

This is an attempt to fix this and keep the same behavior.

### DIFF
--- a/templates/rabbitmq.config.erb
+++ b/templates/rabbitmq.config.erb
@@ -69,13 +69,13 @@
     <%= @config_kernel_variables.sort.map{|k,v| "{#{k}, #{v}}"}.join(",\n    ") %>
   ]}
 <%- end -%>
-<% if @config_management_variables -%>,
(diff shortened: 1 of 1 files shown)


EmilienM:
Nice catch, thanks for the quick fix.

PR #446, opened 2016-03-16
Revert "Replace nanliu-staging by voxpupuli-staging"

lcrisci:
As discussed in https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/442

voxpupuli-staging is currently not published/hosted on Puppet Forge which is causing issues.

We should revert back to using nanliu-staging until voxpupuli-staging gets published


madAndroid:
It looks as though the naming is `puppet-staging`, not `voxpopuli-staging` - if we update the metadata.json to show the correct name, this revert is possible not necessary.

madAndroid:
nope, that's not it either .. my bad .. ignore that :)

PR #448, opened 2016-03-17
Fix dependency name

hunner:
This was recently updated, but with the wrong name.

Closes #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -650,9 +650,9 @@ For Debian systems:
       ensure => 'latest',
     }
 
-This module also depends on the excellent voxpupuli/staging module on the Forge:
+This module also depends on the excellent puppet/staging module on the Forge:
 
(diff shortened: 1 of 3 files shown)


lcrisci:
puppet-staging doesn't seem to exist on Puppet Forge:

```
puppet module install puppet-staging
WARN: Unresolved specs during Gem::Specification.reset:
      json_pure (>= 0)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
Notice: Preparing to install into /Users/laurentcrisci/.puppetlabs/etc/code/modules ...
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Error: Could not install 'puppet-staging' (latest)
  No releases are available from https://forgeapi.puppetlabs.com
    Does 'puppet-staging' have at least one published release?
```

PR #453, opened 2016-03-23
Add handshake_timeout option to rabbitmq config

mpolenchuk:
Bring in handshake_timeout option to rabbitmq config (10 seconds by
default). When clients run in heavily constrained environments,
it may be necessary to increase the timeout.


bogdando:
+1 LGTM, thank you!

bogdando:
@hunner could you please take a look?
CI failure looks not related, AFAICT...

mpolenchuk:
All rabbit.* config options could be handled by $config_variables param of rabbitmq::init class.

PR #454, opened 2016-03-23
Make config world readable no more

paramite:
The file /etc/rabbitmq/rabbitmq.config contains rabbitmq credentials
and should not be world readable.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -71,14 +71,14 @@ class rabbitmq::config {
     ensure => directory,
     owner  => '0',
     group  => '0',
-    mode   => '0644',
+    mode   => '0755',
   }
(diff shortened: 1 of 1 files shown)


wyardley:
How was it not a problem that the directories were 0644 vs. 0755 already? This seems sensible to me, but if RabbitMQ is running as an unprivileged user, wouldn't it make some sense to have $rabbitmq_group own the config file that's 0640?

paramite:
This patch is so old that I really don't remember, exactly. But shouldn't directories have 'x' so you can actually access it? Anyway, yes having rabbitmq group own the files makes sense.

hunner:
Merged in #560

PR #455, opened 2016-04-01
Add support to uninstall rabbitmq plugins.

acowan:
This is an update of pull request #388 from 14 Sep 2015.  There have been quite a number of changes to puppetlabs-rabbitmq in the interim, so I rewrote the patch.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -223,52 +223,48 @@ class rabbitmq(
     }
   }
 
-  if $admin_enable and $service_manage {
+  $admin_ensure = $admin_enable and $service_manage
+
+  if $admin_ensure {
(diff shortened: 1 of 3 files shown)


acowan:
Looks like a broken test run -- unlikely that I broke every test case.

wyardley:
@acowan this may get a little more traction now that the module is under Voxpupuli, however, can you rebase against master (and squash commits if you can)? That should trigger a new run of the acceptance tests too.

wyardley:
This can also probably be done more cleanly now that we have data types and have true booleans for most (all?) of these options.

wyardley:
I'm going to close this for now, since there's been no action for a while.

PR #456, opened 2016-04-06
add default_vhost parameter and test

flyinbutrs:
Ability to set default_vhost in rabbitmq.config.

Part of me thinks this should check in some way require that the specified vhost be defined, though I'm not exactly sure how. Thoughts?

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -11,6 +11,7 @@ class rabbitmq(
   $config_shovel_statics      = $rabbitmq::params::config_shovel_statics,
   $default_user               = $rabbitmq::params::default_user,
   $default_pass               = $rabbitmq::params::default_pass,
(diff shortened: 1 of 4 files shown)


wyardley:
This may be a bit less useful with some of the newer features for setting defaults in newer versions of Puppet, but in either event, seems to need a rebase.

As far as the failing acceptance tests, the error was 
`Evaluation Error: Error while evaluating a Resource Statement, default expression for $config_management_variables tries to illegally access not yet evaluated $config_management_variables at line 1:1 on node testing-worker-linux-docker-251405dc-3372-linux-5.prod.travis-ci.org`, which would seem to be an unrelated issue? If you rebase, squash, and force-push, we can see if the test passes.

wyardley:
I'm going to close now, please rebase and re-open if you're interested in continuing this one.

PR #463, opened 2016-04-20
Add rabbitmq_nodename fact

jyaworski:


hunner:
What is the usecase for this fact?

jyaworski:
The use-case is for cases when certain nodes, for whatever reason, are managed outside of puppet and we still need their nodename. It's also useful in scripts. For example, writing queries to hit /api/nodes/$::rabbitmq_nodename.

PR #464, opened 2016-04-21
Cleanup void tcp options

mpolenchuk:
After Ranch (socket acceptor pool) has come into effect
the following tcp listen options is no longer valid:
binary, packet, reuseaddr.


bogdando:
LGTM +1

nibalizer:
I would prefer a longer commit message explaining why and how ranch affects things. At the least, it seems like version previous to X would still need these flags, can we hide this config behind a version check conditional?

hunner:
Yeah, we'll need a lot more explanation before we just delete three tcp listen options.

If you would prefer to make them conditionally removed based on either a parameter or detection of ranch, that would be much easier to digest.

mpolenchuk:
I'll wrap it in condition of ranch presence.
explanation: https://github.com/rabbitmq/rabbitmq-server/issues/453

mpolenchuk:
done

mpolenchuk:
bump

eputnam:
@mpolenchuk thanks for the PR! we'd like to get some of these rabbitmq PRs merged up. would you mind rebasing this one?

mpolenchuk:
done

PR #465, opened 2016-04-26
(#3295) Give an opportunity to suppress ssl_stomp warning

sorrowless:
Validate ssl_stomp_port as boolean if it is not set as integer.
It allows to set it as 'false' in case when there is disabled
ssl and configured non-ssl stomp, which will suppress whe warning
about '$ssl_stomp_port requires that $ssl => true'.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -131,7 +131,7 @@ class rabbitmq(
     validate_re($ssl_management_port, '\d+')
   }
   if ! is_integer($ssl_stomp_port) {
-    validate_re($ssl_stomp_port, '\d+')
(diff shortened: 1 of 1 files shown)


bogdando:
-1 The port is expected to be a number, not a boolean.  Why to supress a warning? It looks valid

sorrowless:
@bogdando, why do you think that warning is valid here? If user want to disable SSL, he said 'ssl => false' but in this case he will always get a warning about the fact that stomp_ssl_port is set (because there *is* default value for it) but ssl is disabled. I perfectly know that it is disabled, because I did it _manually_, why I should see this warning?

PR #469, opened 2016-05-10
(maint) Fixes acceptance tests for centos7 and v3+

bmjen:
Also adds travis-ci jobs to run acceptance tests.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,4 +1,0 @@
-# the rabbitmq tests currently require a dev version of stdlib, but do not properly configure it
----
-.travis.yml:
-  docker_sets: []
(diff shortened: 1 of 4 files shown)


DavidS:
this change conflicts with https://github.com/puppetlabs/puppetlabs-rabbitmq/blob/master/.sync.yml#L4

DavidS:
You can add only the working ubuntu there (compare to https://github.com/puppetlabs/modulesync_configs/blob/master/config_defaults.yml#L5) to do this right.

bmjen:
@DavidS I still want to keep the centos7 tests in there, despite them failing, as they are only on the 2.8.1 tests, and we should know if anything breaks 3.0+

PR #470, opened 2016-05-10
Improve robustness of rabbitmq_nodename

mnaser:
With certain platforms, the nodename is quoted resulting in a NilClass
error.  In addition, the regex does not properly match if a nodename
has dashes.  This commit resolves these issues.

### DIFF
--- a/lib/facter/rabbitmq_nodename.rb
+++ b/lib/facter/rabbitmq_nodename.rb
@@ -2,7 +2,7 @@ Facter.add(:rabbitmq_nodename) do
   setcode do
     if Facter::Core::Execution.which('rabbitmqctl')
       rabbitmq_nodename = Facter::Core::Execution.execute('rabbitmqctl status 2>&1')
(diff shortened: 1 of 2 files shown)


davewongillies:
Oh yes, this please. Its resulting in this error with RabbitMQ 3.6.1 (I'm not sure which version the `'` characters turned up in the `rabbitmq status` output:
```
Could not retrieve fact='rabbitmq_nodename', resolution='<anonymous>': undefined method `[]' for nil:NilClass
```

Also since its not mentioned in the PR, the regex adds `-` when matching hostnames in some of us have dashes in our host names.

seocam:
Any chance to get a new release with the PR soon? I'm unable to use this module because my servers follow a naming convention that is causing the error:

```Error: Facter: error while resolving custom fact "rabbitmq_nodename": undefined method `[]' for nil:NilClass```

If I patch my installation of puppet rabbitmq with this PR it works fine.

PR #472, opened 2016-05-18
Quick fix to parse the Facter rabbitmq_nodename correctly

Asher256:
This quick fix solves this error (returned by rabbitmq_nodename.rb):

```
Could not retrieve fact='rabbitmq_nodename', resolution='<anonymous>':
undefined method `[]' for nil:NilClass'`
```

The error below happened because the string returned by rabbitmqctl in my servers is:

```
$ rabbitmqctl status
Status of node 'rabbit@vnp-p-web005' ...
```

The old regular expression wasn't able to parse the rabbitmqctl's output below
correctly because it contains - (dashes) and is surrounded by '
(simple quotes):
%r{^Status of node ([\w.]+@[\w.]+) .+$}.match(rabbitmq_nodename)[1]

The new regular expression gives the possibility to parse strings with
OR without simple quotes and dashes:
%r{^Status of node[ "']+([\w.-]+@[\w.-]+)[ "']+.+$}.match(rabbitmq_nodename)[1]

### DIFF
--- a/lib/facter/rabbitmq_nodename.rb
+++ b/lib/facter/rabbitmq_nodename.rb
@@ -2,7 +2,7 @@ Facter.add(:rabbitmq_nodename) do
   setcode do
     if Facter::Core::Execution.which('rabbitmqctl')
       rabbitmq_nodename = Facter::Core::Execution.execute('rabbitmqctl status 2>&1')
-      %r{^Status of node ([\w\.]+@[\w\.]+) \.+$}.match(rabbitmq_nodename)[1]
+      %r{^Status of node[ "']+([\w\.\-]+@[\w\.\-]+)[ "']+\.+$}.match(rabbitmq_nodename)[1]
     end
   end
 end


guikcd:
:+1:

pietervogelaar:
+1

mclueppers:
On my system it still fails to capture the node name unless I enable multi-line, so I had to change the line to
```
%r{^Status of node[ \"\']+([\w\.-]+@[\w\.-]+)[ \"\']+\.+$}m.match(rabbitmq_nodename)[1]
```

oneiros-de:
Note that the line with 3.6.10 is 
`Status of node rabbit@endive4`
i.e. without quotes and without trailing dots

michaelklishin:
@Asher256 may I ask you to rebase this against master and perhaps take @mclueppers's comment above into consideration?

It would be nice to see this finished and merged. Not sure how team RabbitMQ can speed this up since we don't maintain this module.

TraGicCode:
Can this please get merged in if it's good?  We are running into this issue as well...

wyardley:
@TraGicCode @michaelklishin 

I swapped in the node name in question in the test here:
https://github.com/voxpupuli/puppet-rabbitmq/blame/master/spec/unit/rabbitmq_nodename_spec.rb#L29-L37

and it seemed to grok the hostname as  'rabbit@vnp-p-web005' Ok. So I think this should  be fixed as of a year ago  or so? Can someone verify if this problem exists with a recent version of the module (and if so, provide a hostname that's not matching)?

TraGicCode:
Hey, I had this problem when I was using the puppetlabs/rabbitmq module on the forge.  Once I pointed it to a sha on master things are fine.  Therefore I believe this is fixed there is just a missing release



Sent via the Samsung GALAXY S® 5, an AT&T 4G LTE smartphone


-------- Original message --------

wyardley:
Going to close this then, I think this should already have been fixed in the current version of the module.

PR #473, opened 2016-05-18
Quick fix: backward compatibility with Puppet 2 / has_command()

Asher256:
Because has_command() doesn't exist in Puppet 2,
lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb failed in our
servers with this error:

```
err: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not autoload rabbitmq_erlang_cookie: Could not autoload /opt/puppet/environments/production/modules/rabbitmq/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb: undefined method `has_command' for Puppet::Type::Rabbitmq_erlang_cookie::ProviderRuby:Class at /opt/puppet/environments/production/modules/rabbitmq/manifests/config.pp:168 on node vpn-p-sst003.mdc.gameloft.org
```

This quick fix provide an alternative to has_command
for Puppet 2:
commands :COMMANDNAME => "/PATH/COMMANDNAME"

### DIFF
--- a/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
+++ b/lib/puppet/provider/rabbitmq_erlang_cookie/ruby.rb
@@ -1,16 +1,86 @@
 require 'puppet'
 require 'set'
+
+# Backware compatibility with Puppet 2
+# Source: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util.rb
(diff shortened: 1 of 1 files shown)


wyardley:
I think this can be closed since Puppet 2 (and even 3) are no longer supported.

PR #474, opened 2016-05-19
add service provider to service and params

bilsch:
We have noticed the rabbitmq-server service on rhel7 systems always changes. Setting the provider to redhat resolves the issue.

I added the service_provider param to all OS cases but I don't know what the proper value should be. I'm happy to fill in additional values and add spec tests for coverage.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -15,6 +15,7 @@ class rabbitmq::params {
       $rabbitmq_group   = 'rabbitmq'
       $rabbitmq_home    = '/var/lib/rabbitmq'
       $plugin_dir       = "/usr/lib/rabbitmq/lib/rabbitmq_server-${version}/plugins"
+      $service_provider = undef
     }
     'Debian': {
       $package_ensure   = 'installed'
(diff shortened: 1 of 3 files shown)


bilsch:
is there a way to re-trigger the travis job? Looks like a transient failure on the centos-7 validation

alexjfisher:
Is this related to https://tickets.puppetlabs.com/browse/PUP-2744 ?

bilsch:
Looks like it, yes

```
cat /etc/redhat-release ; rpm -qa | grep puppet
Red Hat Enterprise Linux Server release 7.2 (Maipo)
puppet-agent-1.3.2-1.el7.x86_64
```

We have found/reported this issue on a number of other modules. The symptom is puppet always setting the service to enabled => true.

alexjfisher:
https://dl.fedoraproject.org/pub/epel/7/x86_64/r/rabbitmq-server-3.3.5-17.el7.noarch.rpm (which the travis build has installed) *does* use systemd, so setting the provider to 'redhat' breaks the build there.

On the other hand, the official package uses an init.d service script. https://www.rabbitmq.com/releases/rabbitmq-server/v3.6.2/rabbitmq-server-3.6.2-1.noarch.rpm

I can't think of a clean way of having the module automatically figure out whether it needs to workaround https://tickets.puppetlabs.com/browse/PUP-2744

In the situations I've bumped into the same issue, I've had good results setting a 'resource default' in the profile.  eg
```
Service {
  provider = 'redhat',
}
class { '::rabbitmq:
}
```

bilsch:
that works thanks @alexjfisher

PR #475, opened 2016-05-25
rabbitmq signing key has moved

igalic:
(fingerprint is still the same)

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -7,7 +7,7 @@ class rabbitmq::repo::apt(
   $repos        = 'main',
   $include_src  = false,
   $key          = 'F78372A06FF50C80464FC1B4F7B8CEA6056E8E56',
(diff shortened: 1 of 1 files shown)


EmilienM:
that's not enough, you'll need to patch manifests/params.pp: and update all unit tests.

alexjfisher:
@igalic Are you sure about the fingerprint?
https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/477 has it updated.

bilsch:
while you are looking at this one, mind looking at my PR - seems I'm also suffering from travis build failures - #474

michaelklishin:
Heads-up, rabbitmq.com apt repo was [re-signed with the new release key](https://groups.google.com/d/msg/rabbitmq-users/BO5cmEsdEhc/Jupz1_Q4AwAJ).

PR #476, opened 2016-05-25
move rabbitmq signing key

EmilienM:
Signing key moved to https://www.rabbitmq.com/rabbitmq-release-signing-key.asc

This patch aims to update the module.

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -6,8 +6,8 @@ class rabbitmq::repo::apt(
   $release      = 'testing',
   $repos        = 'main',
   $include_src  = false,
-  $key          = 'F78372A06FF50C80464FC1B4F7B8CEA6056E8E56',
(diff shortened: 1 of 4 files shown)


EmilienM:
I took over  https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/475 so we can have a quick fix.

manifests/repo/apt.pp
```diff
@@ -7,7 +7,7 @@
   $repos        = 'main',
   $include_src  = false,
   $key          = 'F78372A06FF50C80464FC1B4F7B8CEA6056E8E56',
```
alexjfisher: @EmilienM Could you confirm whether you think `$key` needs updating too?

EmilienM: updated.

EmilienM:
too many PR try to fix it, I'll close mine.

michaelklishin:
http://www.rabbitmq.com/rabbitmq-signing-key-public.asc is available again.

PR #477, opened 2016-05-25
Signing key patch 1

ovecka:
Due change on rabbitmq.com where they have removed old signaturing key and published new key, puppet and apt is not able to verify package. 
This solution work for me.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -70,7 +70,7 @@ class rabbitmq::params {
   $management_port             = '15672'
   $management_ssl              = true
   $package_apt_pin             = ''
-  $package_gpg_key             = 'https://www.rabbitmq.com/rabbitmq-signing-key-public.asc'
+  $package_gpg_key             = 'https://www.rabbitmq.com/rabbitmq-release-signing-key.asc'
   $repos_ensure                = true
   $manage_repos                = undef
   $service_ensure              = 'running'
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -6,8 +6,8 @@ class rabbitmq::repo::apt(
   $release      = 'testing',
   $repos        = 'main',
   $include_src  = false,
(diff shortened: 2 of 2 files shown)


alexjfisher:
@ovecka @EmilienM Thanks.

I suspect this PR and https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/476 need combining.

ovecka:
I have to say that final installation by apt-get fails because rabbitmq package is still signed with old GPG and old key is not availaible... After they 'll release new version, it will work. Not until then...

michaelklishin:
FTR, http://www.rabbitmq.com/rabbitmq-signing-key-public.asc is available again.

michaelklishin:
Heads-up, rabbitmq.com apt repo was [re-signed with the new release key](https://groups.google.com/d/msg/rabbitmq-users/BO5cmEsdEhc/Jupz1_Q4AwAJ).

ovecka:
Hahaha ... this is more funny than I expected :) 
So this is actually needed now and should work... gonna try atp-get

ovecka:
There is problem with https key for apt-get. When I changed that to http (in params.pp), it worked.

dgmorales:
Note for any user being bitten by this: as a temporary workaround until this gets merged, one can place the following in hiera:

```
rabbitmq::package_gpg_key: 'https://www.rabbitmq.com/rabbitmq-release-signing-key.asc'
rabbitmq::repo::apt::key: '0A9AF2115F4687BD29803A206B73A36E6026DFCA'
```

rski:
@alexjfisher since https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/481 has been merged, can this be closed now?
@ovecka the other patch works for me so you should be fine too

wyardley:
Looks like 0A9AF2115F4687BD29803A206B73A36E6026DFCA is already the key in the module, so I think we can close this.

PR #478, opened 2016-06-01
Do not check cert when acquiring rabbitmqadmin with wget.

modax:
It is an equivalent of curl_option='-k'.


wyardley:
This seems reasonable, since curl has `-k` set already.

It seems like the issue w/ the single failing test is not due to not updating a test, but to a probably unrelated problem while running the Ubuntu acceptance test? I'm inclined to merge this.

PR #479, opened 2016-06-01
Add support for configuration of loopback_users

gfidente:
Makes it possible to configure the list of loopback_users. This is
particularily useful to allow access from all interfaces to the
'guest' user which is limited to 'localhost' by default [1].

1. https://www.rabbitmq.com/access-control.html


gfidente:
Closed in favor of https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/480

PR #480, opened 2016-06-01
Add support for configuration of loopback_users

gfidente:
Makes it possible to configure the list of loopback_users. This is
particularily useful to allow access from all interfaces to the
'guest' user which is limited to 'localhost' by default [1].
1. https://www.rabbitmq.com/access-control.html


wyardley:
@gfidente Do you still want this feature? If so, can you rebase (and squash any additional commits), and fix the tests, as well as (ideally) maybe add any additional tests that might be helpful?

dn1s:
If you want to be able to access management interface with guest just use this:
```
rabbitmq_server::config_variables:
  loopback_users: []
```

wyardley:
This has also now been added in #699 
Closing via #699

PR #481, opened 2016-06-03
Update apt.pp

Yuav:
MODULES-3439 Update apt repo defaults to correspond with manual from: https://www.rabbitmq.com/install-debian.html

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -70,7 +70,7 @@ class rabbitmq::params {
   $management_port             = '15672'
   $management_ssl              = true
   $package_apt_pin             = ''
-  $package_gpg_key             = 'https://www.rabbitmq.com/rabbitmq-signing-key-public.asc'
+  $package_gpg_key             = 'https://www.rabbitmq.com/rabbitmq-release-signing-key.asc'
   $repos_ensure                = true
   $manage_repos                = undef
   $service_ensure              = 'running'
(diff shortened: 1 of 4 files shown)


manojsamtani:
I have tried and was able to resolve this by calling rabbitmq class as

class { 'rabbitmq':
        package_gpg_key => 'http://www.rabbitmq.com/rabbitmq-release-signing-key.asc', }

Because rabbitmq class calling repo/apt.pp from init.pp with package_gpg_key variable which is actually pointing to rabbitmq::params::package_key

So, better to call rabbitmq class as given above example with new key URL

hauleth:
Still doesn't work for me.

manojsamtani:
Please update rabbitmq module as well with following key

class rabbitmq::repo::apt(
  $location    = 'http://www.rabbitmq.com/debian/',
  $release     = 'testing',
  $repos       = 'main',
  $include_src = false,
  #$key         = 'F78372A06FF50C80464FC1B4F7B8CEA6056E8E56',
  **$key         = '0A9AF2115F4687BD29803A206B73A36E6026DFCA',**
  $key_source  = 'http://www.rabbitmq.com/rabbitmq-signing-key-public.asc',

hauleth:
@manojsamtani the issue was with stale machine. Fixed now.

DanEEStar:
Please make a new release with this new signing keys!

yrobla:
I'm hitting the same problem with 5.4.0 , a new release with new keys is needed.

acbox:
+1, please make a new release

PR #482, opened 2016-06-15
Fix regex double escaping of rabbitmqctl list_policies

sspreitzer:


devfaz:
+1

DrOwl:
I have tested this patch and it fixes the problem for us we no longer get an update on each run:

before patch we get the below every run
/Stage[main]/manage_rabbitmq::Federation/Rabbitmq_policy[eu-federation@vhost]/pattern: pattern changed '^eu\\.' to '^eu\.'

eputnam:
@sspreitzer could you add some unit tests for this?

DrOwl:
Ok well i have no idea how add more to a pull request but here is some tests
[spec_test.patch.txt](https://github.com/puppetlabs/puppetlabs-rabbitmq/files/1128617/spec_test.patch.txt)

Theses have not been tested as "you guys and your crazzy Beaker"

wyardley:
If folks don't mind, I added @DrOwl's test (with attribution to @sspreitzer for the whole commit, couldn't figure out how to break it up) and resubmitted as #561

PR #489, opened 2016-07-28
Add definition validation and munge for ha-sync-batch-size

combatdud3:
With the release of RabbitMQ 3.6 there is an option to set the ha-sync-batch-size.
This option takes an integer >0. This commit adds a validation statement to check
the supplied value is an integer and then converts the string input into an
integer.

### DIFF
--- a/lib/puppet/type/rabbitmq_policy.rb
+++ b/lib/puppet/type/rabbitmq_policy.rb
@@ -107,7 +107,7 @@ Puppet::Type.newtype(:rabbitmq_policy) do
     end
     if definition.key? 'ha-sync-batch-size'
       ha_sync_batch_size_val = definition['ha-sync-batch-size']
-      unless ha_sync_batch_size_val.to_i.tos == ha_sync_batch_size_val
+      unless ha_sync_batch_size_val.to_i.to_s == ha_sync_batch_size_val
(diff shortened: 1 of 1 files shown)


combatdud3:
I'm not sure how what I changed has caused tests to fail. Would someone with more experience be able to help me figure out why?

lib/puppet/type/rabbitmq_policy.rb
```diff
@@ -105,6 +105,12 @@ def validate_definition(definition)
         raise ArgumentError, "Invalid shards-per-node value '#{shards_per_node_val}'"
       end
     end
+    if definition.key? 'ha-sync-batch-size'
+      ha_sync_batch_size_val = definition['ha-sync-batch-size']
+      unless ha_sync_batch_size_val.to_i.tos == ha_sync_batch_size_val
+        raise ArgumentErro, "Invalid ha-sync-batch-size value '#{ha_sync_batch_size_val}'"
```
mxftw: **raise ArgumentError**

mxftw:
@combatdud3 You have a typo in `raise ArgumentError`.
I fixed that in my PR https://github.com/puppetlabs/puppetlabs-rabbitmq/pull/500 and added a Test for it.

mxftw:
#500

PR #490, opened 2016-08-03
basic support for multiple routing keys

wyardley:
This still needs some work (and test coverage), but here's how I was able to get support for multiple routing keys to work in a backwards compatible way. Would appreciate feedback / improvements / guidance on best ways to add test cases for this.

I don't see a way to support removing a binding with one routing key while adding or ensuring another... I think this would have to be done by including the routing key as part of the key of the main hash.

I do get this error on runs, though seems to work in basic cases for ensuring presence / absence, including when new bindings are added. Presumably a data structure issue with resources?

```Error: Could not prefetch rabbitmq_binding provider 'rabbitmqadmin': undefined method `[]' for #<Binding:0x00000002a08a28>
Notice: /Stage[main]/Role::Rabbitmq/Rabbitmq_binding[foo.bar.xyz@foo.bar.xyz.queue@/]/ensure: created
```

### DIFF
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -50,13 +50,19 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
           arguments = '{}'
         end
         if (source_name != '')
-          binding = {
-            :destination_type => destination_type,
-            :routing_key      => routing_key,
-            :arguments        => JSON.parse(arguments),
-            :ensure           => :present,
-            :name             => "%s@%s@%s" % [source_name, destination_name, vhost],
-          }
+          if not binding[:name]
+            routing_key = [routing_key]
+            binding = {
+              :destination_type => destination_type,
+              :routing_key      => routing_key,
+              :arguments        => JSON.parse(arguments),
+              :ensure           => :present,
+              :name             => "%s@%s@%s" % [source_name, destination_name, vhost],
+            }
+          else
+            # If we already have the binding once, just append to the routing_key
+            binding[:name][:routing_key].push(routing_key)
+          end
           resources << new(binding) if binding[:name]
         end
       end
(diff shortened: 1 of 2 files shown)


wyardley:
ps - This is related to https://tickets.puppetlabs.com/browse/MODULES-3679

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -50,13 +50,19 @@ def self.instances
...
         end
         if (source_name != '')
-          binding = {
-            :destination_type => destination_type,
-            :routing_key      => routing_key,
-            :arguments        => JSON.parse(arguments),
-            :ensure           => :present,
-            :name             => "%s@%s@%s" % [source_name, destination_name, vhost],
-          }
+          if not binding[:name]
```
nakedsushi: The original code that was deleted creates a hash and assigns it to the variable 'binding'.  This new code assumes the variable that points to the hash already exists -- does it exist yet?  This is the same as saying `nil[:name]`, which causes an error.

Consider: `if not binding`

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -85,27 +91,36 @@ def create
...
-      '-c',
-      '/etc/rabbitmq/rabbitmqadmin.conf',
-      "source=#{name}",
-      "destination=#{destination}",
-      "arguments=#{arguments.to_json}",
-      "routing_key=#{resource[:routing_key]}",
-      "destination_type=#{resource[:destination_type]}"
-    )
+    # Accommodate multiple bindings with different routing keys
+    for routing_key in resource[:routing_key]
```
nakedsushi: This code implies that `resource[:routing_key]` holds an array. Does it?

wyardley:
cut / pasted from my comment in Puppet ticket...

I'm definitely getting a better understanding of why that doesn't work (even though it "works"), but I'm wondering if it's cleaner (and more in line with how the code was designed) to have a 1:1 mapping of binding to binding. However, at that point, we'd need to encode a lot in the resource name to make it unique, e.g.,
name@destination:routing_key@vhost
or
name@destination@routing_key@vhost
or something. This would also change the interface.

Alternately, we could still treat multiple bindings with the same source / dest / vhost as one binding, and rework the code to accommodate, but that seems kind of kludgy.

wyardley:
Tests passing again, and working(ish), but still some todos:
1) Need to fix exists? (and maybe other stuff) so that case where resource is ensured or absent but only with one of the routing keys will only remove the specified ones. For now, if the resource source@destination@vhost: exists, Puppet will not validate 
2) Add (working) tests for multiple routing key scenarios.

PR #497, opened 2016-08-15
Updated app.pp to address deprecation warnings

ilium007:
Puppetlabs Apt module now has deprecation warnings when call in the manner being used by the puppetlabs rabbitmq module. Updated and tested against puppetlabs apt module 2.2.2


ilium007:
Fixed indenting - will raise a new pull request

hunner:
Could you also update the apt in metadata.json for this? We'll put this in the next major bump of rabbitmq.

PR #504, opened 2016-09-01
Multiple routing keys for bindings using namevars

wyardley:
Updated: proposing this as an option instead of #375 
This PR lets you specify legacy style resources (`source@destination@vhost`) and / or lets you specify `source`, `destination` (and, optionally, `routing_key` and `vhost`) as separate parameters, with a descriptive or arbitrary name.

There's still probably some room for improvement, but I've finally got this in shape where I think it could be merged.

### DIFF
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -1,5 +1,7 @@
 require 'json'
 require 'puppet'
+require 'digest'
+
 Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
 
   if Puppet::PUPPETVERSION.to_f < 3
@@ -14,6 +16,9 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
     end
   end
   defaultfor :feature => :posix
+
+  # Without this, the composite namevar stuff doesn't work properly.
+  mk_resource_methods
 
   def should_vhost
     if @should_vhost
@@ -49,16 +54,17 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
         else
           arguments = '{}'
         end
+        hashed_name = Digest::SHA256.hexdigest "%s@%s@%s@%s" % [source_name, destination_name, vhost, routing_key]
         unless(source_name.empty?)
           binding = {
             :source           => source_name,
-            :dest             => destination_name,
+            :destination      => destination_name,
             :vhost            => vhost,
             :destination_type => destination_type,
             :routing_key      => routing_key,
             :arguments        => JSON.parse(arguments),
             :ensure           => :present,
-            :name             => "%s@%s@%s@%s" % [source_name, destination_name, vhost, routing_key],
+            :name             => hashed_name,
           }
           resources << new(binding) if binding[:name]
         end
@@ -67,10 +73,12 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
     resources
   end
 
+  # see
+  # https://github.com/puppetlabs/puppetlabs-netapp/blob/d0a655665463c69c932f835ba8756be32417a4e9/lib/puppet/provider/netapp_qtree/sevenmode.rb#L66-L73
   def self.prefetch(resources)
     bindings = instances
-    resources.keys.each do |name|
-      if provider = bindings.find{ |route| route.source == source && route.dest == dest && route.vhost == vhost && route.routing_key == routing_key }
+    resources.each do |name, res|
+      if provider = bindings.find{ |binding| binding.source == res[:source] && binding.destination == res[:destination] && binding.vhost == res[:vhost] && binding.routing_key == res[:routing_key] }
         resources[name].provider = provider
       end
     end
@@ -94,7 +102,7 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
       '-c',
       '/etc/rabbitmq/rabbitmqadmin.conf',
       "source=#{resource[:source]}",
-      "destination=#{resource[:dest]}",
+      "destination=#{resource[:destination]}",
       "arguments=#{arguments.to_json}",
       "routing_key=#{resource[:routing_key]}",
       "destination_type=#{resource[:destination_type]}"
@@ -104,7 +112,7 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
 
   def destroy
     vhost_opt = should_vhost ? "--vhost=#{should_vhost}" : ''
-    rabbitmqadmin('delete', 'binding', vhost_opt, "--user=#{resource[:user]}", "--password=#{resource[:password]}", '-c', '/etc/rabbitmq/rabbitmqadmin.conf', "source=#{resource[:source]}", "destination_type=#{resource[:destination_type]}", "destination=#{resource[:dest]}", "properties_key=#{resource[:routing_key]}")
+    rabbitmqadmin('delete', 'binding', vhost_opt, "--user=#{resource[:user]}", "--password=#{resource[:password]}", '-c', '/etc/rabbitmq/rabbitmqadmin.conf', "source=#{resource[:source]}", "destination_type=#{resource[:destination_type]}", "destination=#{resource[:destination]}", "properties_key=#{resource[:routing_key]}")
     @property_hash[:ensure] = :absent
   end
 
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -19,7 +19,7 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
     if @should_vhost
       @should_vhost
     else
-      @should_vhost = resource[:name].split('@').last
+      @should_vhost = resource[:vhost]
     end
   end
 
@@ -51,11 +51,14 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
         end
         unless(source_name.empty?)
           binding = {
+            :source           => source_name,
+            :dest             => destination_name,
+            :vhost            => vhost,
             :destination_type => destination_type,
             :routing_key      => routing_key,
             :arguments        => JSON.parse(arguments),
             :ensure           => :present,
-            :name             => "%s@%s@%s" % [source_name, destination_name, vhost],
+            :name             => "%s@%s@%s@%s" % [source_name, destination_name, vhost, routing_key],
           }
           resources << new(binding) if binding[:name]
         end
@@ -65,9 +68,9 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
   end
 
   def self.prefetch(resources)
-    packages = instances
+    bindings = instances
     resources.keys.each do |name|
-      if provider = packages.find{ |pkg| pkg.name == name }
+      if provider = bindings.find{ |route| route.source == source && route.dest == dest && route.vhost == vhost && route.routing_key == routing_key }
         resources[name].provider = provider
       end
     end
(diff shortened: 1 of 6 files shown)


lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -51,11 +51,13 @@ def self.instances
...
         if (source_name != '')
           binding = {
+            :source           => source_name,
+            :dest             => destination_name,
+            :vhost            => vhost,
             :destination_type => destination_type,
             :routing_key      => routing_key,
             :arguments        => JSON.parse(arguments),
             :ensure           => :present,
-            :name             => "%s@%s@%s" % [source_name, destination_name, vhost],
```
hunner: This should be `"%s@%s@%s@%s" % [source_name, destination_name, vhost, routing_key],`

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -51,11 +51,13 @@ def self.instances
...
           }
           resources << new(binding) if binding[:name]
```
hunner: This needs fixing

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -51,11 +51,13 @@ def self.instances
...
         end
```
hunner: Prefetch should match on all 4 properties, not on `.name` like

`if provider = bindings.find { |route| route.source == source && route.dest == dest && ... }`

lib/puppet/type/rabbitmq_binding.rb
```diff
@@ -11,21 +11,48 @@
...
+
+  newparam(:dest, :namevar => true) do
+    desc 'destination of binding'
+    newvalues(/^\S+$/)
+  end
+
+  newparam(:vhost, :namevar => true) do
+    desc 'vhost'
+    newvalues(/^\S+$/)
+    defaultto('/')
```
hunner: Why does a required (because namevar) property have a defaultto? Maybe not needed?

hunner: Comment: keep the default because when explicitly specifying source/dest/routing_key and not using title_pattern, then it's good to have a default.

fatmcgav:
Any progress on this one?

wyardley:
@fatmcgav Sorry for the delayed response, and also to not having updated this PR in forever. I got some good pointers from @hunner and @DavidS, but honestly haven't had the energy to jump back into this so far, hopefully soon. I still do have the use case (though would be just as happy if someone else wanted to take a stab at it).

I've rebased against current master, made some of the suggested changes, and am taking a quick look to see if it works (some of the existing tests for 'name' patterns would need to be dropped, but perhaps there should be some additional tests added).

I've made most of the suggestions @hunner had in the review, but I'm still not getting it to work, whether or not I try to do it in a backwards-compatible way or not, so I think there are still some bits of the ever-fun title_patterns / composite namevars that are eluding me. I know some people have said composite namevars are not necessary, but I think maybe they are?

I briefly got it to work for one use case in Vagrant, but right now I'm getting stuck on:
```
Error: Failed to apply catalog: undefined method `source' for (provider=rabbitmqadmin):Puppet::Type::Rabbitmq_binding::ProviderRabbitmqadmin
```
which seems to be similar to what's described in
https://projects.puppetlabs.com/issues/15046
(so maybe title_patterns is somehow not matching)?

These are my test cases:

```
      rabbitmq_binding { 'exchange1@queue1@host1':
        user             => 'dan',
        password         => 'bar',
        destination_type => 'queue',
        routing_key      => '#',
        ensure           => present,
      }
```
```
      rabbitmq_binding { 'test1':
        source           => 'exchange1',
        dest             => 'queue1',
        vhost            => 'host1',
        user             => 'dan',
        password         => 'bar',
        destination_type => 'queue',
        routing_key      => 'test3',
        ensure           => present,
      }

      rabbitmq_binding { 'test2':
        source           => 'exchange1',
        dest             => 'queue1',
        vhost            => 'host1',
        user             => 'dan',
        password         => 'bar',
        destination_type => 'queue',
        routing_key      => 'test4',
        ensure           => present,
      }
```

DavidS:
The error is coming (presumably) from https://github.com/voxpupuli/puppet-rabbitmq/pull/504/files#diff-6848b732a3c79af87de0ad6905d8387eR73 and I don't think that can work like this. Print out what the `name` is there, maybe that gives you a clue on how to build that matching there.

It's been a long time since I played around with that part.

fatmcgav:
OK, so I've been able to re-produce the failure in a unit test. 
See https://github.com/fatmcgav/puppetlabs-rabbitmq/commit/1dfa47b74d22f70c70e8de3d2fc02f5755244446
which results in this failure:
```
Failures:

  1) Puppet::Type::Rabbitmq_binding::ProviderRabbitmqadmin#prefetch matches
     Failure/Error: if provider = bindings.find{ |route| route.source == source && route.dest == dest && route.vhost == vhost && route.routing_key == routing_key }

     NoMethodError:
       undefined method `source' for (provider=rabbitmqadmin):Puppet::Type::Rabbitmq_binding::ProviderRabbitmqadmin
     # ./lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb:73:in `block (2 levels) in prefetch'
     # ./lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb:73:in `each'
     # ./lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb:73:in `find'
     # ./lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb:73:in `block in prefetch'
     # ./lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb:72:in `each'
     # ./lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb:72:in `prefetch'
     # ./spec/unit/puppet/provider/rabbitmq_binding/rabbitmqadmin_spec.rb:113:in `block (3 levels) in <top (required)>'

Deprecation Warnings:

Using `should` from rspec-expectations' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` with `config.expect_with(:rspec) { |c| c.syntax = :should }` instead. Called from /Users/gavinw/Puppet/modules/puppetlabs/puppetlabs-rabbitmq/spec/unit/puppet/provider/rabbitmq_binding/rabbitmqadmin_spec.rb:29:in `block (3 levels) in <top (required)>'.


If you need more of the backtrace for any of these deprecations to
identify where to make the necessary changes, you can configure
`config.raise_errors_for_deprecations!`, and it will turn the
deprecation warnings into errors, giving you the full backtrace.

1 deprecation warning total

Finished in 0.02138 seconds (files took 0.95047 seconds to load)
8 examples, 1 failure

Failed examples:

rspec ./spec/unit/puppet/provider/rabbitmq_binding/rabbitmqadmin_spec.rb:93 # Puppet::Type::Rabbitmq_binding::ProviderRabbitmqadmin#prefetch matches
```

That makes it a bit easier to debug... I'll see if can spot the cause later...

wyardley:
@fatmcgav @DavidS thanks for the help, was reading up on debugging types / providers yesterday, but these will both help a lot. Thanks also for creating a unit test @fatmcgav -- also, I'm going to merge your tests to my branch (with attribution), hope that's Ok?

wyardley:
@hunner @DavidS if there's time, can you discuss this one in the meeting?
I think my updated version (thanks to some big help from @fatmcgav) is functional.

Probably should add some additional acceptance tests, but this is getting a lot closer.

wyardley:
I am curious if there's a way to validate whether we have a valid source / destination / vhost / routing_key combination in total, between defaults and splitting up the name in the legacy case, but not sure what the cleanest way to do that validation would be.

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
hunner: Done.

lib/puppet/type/rabbitmq_binding.rb
```diff
@@ -11,21 +11,68 @@
...
+    ]
+  end
+
+  newparam(:name) do
+    desc 'resource name, either source@destination@vhost or arbitrary name with params'
+
+    isnamevar
+  end
+
+  newparam(:source) do
```
hunner: source/destination/vhost/routing_key should be properties because they can be read from & written to the system's configuration. This will also allow them to be printed when running `puppet resource rabbitmq_binding`

hunner: Also the global `validate do` block can check `raise ArgumentError, "source is required" if ! self[:source] and ! self.provider.source` etc. to ensure that the values are actually passed.

wyardley:
Other questions: should routing_key default to `#`, and does it need to be defined? Can it / should it be an empty string if it's not used, and would this mess up all our logic?

lib/puppet/type/rabbitmq_binding.rb
wyardley: I updated those, as well as a couple other things that are read from / written to the config to be properties, let me know how that looks. Will work on validation.

wyardley:
ps - I took out the default for `routing_key`, which I think is closer to the existing behavior.

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
wyardley: so, like this? ```        sha256 = Digest::SHA256.new
        hashed_name = sha256.digest "%s@%s@%s@%s" % [source_name, destination_name, vhost, routing_key```

lib/puppet/type/rabbitmq_binding.rb
wyardley: I did now have to update title_patterns to match `:name` where I didn't before, not sure why that is. With the `a@b@c` style naming, I got a name error

lib/puppet/type/rabbitmq_binding.rb
```diff
@@ -11,23 +11,70 @@
...
+
+    newvalues(/^\S+$/)
+    isnamevar
+  end
+
+  newproperty(:destination, :namevar => true) do
+    desc 'destination of binding'
+
+    newvalues(/^\S+$/)
+    isnamevar
```
fatmcgav: Is this needed aswell as the `:namevar => true`?

wyardley: Good point, they're redundant.

wyardley:
TODO: global validate do block

lib/puppet/type/rabbitmq_binding.rb
wyardley:
```
[root@centos-7-x64 tmp]# puppet apply apply_manifest.pp.1fJLTF
[...]
Notice: /Stage[main]/Main/Rabbitmq_exchange[exchange1@host1]/ensure: created
Notice: /Stage[main]/Main/Rabbitmq_binding[binding 1]/ensure: created
Notice: /Stage[main]/Main/Rabbitmq_binding[binding 2]/ensure: created
Notice: Applied catalog in 4.25 seconds
[root@centos-7-x64 tmp]# puppet resource rabbitmq_binding
rabbitmq_binding { '3716cc442c568f6be3772365f29af0ca786ea852b527e3c564dcb3cd83f928a9':
  ensure           => 'present',
  destination      => 'queue1',
  destination_type => 'queue',
  routing_key      => 'test2',
  source           => 'exchange1',
  vhost            => 'host1',
}
rabbitmq_binding { 'b0f96dde732ed9be309bb466b680faa49117a151716a524eceaeb91080a34d64':
  ensure           => 'present',
  destination      => 'queue1',
  destination_type => 'queue',
  routing_key      => 'test1',
  source           => 'exchange1',
  vhost            => 'host1',
}
```

wyardley:
@hunner @fatmcgav I think this is good to go maybe!?

Technically, the user / password are managed by Puppet but not this resource, so am I right in thinking they should be params not properties? A lot of the things that should have been properties were params before, here's  how I have it now:
```
param: name
property source
property destination
property vhost
property routing_key
property destination_type
property arguments
param user
param password
```

PR #506, opened 2016-09-20
Allow Configuring Management IP Binding

rfkrocktk:
This change allows users to change which IP address that the management web-server binds on.

This makes it possible to bind the management console on the local interface only, so that it can be proxied by e.g. NGINX and not be facing the public.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -18,6 +18,7 @@ class rabbitmq::config {
   $env_config_path            = $rabbitmq::env_config_path
   $erlang_cookie              = $rabbitmq::erlang_cookie
   $interface                  = $rabbitmq::interface
+  $management_ip              = $rabbitmq::management_ip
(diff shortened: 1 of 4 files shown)


fatmcgav:
@naftulikay Useful addition, looks like it needs a rebase though. Could you do the honours?

naftulikay:
@fatmcgav Rebased.

wyardley:
Just curious, isn't this what `$interface` and `$ssl_interface` allow you to do? 
https://www.rabbitmq.com/networking.html
I think you can just set `$interface` to `127.0.0.1`, no?

That said, using that old `'UNSET'` pattern seems kind of ugly, I should throw together a PR to use `undef` instead.

wyardley:
Oh I get it, binding the management interface to a different interface from the service itself.

PR #507, opened 2016-10-01
set $real_package_source to undef instead of empty

bastelfreak:
This param is checked in install.pp:
https://github.com/puppetlabs/puppetlabs-rabbitmq/blob/master/manifests/install.pp#L19

The check doesn't work if the variable contains empty string.


binford2k:
The spec failure looks unrelated.

bastelfreak:
yep, seems to be a ruby/rspec bug.

PR #508, opened 2016-10-03
fix test breakage

wyardley:
I'm not sure if this is intentional, but https://github.com/puppetlabs/puppetlabs-stdlib/commit/970852dd317fbb2699b406dd25aeddef496a7c92
seems to break the acceptance tests. @dhollinger helpfully suggested this fix; if stdlib is not going to be changed in a way that is backwards compatible for tests to at least run, we'll probably need to pin stdlib until the tests can be changed (maybe as part of the 6.0 release?)

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,9 @@
 fixtures:
+  forge_modules:
+    stdlib:
+      repo: "puppetlabs/stdlib"
+      ref: "4.12.0"
   repositories:
-    "stdlib": "git://github.com/puppetlabs/puppetlabs-stdlib.git"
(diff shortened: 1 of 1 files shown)


dhollinger:
This probably is no longer needed as the issue has been fixed in puppetlabs/puppetlabs-stdlib#662 and puppetlabs/puppetlabs-stdlib#663

If you do want to test only against stable releases, it might be better/safer to configure `.fixtures.yml` to pull from the Puppet Forge than from a Git Ref.

wyardley:
Agreed that this is probably no longer needed now that the bug has been fixed (wasn't sure if it was a bug or intentional behavior)

PR #512, opened 2016-10-20
Bug fix for when queue names include spaces

Bubbad:
Queue arguments and properties are now parsed by tabs instead of spaces.


wyardley:
This might actually pass with the current set of Travis tests (I think someone should be able to trigger off a new run, but force-pushing the branch might also work).

Are they reliably tab-delimited? Would it be possible to add a test case for when there's a space in the queue name?

PR #529, opened 2017-02-01
Add HTTP API provider for rabbitmq

alexizmailov:
This request adds new provider for managing rabbitmq  server.
It uses HTTP API, rabbitmq http gem from https://github.com/ruby-amqp/rabbitmq_http_api_client and significantly speeds up puppet runs. Currently the following features are implemented:
- rabbitmq_user
- rabbitmq_vhost
- rabbitmq_user_permission
- rabbitmq_policy
- rabbitmq_exchange
- HTTP connections are currently supported, however TLS has not yet been implemented

Performance metrics:
- rabbitmqctl provider
```
real 4m58.141s 
user 3m47.098s 
sys 1m23.255s 
```
- rabbitmqhttp provider
```
real 0m58.635s 
user 0m33.745s 
sys 0m5.756s
```

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -4,5 +4,6 @@ fixtures:
     "apt": "git://github.com/puppetlabs/puppetlabs-apt.git"
     "staging": "git://github.com/voxpupuli/puppet-staging.git"
     "erlang:": "git://github.com/garethr/garethr-erlang.git"
+    "rabbitmq_http_api_client": "https://github.com/ruby-amqp/rabbitmq_http_api_client"
   symlinks:
     "rabbitmq": "#{source_dir}"
--- a/lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin_http.rb
+++ b/lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin_http.rb
@@ -1,0 +1,112 @@
+require 'puppet'
+
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
+
+Puppet::Type.type(:rabbitmq_exchange).provide(:rabbitmqhttp, :parent => Puppet::Type.type(:rabbitmq_exchange).provider(:rabbitmqctl)) do
+
+
+  defaultfor :feature => :posix
+
+  def should_vhost
+    if @should_vhost
+      @should_vhost
+    else
+      @should_vhost = resource[:name].split('@')[1]
+    end
+  end
+
+  def self.all_vhosts
+    vhosts = []
+    vhost_list = $client.list_vhosts
+    vhost_list.each do |vhost|
+      vhosts.push(vhost[:name])
+    end
+    vhosts
+  end
+
+  def self.all_exchanges(vhost)
+    exchanges = []
+    exchange_list = $client.list_exchanges(vhost)
+    exchange_list.each do |exchange|
+      next if exchange.name.eql? ''
+      json_args = exchange.arguments.to_json
+      tmp_exchange = "#{exchange.name} #{exchange.type} #{exchange.internal} #{exchange.durable} #{exchange.auto_delete} #{json_args}"
+      exchanges.push(tmp_exchange)
+    end
+    exchanges
+  end
+
+  def self.instances
+    resources = []
+    all_vhosts.each do |vhost|
+        all_exchanges(vhost).each do |line|
+          name, type, internal, durable, auto_delete, arguments = line.split()
+            if type.nil?
+                # if name is empty, it will wrongly get the type's value.
+                # This way type will get the correct value
+                type = name
+                name = ''
+            end
+            # Convert output of arguments from the rabbitmqctl command to a json string.
+            if !arguments.nil?
+              arguments = arguments.gsub(/^\[(.*)\]$/, "").gsub(/\{("(?:.|\\")*?"),/, '{\1:').gsub(/\},\{/, ",")
+              if arguments == ""
+                arguments = '{}'
+              end
+            else
+              arguments = '{}'
+            end
+            exchange = {
+              :type   => type,
+              :ensure => :present,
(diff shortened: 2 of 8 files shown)


wyardley:
I would love if , but I think it's a great idea, do the acceptance tests pass for you locally? Is this something that you're still interested in seeing get merged at some point?

alexizmailov:
Hi.

It's been working in our environment for the last 7 months and no issues were noticed.

wyardley:
I don't think there's a clean way to ensure that the gem you're using will get installed, so may be hard to incorporate this into the module, as much as I love the idea. Maybe some other folks will have an idea, but this is failing the acceptance tests, and would presumably fail on any system that doesn't have https://github.com/ruby-amqp/rabbitmq_http_api_client gem installed. Potentially we could use gem or something else to install it on the system in the manifests, but this would add some new dependencies and potential problems.

```       	
Error: Could not autoload puppet/provider/rabbitmq_exchange/rabbitmqadmin_http: cannot load such file -- rabbitmq/http/client

       	Error: Could not autoload puppet/type/rabbitmq_exchange: Could not autoload puppet/provider/rabbitmq_exchange/rabbitmqadmin_http: cannot load such file -- rabbitmq/http/client

       	Error: Evaluation Error: Error while evaluating a Virtual Query, Could not autoload puppet/type/rabbitmq_exchange: Could not autoload puppet/provider/rabbitmq_exchange/rabbitmqadmin_http: cannot load such file -- rabbitmq/http/client at /etc/puppetlabs/code/modules/rabbitmq/manifests/init.pp:196:52 on node ubuntu-1404-x64.c.eco-emissary-99515.internal
```

lib/puppet/provider/rabbitmqhttp.rb
```diff
@@ -0,0 +1,12 @@
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
+
+class Puppet::Provider::Rabbitmqhttp < Puppet::Provider
```
hunner: Nothing appears to use this.

.fixtures.yml
```diff
@@ -4,5 +4,6 @@ fixtures:
     "apt": "git://github.com/puppetlabs/puppetlabs-apt.git"
     "staging": "git://github.com/voxpupuli/puppet-staging.git"
     "erlang:": "git://github.com/garethr/garethr-erlang.git"
+    "rabbitmq_http_api_client": "https://github.com/ruby-amqp/rabbitmq_http_api_client"
```
hunner: Interesting. I've never thought about adding a gem as a modules fixture. I think this should just be handled via the Gemfile though, and should not be put here.

lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin_http.rb
```diff
@@ -0,0 +1,112 @@
+require 'puppet'
+
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
```
hunner: No global variables please.

lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin_http.rb
```diff
@@ -0,0 +1,112 @@
...
+      tmp_exchange = "#{exchange.name} #{exchange.type} #{exchange.internal} #{exchange.durable} #{exchange.auto_delete} #{json_args}"
+      exchanges.push(tmp_exchange)
+    end
+    exchanges
+  end
+
+  def self.instances
+    resources = []
+    all_vhosts.each do |vhost|
+        all_exchanges(vhost).each do |line|
```
hunner: Indentation.

lib/puppet/provider/rabbitmq_policy/rabbitmqhttp.rb
```diff
@@ -0,0 +1,102 @@
+require 'json'
+require 'puppet/util/package'
+
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
```
hunner: Same.

lib/puppet/provider/rabbitmq_user/rabbitmqhttp.rb
```diff
@@ -0,0 +1,98 @@
+require 'puppet'
+require 'set'
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
```
hunner: Same.

lib/puppet/provider/rabbitmq_user_permissions/rabbitmqhttp.rb
```diff
@@ -0,0 +1,109 @@
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
```
hunner: Same.

lib/puppet/provider/rabbitmq_vhost/rabbitmqhttp.rb
```diff
@@ -0,0 +1,31 @@
+require "rabbitmq/http/client"
+
+$endpoint = "http://localhost:15672"
+$client = RabbitMQ::HTTP::Client.new($endpoint, :username => "guest", :password => "guest")
```
hunner: Same.

wyardley:
FWIW, there is some precedent for requiring gems after all, so if the idea would be to convert everything to use the API only, and make that the sole provider for everything, it's definitely an interesting idea (but also a big project).

PR #532, opened 2017-02-21
moved username:password to separate parameter

vdmkenny:
Having username:password@hostname gives an error when the password contains an "@", as in my case.


vdmkenny:
This fixes an issue where the code would fail if the password contains an "@" character.

vdmkenny:
Fixed my tests, sorry about that.

wyardley:
This seems sane to me (though as @hunner pointed out, should probably switch to `archive` module)

PR #533, opened 2017-02-23
Set configuration files to rabbitmq owner/group

Spredzy:
Currently the module has a hardcoded value of '0' for RabbitMQ
configuration files. This matter should ideally be left to packager,
however since it could still be possible to set the configuration we
should default to rabbitmq user for owner and group.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -79,15 +79,15 @@ class rabbitmq::config {
 
   file { '/etc/rabbitmq':
     ensure => directory,
-    owner  => '0',
-    group  => '0',
+    owner  => $rabbitmq_user,
+    group  => $rabbitmq_group,
     mode   => '0644',
   }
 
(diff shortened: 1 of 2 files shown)


hunner:
This doesn't appear to be a bugfix, so it's probably backwards incompatible. Is there something that breaks by leaving it uid/gid 0?

wyardley:
I can see the benefit of having rabbitmq run as an unprivileged user, but I'd think it's actually a better practice to leave the config files owned by root as they are now, unless there's some specific need to have them writable by the rabbitmq user.

wyardley:
See also my comment in #454, though. 

What about making the config 0640 `root:$rabbitmq_user`? This should be backwards compatible but also resolve the issues brought up there, namely: directories 0644 instead of 0755, and that credentials are stored in a world-readable file?

hunner:
Closed in favor of #560

PR #536, opened 2017-03-08
	replace validate_* with datatypes

bastelfreak:


bastelfreak:
There are a few strings that could also be converted to Integers. Let me know if I should do this as well. The goal of this PR is to get rid of all the validate_* function calls. They are too noisy since Puppet deprecated them.

benh57:
This module still warns about the validate_re() calls.  can you fix those as well?

benh57:
Warning: This method is deprecated, please use the stdlib validate_legacy function, with Pattern[]. There is further documentation for validate_legacy function in the README. at [".../rabbitmq/manifests/init.pp", 85]:

PR #542, opened 2017-04-22
add FreeBSD support

olevole:


spec/classes/rabbitmq_spec.rb
```diff
@@ -1384,6 +1384,17 @@
     end
   end
 
+  context "on FreeBSD" do
+    with_freebsd_facts
+    it 'installs the rabbitmq package' do
+      should contain_package('rabbitmq-server').with(
+        'ensure'   => 'installed',
+        'name'     => 'rabbitmq',
+        'provider' => 'freebsd'
```
wyardley: This is why your test is failing, it's expecting pkgng.
Though I think provider isn't actually needed. Not sure if #493 will get merged, but would have to adjust slightly to get FreeBSD support working if / when that gets merged.

manifests/params.pp
```diff
@@ -38,6 +38,17 @@
       $rabbitmq_home    = '/var/rabbitmq'
       $plugin_dir       = '/usr/local/lib/rabbitmq/plugins'
     }
+    'FreeBSD': {
+      $package_ensure   = 'installed'
+      $package_name     = 'rabbitmq'
+      $service_name     = 'rabbitmq'
+      $package_provider = 'pkgng'
+      $version          = '3.4.2'
```
wyardley: Also, keep in mind that while `$version` is used internally as things work now, it's ignored when using the system package provider, so it doesn't really matter in this case.

wyardley:
ps - https://github.com/voxpupuli/puppet-rabbitmq/blob/master/metadata.json may also need to be updated (FreeBSD and OpenBSD support are both not currently included there)

wyardley:
@olevole would you like to rework this one?

olevole:
@wyardley Hi Will! 
Thank you very much for being able to spare time and for comments. I took into account your wishes. Unfortunately, I now have a very gigantic problem with time at the moment, sooner or later I will be able to finish and re-send patch. If you can quickly fix my shortcomings - don't wait for me, I'll be only too glad to do it and ready to test module on my FreeBSD cluster's. I apologize for the delay.

wyardley:
I can take a stab at it once #493 is merged, shouldn't be hard.

wyardley:
@olevole ps - one question, though - is `pkgng` the default provider in Puppet on FreeBSD? I'm pretty familiar with FreeBSD (a user for years), but haven't run Puppet with it before.

wyardley:
Haven't submitted yet, but check out this:
https://github.com/voxpupuli/puppet-rabbitmq/compare/master...wyardley:freebsd?expand=1
Will update it after we merge #493.

wyardley:
reworking via #567

PR #547, opened 2017-06-08
Extra ssl options to harden rabbitmq listener

xepa:
This patch adds the following options to the ssl config to harden the rabbitmq ssl setup

- ssl_secure_renegotiate (boolean default true)
- ssl_reuse_sessions (boolean default true)
- ssl_honor_cipher_order (boolean default true)
- ssl_dhfile (string default empty)


hunner:
This PR will cause every user to get ssl_secure_renegotiate, ssl_reuse_sessions, and ssl_honor_cipher_order added to their configuration. This seems like a breaking change; what do you think?

hunner:
http://erlang.org/doc/man/ssl.html#id61579 documents that at least two of these default to `false` in erlang.

hunner:
I merged this by accident, and it looks like there is no way to reopen it :\. Sorry, but could you resubmit?

PR #549, opened 2017-06-21
Fix error text in `rabbitmq_vhost` provider

hybby:
Just a small thing that threw me when looking through the code.

This provider is definitely calling `rabbitmqctl -q list_vhosts`, so if that output can't be parsed it should complain about invalid vhost lines, rather than invalid user lines.

Looks like a copy-paste error from other similar providers.

This will improve user experience when they encounter this error.


hunner:
Failures unrelated to the PR change: https://travis-ci.org/voxpupuli/puppet-rabbitmq/jobs/245405735

hunner:
Reopened to retrigger tests.

PR #552, opened 2017-06-26
Add options for IPv6 and inetrc

jeckersb:
This adds a new parameter, ipv6, which if set to true will configure
the erlang runtime to use IPv6 for distribution and name resolution.

It also adds two other parameters, inetrc_config and
inetrc_config_path, which support ipv6 enablement.

The environment variable ERL_INETRC is set to the value of
inetrc_config_path, by default /etc/rabbitmq/inetrc.  If ipv6 is
disabled (default), this file is empty and has no effect.  If ipv6 is
enabled, it sets the option 'inet6' to 'true', which prefers IPv6 name
resolution[1].

Users may pass their own inetrc_config to supply an alternate
template.

In addition, if ipv6 is set to true, the flag '-proto_dist inet6_tcp'
is appended to the environment variables RABBITMQ_SERVER_ERL_ARGS and
RABBITMQ_CTL_ERL_ARGS.  This enables IPv6 as the distribution protocol.

[1] https://github.com/erlang/otp/commit/5d7dcfc2e0de7e93b29d01f07a2f970720d62f9d

### DIFF
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -542,26 +542,36 @@ LimitNOFILE=1234
       end
 
       describe 'rabbitmq-env configuration' do
-        let(:params) {{ :environment_variables => {
-          'NODE_IP_ADDRESS'    => '1.1.1.1',
-          'NODE_PORT'          => '5656',
-          'NODENAME'           => 'HOSTNAME',
-          'SERVICENAME'        => 'RabbitMQ',
-          'CONSOLE_LOG'        => 'RabbitMQ.debug',
-          'CTL_ERL_ARGS'       => 'verbose',
-          'SERVER_ERL_ARGS'    => 'v',
-          'SERVER_START_ARGS'  => 'debug'
-        }}}
-        it 'should set environment variables' do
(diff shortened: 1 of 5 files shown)


mbaldessari:
Tested this in my environment and I finally can correctly get rabbitmq up and working in my ipv6 environment.

jeckersb:
Any thoughts on this, good or bad?

EmilienM:
:+1: it seems like I can't merge PRs anymore on this module, probably since the move... Anyway, this is good to go - ping @bastelfreak

spec/classes/rabbitmq_spec.rb
```diff
@@ -1153,6 +1163,51 @@
...
+
+          it 'should enable inet6 distribution and quote properly' do
+            should contain_file('rabbitmq-env.config') \
+              .with_content(%r{^RABBITMQ_SERVER_ERL_ARGS="foo -proto_dist inet6_tcp"$}) \
+              .with_content(%r{^RABBITMQ_CTL_ERL_ARGS="bar -proto_dist inet6_tcp"$})
+          end
+
+        end
+      end
+
```
EmilienM: nice tests coverage, thanks!

PR #556, opened 2017-08-01
new example file

dshuvar:
I added new example file  which use functions rabbitmq_management and ssl certificates for rabbitMQ.


examples/rabbitmq_management_ssl.pp
```diff
@@ -0,0 +1,34 @@
+
+node 'rabbit.example.com' {
```
hunner: I think either the node has to be taken out, or the rest of the code has to be indented to match.

examples/rabbitmq_management_ssl.pp
```diff
@@ -0,0 +1,34 @@
...
+  ssl_versions          => ['tlsv1.2', 'tlsv1.1'],
+  environment_variables => {
+    'NODENAME'    => 'rabbit',
+    'SERVICENAME' => 'Rabbit',
+    'CONFIG_FILE' =>  '/etc/rabbitmq/rabbitmq'
+  }
+
+}
+
+rabbitmq_plugin {'rabbitmq_management':
```
hunner: Space after the open `{`

wyardley:
Now that we have docs in puppet-strings, we should probably rework / review the examples, and maybe move the ones that are helpful into puppet-strings format. Let me know if you're interested in helping w/ docs work in general.

We should also probably just make `rabbitmq::server` private, and update the docs that declare it that way.

wyardley:
I'm going to close this one for now, if there are additional examples, let's add them via Puppet strings, now that docs are auto-generated.

PR #559, opened 2017-08-16
Switch from 'UNSET' to undefs, rename management_ip to management_ip_address

wyardley:
switch to optional types and use undef instead of the string `'UNSET'` in various places.

Need to fix an issue with the template rendering

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,4 +1,12 @@
-# Main rabbitmq class
+# rabbitmq
+#
+# @summary Main class for rabbitmq module
+#
+# @example Declaring the class
+#   class { 'rabbitmq': }
+#
+# @param management_ip_address Allows you to set the IP for management interface to bind to separately. Set to 127.0.0.1 to bind to localhost only, or 0.0.0.0 to bind to all interfaces.
+# @param node_ip_address Allows you to set the IP for RabbitMQ service to bind to. Set to 127.0.0.1 to bind to localhost only, or 0.0.0.0 to bind to all interfaces.
 class rabbitmq(
   Boolean $admin_enable                          = $rabbitmq::params::admin_enable,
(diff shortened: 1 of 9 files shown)


wyardley:
seems like the issues with undef not being `nil` in the hash may be related to the `merge()` function from stdlib?

manifests/init.pp
```diff
@@ -92,7 +92,7 @@
     validate_re($management_port, '\d+')
   }
   if ! is_integer($port) {
-    validate_re($port, ['\d+','UNSET'])
+    validate_re($port, '\d+')
```
hunner: This would trigger when `$port` is `undef`

wyardley:
@naftulikay @fatmcgav 
Wondering how it makes sense to use `management_ip_address`  (updated this for consistency naming-wise with `node_ip_address`) vs `node_ip_address`, as well as what the behavior should be if `management_ip_address` is undef (i.e., should if default to the setting of `node_ip_address` or just not be set)?

I think may have unintentionally caused some new problems that the test didn't detect, because it will need to download rabbitmqadmin via the admin interface. Also not sure if `0.0.0.0` should be the default.

wyardley:
There's room  for more improvement, but did some logic where `$management_ip_address` will default to `$node_ip_address` if the former is undef but the latter is defined.

I added some basic spec tests, because it seems like there were no actual tests that checked that the service was listening on the configured port. Still room for improvement here, like stubbing in SSL and testing that the service comes up and is listening on the right ports (I did do a quick functional test to make sure that my changes to the templates here worked in the case of SSL as well).

I'm guessing these tests should go in `class_spec` vs. `server_spec` (the two seem to do basically the same things anyway, and I'm not sure if we really should support calling `rabbitmq::server` directly anyway?). Or should I make a new `config_spec` file?

PR #561, opened 2017-08-17
Fix regex double escaping of rabbitmqctl list_policies

wyardley:
This reworks #482, and keeps attribution, but adds the spec test that @DrOwl posted. cc: @eputnam @sspreitzer @devfaz


wyardley:
Fixed one other problem with the test. I don't have a strong personal need for this one, but seems sane enough if someone wants to merge it.

DrOwl:
Excellent thanks for submitting this in an orderly fashion

PR #562, opened 2017-08-18
Move examples and params to puppet strings style docs

wyardley:
This is a first pass at moving docs to puppet-strings, starting with params and examples. Only bad thing is that we might want to move some of the params that aren't os specific back to init.pp at some point, so that it's easier to see default params in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,6 @@
 3. [Setup - The basics of getting started with rabbitmq](#setup)
     * [What rabbitmq affects](#what-rabbitmq-affects)
     * [Setup requirements](#setup-requirements)
-    * [Beginning with rabbitmq](#beginning-with-rabbitmq)
 4. [Usage - Configuration options and additional functionality](#usage)
 5. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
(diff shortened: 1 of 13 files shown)


wyardley:
Let's also wait to merge this until some of the other changes are merged, don't want to create nasty rebase hell. Other than that, though, I think this should be pretty safe.

HAIL9000:
Aw shoot sorry, I just saw this PR I will make sure to look it over this week!

wyardley:
@HAIL9000 thanks! Ping me on Slack or IRC sometime if you want to discuss more (and I'll try to rebase as well). I think the big question is where the examples should go... I'm leaning towards that they should go in the types / providers they correspond with, but then I'm worried that they'll get buried unless there's some tooling to generate a README.

wyardley:
One other thing I noticed -- moving examples out of README doesn't let us do:
```
 ```puppet
```
as best I can tell. Is there a way (in puppet strings) to set the code type?

wyardley:
Rebased

wyardley:
@HAIL9000 how's this?

PR #563, opened 2017-08-18
Harden SSL Setup

xepa:
This patch adds the following options to the ssl config to harden the rabbitmq ssl setup

ssl_secure_renegotiate (boolean default true)
ssl_reuse_sessions (boolean default true)
ssl_honor_cipher_order (boolean default true)
ssl_dhfile (string default empty)

Cherry-Picking: 7b5afa9e489e7354fb5a6a32fa0d38f3a4e38dbc


michaelklishin:
👍.

As a general note for those landing to this PR from Google, there is a [section on TLS setup validation](http://www.rabbitmq.com/ssl.html) in RabbitMQ TLS guide. 3 of the options contributed here are necessary to pass most [testssl.sh](https://testssl.sh) tests, for example.

wyardley:
I like this.
I am working on switching from the `'UNSET'` pattern to using undefs (as well as possibly switching to puppet-strings for documenting params; see #559 #562), so depending if / when some of these other changes get merged, there might be  some rebasing necessary. I personally think it would be good to just switch to using `undef` vs `'UNSET'` here as well (and move it from params to init).

My biggest question is whether there maybe ought to be some tests added for this (also, there aren't acceptance tests yet to validate any of the ssl related configs as best I can tell).

manifests/params.pp
```diff
@@ -113,6 +113,10 @@
   $ssl_stomp_port              = '6164'
   $ssl_verify                  = 'verify_none'
   $ssl_fail_if_no_peer_cert    = false
+  $ssl_secure_renegotiate      = true
+  $ssl_reuse_sessions          = true
+  $ssl_honor_cipher_order      = true
+  $ssl_dhfile                  = 'UNSET'
```
wyardley: Even though this matches the existing pattern, I think `undef` would be a better way to do this (and apparently putting `undef`s in `params.pp` may be an anti-pattern, so maybe just add above

manifests/init.pp
```diff
@@ -55,6 +55,10 @@
   $ssl_stomp_port                                = $rabbitmq::params::ssl_stomp_port,
   $ssl_verify                                    = $rabbitmq::params::ssl_verify,
   $ssl_fail_if_no_peer_cert                      = $rabbitmq::params::ssl_fail_if_no_peer_cert,
+  Boolean $ssl_secure_renegotiate                = $rabbitmq::params::ssl_secure_renegotiate,
+  Boolean $ssl_reuse_sessions                    = $rabbitmq::params::ssl_reuse_sessions,
+  Boolean $ssl_honor_cipher_order                = $rabbitmq::params::ssl_honor_cipher_order,
+  String $ssl_dhfile                             = $rabbitmq::params::ssl_dhfile,
```
wyardley: maybe `Optional[String] $ssl_dhfile                             = undef,` ?

wyardley:
@xepa if you don't mind, I'm going to rework this, along with tests, rebased against current master and switched to undefs.

wyardley:
Closing in favor of #632

PR #567, opened 2017-08-25
Add FreeBSD support

wyardley:
replaces #542

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -36,6 +36,12 @@
         "12.04",
         "14.04"
       ]
+    },
+    {
+      "operatingsystem": "FreeBSD"
+    },
+    {
+      "operatingsystem": "OpenBSD"
     }
   ],
   "requirements": [
--- a/spec/spec_helper_local.rb
+++ b/spec/spec_helper_local.rb
@@ -38,11 +38,11 @@ def with_freebsd_facts
   # operatingsystemrelease may contain X.X-current
   # or other prefixes
   let :facts do
-    super().merge({
-      :kernelversion             => '12',
-      :osfamily                  => 'FreeBSD',
-      :staging_http_get          => '',
-    })
+    super().merge(
+      kernelversion: '12',
+      osfamily: 'FreeBSD',
+      staging_http_get: ''
+    )
   end
 end
 
--- a/spec/spec_helper_local.rb
+++ b/spec/spec_helper_local.rb
@@ -33,6 +33,19 @@ def with_openbsd_facts
   end
 end
 
+def with_freebsd_facts
+  # operatingsystemmajrelease is too broad
+  # operatingsystemrelease may contain X.X-current
+  # or other prefixes
+  let :facts do
+    super().merge({
+      :kernelversion             => '12',
+      :osfamily                  => 'FreeBSD',
+      :staging_http_get          => '',
+    })
+  end
+end
+
 def with_redhat_facts
   let :facts do
     super().merge(operatingsystemmajrelease: '7',
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -33,12 +33,9 @@ class rabbitmq::params {
       $package_ensure   = 'installed'
       $package_name     = 'rabbitmq'
       $service_name     = 'rabbitmq'
-      $package_provider = 'pkgng'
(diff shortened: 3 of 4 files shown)


wyardley:
@olevole check out this one if you get a chance.

olevole:
@wyardley Thanks! Looks good for me and my FBSD:

~~
rabbit1:/root@[13:52] # uname -a
FreeBSD rabbit1.my.domain 12.0-CURRENT FreeBSD 12.0-CURRENT #0: Fri Aug 25 21:51:50 MSK 2017     root@samson.bsdstore.ru:/usr/obj/usr/jails/src/src_12/src/sys/CBSD  amd64

rabbit1:/root@[13:53] # puppet --version
4.10.6

rabbit1:/root@[13:53] # ps axf |grep erlan
32156  -  SJ   0:00.02 /usr/local/lib/erlang19/erts-8.3.5.2/bin/epmd -daemon
40005  -  IJ   0:05.10 /usr/local/lib/erlang19/erts-8.3.5.2/bin/beam.smp -W w -A 128 -P 1048576 -t 5000000 -stbt db -zdbbl 32000 -B i -- -root /usr/local/lib/erlang19 -progname erl -- -home /var/db/rabbitmq -- -pa /usr/local/lib/erlang/lib/rabbitmq_server-3.6.10/ebin -noshell -noinput -s rabbit boot -sname rabbit@rabbit1 -boot start_sasl -config /usr/local/etc/rabbitmq/rabbitmq -kernel inet_default_connect_options [{nodelay,true}] -sasl errlog_type error -sasl sasl_error_logger false -rabbit error_logger {file,"/var/log/rabbitmq/rabbit@rabbit1.log"} -rabbit sasl_error_logger {file,"/var/log/rabbitmq/rabbit@rabbit1-sasl.log"} -rabbit enabled_plugins_file "/usr/local/etc/rabbitmq/enabled_plugins" -rabbit plugins_dir "/usr/local/lib/erlang/lib/rabbitmq_server-3.6.10/plugins" -rabbit plugins_expand_dir "/var/db/rabbitmq/mnesia/rabbit@rabbit1-plugins-expand" -os_mon start_cpu_sup false -os_mon start_disksup false -os_mon start_memsup false -mnesia dir "/var/db/rabbitmq/mnesia/rabbit@rabbit1" -kernel inet_dist_listen_min 25672 -kernel inet_dist_listen_max 25672 inet_dist_listen_max 25672

~~

metadata.json
```diff
@@ -36,6 +36,12 @@
         "12.04",
         "14.04"
       ]
+    },
+    {
+      "operatingsystem": "FreeBSD"
```
bastelfreak: freebsd isnt rolling release (well, kind of), so we shul list the versions we support. same for openbsd

spec/spec_helper_local.rb
```diff
@@ -33,6 +33,19 @@ def with_openbsd_facts
   end
 end
 
+def with_freebsd_facts
+  # operatingsystemmajrelease is too broad
```
bastelfreak: do we mock stuff here that is totally different in the real world, depending on the freebsd trouble? does it bring us into trouble?

metadata.json
wyardley: Yeah, I wasn't sure which versions this would work with, and was mostly relying on examples from other modules. Because of the ports system, my guess is that the range of present and future supported versions might be pretty broad, though I don't know how far back we'd want to support.

I do use FreeBSD and OpenBSD (FBSD since the 4.x days), but not with Puppet, so I didn't really know what values are reasonable here. Suggestions or PR to my PR are welcome 😉

spec/spec_helper_local.rb
wyardley: Not quite sure I follow?

bastelfreak: are the facts we simulate here used? If so, we should probably simulate them correctly, for all supported freebsd versions (or rely on rspec-puppet-facts)

wyardley: ps - reformatted this one. I think they are (like the other OSes in this file) valid, but not the only possible value.
For example, we support CentOS 6, but only supply facts for 7.

PR #568, opened 2017-08-25
convert to "expect" spec syntax (auto-converted via transpec)

wyardley:
I used transpec to convert spec tests to expect syntax. This should be merged soon or it will get out of date, and it will of course cause some hassles for existing PRs in terms of rebasing.


wyardley:
Closing since this is getting pulled into modulesync changes

PR #569, opened 2017-08-26
modulesync, rubocop -a, and other minor syntax / lint fixes

wyardley:
Here's a first pass at doing a modulesync.

Would be happy if someone could review (and help with fixing remaining warnings or providing some guidance on fixing them). We're down from > 3000 rubocop warnings to ~ 70. The "multiple expectations" ones, I can probably sort out how to adjust, but there are some others that I could use another pair of eyes on.

Not sure on the best way to replace or include the stuff in `spec_helper_local.rb`. 

Added `docker_sets` to `.sync.yml`, tried running another modulesync but didn't seem to add to `.travis.yml`.  I think someone else will have to generate the secret for Forge deploy.

Acceptance tests seem to be passing.

I would propose we postpone most other PRs until this is merged.


wyardley:
Closing in favor of #575

Issue #566, opened 2017-08-23
Enabling new plugin fails


rmestrum:
When i try to enable a extra plugin, the plugin is enabled bot not running.

When executing the command that the provider uses (`/usr/lib/rabbitmq/bin/rabbitmq-plugins`) manually an error shows up

```
[root@rmq ~]# rabbitmq-plugins list
 Configured: E = explicitly enabled; e = implicitly enabled
 | Status:   * = running on rabbit@rmq
 |/
[e*] amqp_client                       3.6.5
[  ] cowboy                            1.0.3
[  ] cowlib                            1.0.1
[e*] mochiweb                          2.13.1
[  ] rabbitmq_amqp1_0                  3.6.5
[  ] rabbitmq_auth_backend_ldap        3.6.5
[  ] rabbitmq_auth_mechanism_ssl       3.6.5
[  ] rabbitmq_consistent_hash_exchange 3.6.5
[  ] rabbitmq_event_exchange           3.6.5
[  ] rabbitmq_federation               3.6.5
[  ] rabbitmq_federation_management    3.6.5
[  ] rabbitmq_jms_topic_exchange       3.6.5
[E*] rabbitmq_management               3.6.5
[e*] rabbitmq_management_agent         3.6.5
[  ] rabbitmq_management_visualiser    3.6.5
[  ] rabbitmq_mqtt                     3.6.5
[  ] rabbitmq_recent_history_exchange  1.2.1
[  ] rabbitmq_sharding                 0.1.0
[  ] rabbitmq_shovel                   3.6.5
[  ] rabbitmq_shovel_management        3.6.5
[  ] rabbitmq_stomp                    3.6.5
[  ] rabbitmq_top                      3.6.5
[  ] rabbitmq_tracing                  3.6.5
[  ] rabbitmq_trust_store              3.6.5
[e*] rabbitmq_web_dispatch             3.6.5
[  ] rabbitmq_web_stomp                3.6.5
[  ] rabbitmq_web_stomp_examples       3.6.5
[  ] sockjs                            0.3.4
[e*] webmachine                        1.10.3
[root@rmq ~]# /usr/lib/rabbitmq/bin/rabbitmq-plugins enable rabbitmq_web_stomp
The following plugins have been enabled:
  rabbitmq_stomp
  cowlib
  cowboy
  sockjs
  rabbitmq_web_stomp

Applying plugin configuration to rabbit@rmq... failed.
 * Could not contact node rabbit@rmq.
[...]

Issue #134, opened 2013-10-17
EPEL RPM does not install /usr/sbin/rabbitmq-plugins


someword:
I ran into an issue using this module on centos 6.X as the provider would not load because /usr/sbin/rabbitmq-plugins did not exist.  I noticed that the rpm that comes from EPEL does not include /usr/sbin/rabbitmq-plugins where as the one from rabbitmq does.

I opened https://bugzilla.redhat.com/show_bug.cgi?id=1020447 to find out if the missing binary is by design or by accident.  If it's by design we would need to add the following resources.
```
  file {
    '/usr/sbin/rabbitmq-plugins':
      ensure  => link,
      target  => '/usr/lib/rabbitmq/bin/rabbitmq-plugins',
      require => Package['rabbitmq-server'],;
    '/usr/sbin/rabbitmq-env':
      ensure  => link,
      target  => '/usr/lib/rabbitmq/bin/rabbitmq-env',
      require => Package['rabbitmq-server'],;
  }
```

wyardley:
This did seem to get handled (differently) in the plugin provider, though I have been thinking of taking it back out.

I think this is good now, though:
```
[root@centos-7-x64 ~]# rpm -qi rabbitmq-server | egrep "^(Version|Vendor)"
Version     : 3.3.5
Vendor      : Fedora Project
[root@centos-7-x64 ~]# yum info rabbitmq-server
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: repos.dfw.quadranet.com
 * epel: mirror.hmc.edu
 * extras: mirror.hmc.edu
 * updates: mirror.hmc.edu
Installed Packages
Name        : rabbitmq-server
Arch        : noarch
Version     : 3.3.5
Release     : 34.el7
Size        : 4.8 M
Repo        : installed
From repo   : epel
Summary     : The RabbitMQ server
URL         : http://www.rabbitmq.com/
License     : MPLv1.1
Description : RabbitMQ is an implementation of AMQP, the emerging standard for high
            : performance enterprise messaging. The RabbitMQ server is a robust and
            : scalable implementation of an AMQP broker.

[root@centos-7-x64 ~]# which rabbitmq-plugins
/sbin/rabbitmq-plugins
```

PR #570, opened 2017-08-26
Switch back to using rabbitmq-plugins from system path (#566)

wyardley:
From #566, it seems as if the command in `/sbin` should be preferred, at least with modern RabbitMQ versions. Not sure if we need to get too smart about it, or how far back we need to support in terms of versions... presumably it would be possible to have it fall back to the second path, but not sure if that's overkill.

### DIFF
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -1,10 +1,14 @@
 require File.expand_path(File.join(File.dirname(__FILE__), '..', 'rabbitmqctl'))
 Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Provider::Rabbitmqctl) do
-  case Facter.value(:osfamily)
-  when 'RedHat'
-    has_command(:rabbitmqplugins, '/usr/lib/rabbitmq/bin/rabbitmq-plugins') { environment HOME: '/tmp' }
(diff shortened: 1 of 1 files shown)


wyardley:
Also some discussion in #134

wyardley:
FWIW, RabbitMQ 3.3.5 on CentOS 7 has this in `/usr/sbin`. But 3.1.5 on CentOS 7 does not, so that could be a good test case.

lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
```diff
@@ -1,10 +1,14 @@
 require File.expand_path(File.join(File.dirname(__FILE__), '..', 'rabbitmqctl'))
 Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Provider::Rabbitmqctl) do
-  case Facter.value(:osfamily)
-  when 'RedHat'
-    has_command(:rabbitmqplugins, '/usr/lib/rabbitmq/bin/rabbitmq-plugins') { environment HOME: '/tmp' }
+  # Prefer rabbitmq-plugins if it's in $PATH, but fall back to /usr/lib/rabbitmq/bin
+  if Puppet::Util.which('rabbitmq-plugins')
+    has_command(:rabbitmqplugins, 'rabbitmq-plugins') do
+      environment HOME: '/tmp'
```
wyardley: I did try manipulating PATH directly, but didn't have any luck when doing actual tests with it.

PR #571, opened 2017-08-28
removed package_provider var from readme and added deprecation check …

bostrowski13:
doesnt seem like package_provider is used anymore, so removing it from the readme and adding a deprecation check.


wyardley:
Looks good. @bostrowski13: are you able to squash this into one commit?

wyardley:
I squash-merged this.

PR #575, opened 2017-08-29
BREAKING: Drop Ruby 1.8 support. Rubocop auto-fixes in preparation for modulesync

wyardley:
This (will fail tests) is the first attempt at auto-correcting spec syntax and rubocop warnings using only `transpec` and `rubocop  -a` (no manual fixes; these will be handled separately). These commits are intentionally not squashed.


wyardley:
https://github.com/wyardley/puppet-rabbitmq/compare/auto_fixes...voxpupuli:modulesync can show some additional manual fixes I made, will clear out that branch before the next time we run a modulesync.

lib/puppet/provider/rabbitmq_user_permissions/rabbitmqctl.rb
```diff
@@ -1,28 +1,27 @@
...
-      self.run_with_retries {
+      run_with_retries do
         rabbitmqctl('-q', 'list_user_permissions', name)
-      }.split(/\n/).each do |line|
-        line = self::strip_backslashes(line)
-        if line =~ /^(\S+)\s+(\S*)\s+(\S*)\s+(\S*)$/
-          @users[name][$1] =
-            {:configure => $2, :read => $4, :write => $3}
+      end.split(%r{\n}).each do |line|
+        line = .strip_backslashes(line)
```
alexjfisher: Is this right?
Tests are failing here due to syntax, not rubocop.

wyardley: @alexjfisher yeah - I saw this last time too, for some reason, `rubocop -a` changed `self.strip_backslashes` to `.strip_backslashes` in that case. I added a separate commit (not squashed for this case, since it's a manual fix) that should resolve that particular issue. Let's see if the tests are at least somewhat happier after that.

There's one other issue that relates to spacing in the tests, I think I'll need to switch it to a heredoc to match it.

lib/puppet/provider/rabbitmq_user_permissions/rabbitmqctl.rb
```diff
@@ -18,7 +18,7 @@ def self.users(name, vhost)
       run_with_retries do
         rabbitmqctl('-q', 'list_user_permissions', name)
       end.split(%r{\n}).each do |line|
-        line = .strip_backslashes(line)
+        line = strip_backslashes(line)
```
wyardley: @alexjfisher I think this change (and this commit) should resolve the issues mentioned in your review.

Issue #573, opened 2017-08-28
repos_ensure and version on RHEL 7


bostrowski13:
This module doesn't seem to work on RHEL 7 after trying to spin up a new cluster this morning.  Its been a while since we've had to do this, but I've grabbed the latest from this repo, and after updating my code and removing params that are deprecated, I'm now unable to get the server pulled down from package cloud and receiving the error...

"Error: Could not update: Execution of '/bin/yum -d 0 -e 0 -y install rabbitmq-server-3.6.11' returned 1: One of the configured repositories failed (rabbitmq_rabbitmq-server),
 and yum doesn't have enough cached data to continue."
...
failure: repodata/repomd.xml from rabbitmq_rabbitmq-server: [Errno 256] No more mirrors to try.
https://packagecloud.io/rabbitmq/rabbitmq-server/el/7Server/x86_64/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found

Why is the rabbitmq repo not being used?  The one at https://www.rabbitmq.com/releases/

You could use version, with package provider RPM, and specify the source as the rabbitmq.com repo...

        version                  => $rabbitmq_version,
        package_provider         => 'rpm',
        package_source           => "https://www.rabbitmq.com/releases/rabbitmq-server/v${rabbitmq_version}/rabbitmq-server-${rabbitmq_version}-1.el${$osmajversion}.noarch.rpm",

You could also do the same thing with erlang, which is also available on the rabbitmq.com repo...

    package { 'erlang':
        ensure   => 'installed',
        source   => "https://www.rabbitmq.com/releases/erlang/erlang-${erlang_version}-1.el${osmajversion}.${distribution}.x86_64.rpm",
        provider => 'rpm'
    }

i would think this could be put in place of the repo::rhel class and you'd get all of your dependencies without including any other modules.

Looking at https://github.com/voxpupuli/puppet-rabbitmq/pull/493, theres some confusion about which param to use to pin the version, which again, doesn't seem to work when using repos_ensure => true on RHEL 7.

bostrowski13:
so at the time I had implemented the manifest below, i was able to use "version" to pin v3.6.3, even though 3.6.6 was just released.  Below is our implementation of the class.  the cluster is 2 nodes, and we're attempting to use the latest version, 3.6.11.

`class { '::rabbitmq':
    package_ensure           => $rabbitmq_version,
    repos_ensure             => true,
    service_manage           => true,
[...]

PR #577, opened 2017-08-30
Switch $releasevar to ${::os[release][major]}

wyardley:
Proposed fix for #573 -- this works around the issue where RHEL wants to hit 7Client / 7Server instead of 7 (which Packagecloud doesn't seem to have setup as valid names).

### DIFF
--- a/manifests/repo/rhel.pp
+++ b/manifests/repo/rhel.pp
@@ -1,7 +1,7 @@
 # Class: rabbitmq::repo::rhel
 # Makes sure that the Packagecloud repo is installed
 class rabbitmq::repo::rhel(
-    $location       = 'https://packagecloud.io/rabbitmq/rabbitmq-server/el/$releasever/$basearch',
+    $location       = "https://packagecloud.io/rabbitmq/rabbitmq-server/el/${::os[release][major]}/\$basearch",
     $key_source     = 'https://www.rabbitmq.com/rabbitmq-release-signing-key.asc',
   ) {
 
--- a/spec/spec_helper_local.rb
+++ b/spec/spec_helper_local.rb
@@ -12,14 +12,17 @@ end
 
 def with_debian_facts
   let :facts do
-    super().merge(operatingsystemmajrelease: '6',
-                  lsbdistcodename: 'squeeze',
-                  lsbdistid: 'Debian',
-                  osfamily: 'Debian',
-                  os: {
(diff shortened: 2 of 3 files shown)


manifests/repo/rhel.pp
```diff
@@ -1,7 +1,7 @@
 # Class: rabbitmq::repo::rhel
 # Makes sure that the Packagecloud repo is installed
 class rabbitmq::repo::rhel(
-    $location       = 'https://packagecloud.io/rabbitmq/rabbitmq-server/el/$releasever/$basearch',
+    $location       = "https://packagecloud.io/rabbitmq/rabbitmq-server/el/${::os[release][major]}/\$basearch",
```
bastelfreak: :+1: for structured facts

spec/spec_helper_local.rb
```diff
@@ -36,7 +36,11 @@ def with_openbsd_facts
 def with_redhat_facts
   let :facts do
     super().merge(operatingsystemmajrelease: '7',
```
bastelfreak: nitpicking, but could you add a newline after the ( in line 38 and before the { in line 40?

spec/spec_helper_local.rb
```diff
@@ -28,15 +30,24 @@ def with_openbsd_facts
...
+      osfamily: 'OpenBSD'
+    )
   end
 end
 
 def with_redhat_facts
   let :facts do
-    super().merge(operatingsystemmajrelease: '7',
-                  osfamily: 'Redhat')
+    super().merge(
```
wyardley: is this what you had in mind? I think the previous layout may have just been what `rubocop -a` did, or else copying the example of what was there before.

wyardley: Pushed one more update, let me know if it looks right. It didn't look right when I indented the curly brace after the newline.

wyardley: As you can probalby see, I also updated the Debian / OpenBSD facts.

But agree that we should rework this whole thing at some point.

manifests/repo/rhel.pp
alexjfisher: Bonus points if you also use the $facts hash?

wyardley: @alexjfisher should have thought of that... too bad it's already merged, but feel free to submit a PR 😉

PR #579, opened 2017-08-30
switch to structured facts for os* and rabbitmq_version

wyardley:
This is failing one test currently

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -132,8 +132,8 @@ class rabbitmq::config {
   # Get ranch (socket acceptor pool) availability,
   # Now that we have to rely on the fact, this may cause some chicken / egg
   # or idempotency problems
-  if $::rabbitmq_version {
-    $ranch = versioncmp($::rabbitmq_version, '3.6') >= 0
+  if $facts['rabbitmq_version'] {
+    $ranch = versioncmp($facts['rabbitmq_version'], '3.6') >= 0
   }
 
   file { '/etc/rabbitmq':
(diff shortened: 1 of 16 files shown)


wyardley:
May have to rebase this anyway after #581

wyardley:
@alexjfisher rebased, also updated stuff for @ekohl's recent changes and acceptance tests to use structured facts.

manifests/config.pp
```diff
@@ -192,9 +192,9 @@
     }
   }
 
-  case $::osfamily {
+  case $facts['os']['family'] {
     'Debian': {
-      if versioncmp($::operatingsystemmajrelease, '16.04') >= 0 {
+      if versioncmp($facts['os']['release']['major'], '16.04') >= 0 {
```
ekohl: I know this is a 1-to-1 conversion but the major is ```16``` so does that mean it'll only be true for ```17.04```? Maybe the ```['full']``` fact would be better.

Things might also need explicit Debian/Ubuntu checks if you add Debian 9 to the mix but that's for later.

wyardley: maybe we should just do major => 16?
I don't think the module supports 16 yet, though we should add support for it (someone was asking for it this morning)

wyardley: Though I guess it should eval true anyway because it's `versioncmp()`? I'll just switch to full assuming tests pass.

Would be great if someone could update the acceptance tests for 16.x

wyardley: Updated in new version

PR #580, opened 2017-08-30
update various small warnings

wyardley:
This just fixes some linter warnings, plus 2 minor rubocop warnings I forgot to enforce for another recent PR.

### DIFF
--- a/lib/facter/erl_ssl_path.rb
+++ b/lib/facter/erl_ssl_path.rb
@@ -11,6 +11,6 @@ Facter.add('erl_ssl_path') do
       end
     end
     # erl returns the string with quotes, strip them off
-    data.gsub!(/\A"|"\Z/, '')
+    data.gsub!(%r{\A"|"\Z}, '')
   end
 end
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -76,7 +76,7 @@ class rabbitmq::config {
   } else {
     $default_ssl_env_variables = {
       'NODE_PORT'        => $port,
-      'NODE_IP_ADDRESS'  => $node_ip_address
+      'NODE_IP_ADDRESS'  => $node_ip_address,
     }
   }
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -5,8 +5,10 @@
 # @example Declaring the class
 #   class { 'rabbitmq': }
 #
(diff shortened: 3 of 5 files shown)


wyardley:
Added some changes to #581 to the original request, including some more `rubocop -a` fixes, and removing some checks that were duplicated inside conditionals in a single `if` block.

Not 100% sure if this reduce coverage too much, happy for suggestions on how to satisfy the duplicate expectation within a block error differently.

cc @ekohl

spec/classes/rabbitmq_spec.rb
```diff
@@ -41,37 +41,35 @@
...
-              .with_release('testing')
-              .with_repos('main')
+            is_expected.to contain_apt__source('rabbitmq').
+              with_ensure('present').
+              with_location('http://www.rabbitmq.com/debian/').
+              with_release('testing').
+              with_repos('main')
           end
 
-          it { is_expected.not_to contain_class('rabbitmq::repo::rhel') }
```
ekohl: This feels like a false positive to me.

spec/classes/rabbitmq_spec.rb
```diff
@@ -41,37 +41,35 @@
...
         when 'RedHat'
-          it { is_expected.not_to contain_class('rabbitmq::repo::apt') }
```
ekohl: This too

spec/classes/rabbitmq_spec.rb
wyardley: Yeah, I'm wondering if it should just be reordered or have slightly different logic?
Maybe it could just be inside a `when not 'xyz'` block?

wyardley: See how the updated version I'm about to push looks (oddly, it needs to be `when !('xyz')`)

ekohl: Not pretty but if it works ...

spec/classes/rabbitmq_spec.rb
```diff
@@ -113,10 +114,8 @@
         context "with file_limit => '#{value}'" do
           let(:params) { { file_limit: value } }
 
-
           case facts[:osfamily]
           when 'RedHat'
-            it { is_expected.not_to contain_file('/etc/default/rabbitmq-server') }
```
ekohl: This also needs the ```!'RedHat'``` and ```!'Debian'``` treatment I guess

PR #581, opened 2017-08-31
Switch to rspec-puppet-facts

ekohl:


spec/spec_helper_local.rb
```diff
@@ -1,82 +1,6 @@
...
-  end
-end
-
-def with_distro_facts(distro)
-  send("with_#{distro.downcase}_facts")
-end
+require 'rspec-puppet-facts'
+include RspecPuppetFacts
+                                                    # Original fact sources:
+add_custom_fact :puppetversion, Puppet.version      # Facter, but excluded from rspec-puppet-facts
```
wyardley: Is there a more up to date way to include these (maybe in `spec/default_module_facts.yaml`)?

spec/classes/rabbitmq_spec.rb
```diff
@@ -9,312 +9,187 @@
...
-  end
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts }
 
-  context 'on Debian' do
-    let(:params) { { repos_ensure: true } }
+      has_systemd = (
+        (facts[:osfamily] == 'RedHat' && facts[:os]["release"]["major"].to_i >= 7) ||
+        (facts[:osfamily] == 'Debian' && facts[:os]["release"]["full"] == '16.04')
```
bastelfreak: can you add archlinux here as well?

spec/spec_helper_local.rb
bastelfreak: after a modulesync, puppetversion will be set. staging:http_get can be set via spec/default_module_facts.yml

spec/classes/rabbitmq_spec.rb
bastelfreak: I will add arch suport in a separate PR

wyardley:
Did introduce some rubocop warnings, which I'll fix  while rebasing #580

PR #583, opened 2017-08-31
Add official archlinux support

bastelfreak:


spec/classes/rabbitmq_spec.rb
```diff
@@ -17,7 +17,8 @@
 
       has_systemd = (
         (facts[:osfamily] == 'RedHat' && facts[:os]["release"]["major"].to_i >= 7) ||
-        (facts[:osfamily] == 'Debian' && facts[:os]["release"]["full"] == '16.04')
+        (facts[:osfamily] == 'Debian' && facts[:os]["release"]["full"] == '16.04') ||
+        (facts[:osfamily] == 'Archlinux')
```
ekohl: This is not in the actual code and the tests only pass because rspec-puppet-facts always needs a major version. That's also why OpenBSD is currently not tested.

PR #589, opened 2017-09-01
Update dependencies, name, and required Puppet version

wyardley:
Update dependencies, name, and required Puppet version

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,12 +1,12 @@
 {
-  "name": "puppetlabs-rabbitmq",
-  "version": "5.6.0",
-  "author": "puppetlabs",
+  "name": "puppet-rabbitmq",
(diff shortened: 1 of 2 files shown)


metadata.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "puppetlabs-rabbitmq",
+  "name": "puppet-rabbitmq",
   "version": "5.6.0",
```
alexjfisher: Maybe update version to an 'rc' version? 6.0.0-rc??

wyardley: works for me, 6.0.0-rc0? I figured that might get set when we do modulesync, but I don't see any reason not to update it, and next release will have a lot of breaking changes, I think.

wyardley:
@alexjfisher updated and tests completed

PR #592, opened 2017-09-03
Fix stdlib requirement

alexjfisher:


bastelfreak:
oh a typo, my bad I introduced this :(

PR #594, opened 2017-09-03
Remove redundant specs, rework others to avoid multiple expectation warnings

wyardley:
This should help reduce the # of multiple expectation warnings.


spec/unit/puppet/provider/rabbitmq_user/rabbitmqctl_spec.rb
```diff
@@ -71,6 +71,8 @@
 foo [administrator]
 EOT
     expect(provider.admin).to eq(:true)
+  end
```
wyardley: I think this is actually better this way?

PR #596, opened 2017-09-03
modulesync 1.0.0

wyardley:
This will probably fail, I can make the changes to fix stuff as a separate commit.

I think will also require #595 to be merged


wyardley:
I kind of cheated to add back in the changes that .sync.yml would add,  but I think this should work.
Rubocop is still failing but other checks are looking Ok.

PR #460, opened 2016-04-18
Update rabbitmq_user provider

mpolenchuk:
Rabbitmq_user provider has been rewritten
by means of prefetch/flush methods.

Issue #147, opened 2013-11-27
Error on using puppet resource rabbitmq_user


matthewbarr:
Per the docs, you should be able to see the current users with puppet resource rabbitmq_user.

That currently gives an error on both puppet 3.1.1 & 3.3.2:

Error: Could not run: Validation of Rabbitmq_user[guest] failed: must set password when creating user
Wrapped exception:
must set password when creating user
```
Manifest on master:
  class {'::rabbitmq':
}

  rabbitmq_user { 'usernamehere':
    admin    => true,
    password => 'XXXX',
  }
```

wyardley:
I'm able to verify this, even with current master.

wyardley:
https://tickets.puppetlabs.com/browse/MODULES-3084 seems to be the same issue as well.

wyardley:
I think #598 (rework of #460 by @mpolenchuk) might end up fixing this if I can get it to work.

wyardley:
The actual issue seems to be unrelated to those two. It's actually much simpler. The code that's trying to ensure that you don't create a user without defining a password is also preventing `puppet resource rabbitmq_user` from working.

```ruby
  validate do
    if self[:ensure] == :present && !self[:password]
      raise ArgumentError, 'must set password when creating user' unless self[:password]
    end
  end
```

wyardley:
Closing via #598 (I'll cherry-pick the fix for this or break it out into another pull if that doesn't get merged for some reason).

PR #598, opened 2017-09-04
Refactor rabbitmq_user provider (mpolenchuk)

wyardley:
This is an attempt to rework #460 by @mpolenchuk (keeping original attribution).
Could use careful review, rebasing was pretty hairy, and not sure I got everything.

As of now, failing at least one test.

### DIFF
--- a/lib/puppet/type/rabbitmq_user.rb
+++ b/lib/puppet/type/rabbitmq_user.rb
@@ -39,20 +39,12 @@ DESC
 
   newproperty(:password) do
     desc 'User password to be set *on creation* and validated each run'
-    def insync?(_is)
-      provider.check_password
+    def insync?(is)
+      provider.check_password(should)
     end
 
-    def set(_value)
-      provider.change_password
-    end
-
-    def change_to_s(_current, _desired)
-      'password has been changed'
-    end
-
-    def should_to_s(_newvalue = @should)
-      '<new password>'
+    def change_to_s(current, desired)
+      "password has been changed"
     end
   end
 
@@ -80,23 +72,11 @@ DESC
     defaultto []
 
     def insync?(is)
-      is_to_s(is) == should_to_s
+      is.sort == should.sort
     end
 
-    def is_to_s(currentvalue = @is) # rubocop:disable Style/PredicateName
-      if currentvalue
-        "[#{currentvalue.sort.join(', ')}]"
-      else
-        '[]'
-      end
-    end
-
-    def should_to_s(newvalue = @should)
-      if newvalue
-        "[#{newvalue.sort.join(', ')}]"
-      else
-        '[]'
-      end
+    def should_to_s(value)
+      Array(value)
     end
   end
 end
--- a/lib/puppet/type/rabbitmq_user.rb
+++ b/lib/puppet/type/rabbitmq_user.rb
@@ -39,12 +39,12 @@ DESC
 
   newproperty(:password) do
     desc 'User password to be set *on creation* and validated each run'
-    def insync?(is)
+    def insync?(_is)
       provider.check_password(should)
     end
 
-    def change_to_s(current, desired)
-      "password has been changed"
+    def change_to_s(_current, _desired)
+      'password has been changed'
     end
   end
 
--- a/spec/unit/puppet/provider/rabbitmq_user/rabbitmqctl_spec.rb
+++ b/spec/unit/puppet/provider/rabbitmq_user/rabbitmqctl_spec.rb
@@ -1,222 +1,123 @@
 require 'spec_helper'
 
-provider_class = Puppet::Type.type(:rabbitmq_user).provider(:rabbitmqctl)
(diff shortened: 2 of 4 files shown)


wyardley:
@ekohl @hunner I think this is ready for review, and would very much appreciate a review on this one if at all possible. In particular, I think the tests could use some improvement, especially for the change password situation.

However, at a functional level, I think this is working(ish), and an improvement.

spec/unit/puppet/provider/rabbitmq_user/rabbitmqctl_spec.rb
```diff
@@ -3,213 +3,174 @@
...
+{ok,{user,<<"rmq_x">>,[],rabbit_auth_backend_internal,
+          {internal_user,<<"rmq_x">>,
+                         <<193,81,62,182,129,135,196,89,148,87,227,48,86,2,154,
+                           192,52,119,214,177>>,
+                         []}}}
+EOT
+      end
+
+      it do
+        provider.check_password('secret')
```
wyardley: I don't think this is doing anything useful, but need to come up with a better test here.

wyardley:
@mpolenchuk - don't know if you've got time to review, but let me know if this looks sane. I'm still trying to come up with a better test case for the password change bit.

lib/puppet/type/rabbitmq_user.rb
```diff
@@ -40,20 +40,12 @@
...
     def change_to_s(_current, _desired)
       'password has been changed'
     end
-
-    def should_to_s(_newvalue = @should)
-      '<new password>'
-    end
   end
 
   newproperty(:admin) do
```
ekohl: You can use ```newproperty(:admin, boolean: true, parent: Parent::Property::Boolean)``` and remove the body (except ```desc```). Then you can use true booleans in the provider. You might need a ```require 'puppet/property/boolean``` but I'm not sure about that.

wyardley: I think this is a good idea, but I feel like it should be implemented separately, and then we can review all the type parameters at once, and adjust any that we can switch to true booleans / integers / etc in one 'breaking' changeset.

wyardley:
Removing bug label since #147 was fixed via #629

lib/puppet/type/rabbitmq_user.rb
ekohl: That's fair

spec/unit/puppet/provider/rabbitmq_user/rabbitmqctl_spec.rb
```diff
@@ -3,220 +3,174 @@
...
-one [administrator]
-foo []
-icinga  [monitoring]
-kitchen []
-kitchen2        [abc, def, ghi]
-EOT
-    provider.expects(:rabbitmqctl).with('set_user_tags', 'foo', %w[tag1 tag2])
-    provider.tags = %w[tag1 tag2]
-  end
+  describe '#check_password' do
```
wyardley: Is this doing anything useful? Or should this (syncing password) be checked in the type spec instead, or by calling another method vs. the check_password one?

wyardley:
> You could consider expanding the user acceptance test with purging

@ekohl added a test of ensuring absent on the user resource, which I think accomplishes the same thing, and also switched to the shared example now that #632 is merged.

wyardley:
@ekohl: the shared example seems to get executed after the other elements in the block, which is causing an ordering issue.
For example
```
    it_behaves_like 'an idempotent resource'

    it 'does not have the user' do
      shell('rabbitmqctl list_users -q') do |r|
        expect(r.stdout).not_to match(%r{dan\s+})
      end
    end
```
results in
```
  destroy user resource
    does not have the user (FAILED - 1)
    behaves like an idempotent resource
localhost $ scp /var/folders/0z/yrnglm6n1l5_q19dv1yy0w3c0000gn/T/beaker20170913-21543-xcrm0o centos-7-x64:/tmp/apply_manifest.pp.Kz0FfQ {:ignore => }
      applies with no errors
localhost $ scp /var/folders/0z/yrnglm6n1l5_q19dv1yy0w3c0000gn/T/beaker20170913-21543-ez5qfe centos-7-x64:/tmp/apply_manifest.pp.BUFI7K {:ignore => }
      applies a second time without changes
```

It's failing because the test for the user being gone runs _before_ the shared example runs.

lib/puppet/type/rabbitmq_user.rb
wyardley: It seems like the existing code is (in some cases) as simple as:
```
  newparam(:force) do
    defaultto(:false)
    newvalues(:true, :false)
  end
```
which actually seems simpler than the suggested code in https://docs.puppet.com/puppet/5.1/custom_types.html, and without that extra `require`. Also, the puppet boolean type allows `'true'` `yes`, and various other things, which seems kind of regressive.

PR #601, opened 2017-09-04
Eliminate more "multiple expectations" warnings and remove some redundant specs

wyardley:
Would appreciate feedback on this one.

### DIFF
--- a/spec/unit/puppet/type/rabbitmq_policy_spec.rb
+++ b/spec/unit/puppet/type/rabbitmq_policy_spec.rb
@@ -62,11 +62,13 @@ describe Puppet::Type.type(:rabbitmq_policy) do
     expect(policy[:definition]).to eq(definition)
   end
 
-  it 'does not accept invalid hash for definition' do
+  it 'does not accept a string for definition' do
     expect do
       policy[:definition] = 'ha-mode'
     end.to raise_error(Puppet::Error, %r{Invalid definition})
+  end
 
+  it 'does not accept invalid hash for definition' do
     expect do
       policy[:definition] = { 'ha-mode' => %w[a b] }
     end.to raise_error(Puppet::Error, %r{Invalid definition})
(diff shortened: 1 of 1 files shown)


spec/unit/puppet/type/rabbitmq_policy_spec.rb
```diff
@@ -171,13 +167,25 @@
...
   end
 
-  it 'accepts list value in ha-params when ha-mode = nodes' do
-    definition = { 'ha-mode' => 'nodes', 'ha-params' => ['rabbit@rabbit-01', 'rabbit@rabbit-02'] }
-    policy[:definition] = definition
-    expect(policy[:definition]['ha-mode']).to eq('nodes')
-    expect(policy[:definition]['ha-params']).to be_a(Array)
-    expect(policy[:definition]['ha-params'][0]).to eq('rabbit@rabbit-01')
-    expect(policy[:definition]['ha-params'][1]).to eq('rabbit@rabbit-02')
+  context 'accepts list value in ha-params when ha-mode = nodes' do
```
wyardley: I think this works Ok here, but not sure if there's a better way to do it.

spec/unit/puppet/type/rabbitmq_policy_spec.rb
```diff
@@ -171,13 +167,25 @@
...
+    before do
+      policy[:definition] = definition
+    end
+
+    let(:definition) { { 'ha-mode' => 'nodes', 'ha-params' => ['rabbit@rabbit-01', 'rabbit@rabbit-02'] } }
+
+    it do
+      expect(policy[:definition]['ha-mode']).to eq('nodes')
+    end
```
alexjfisher: I might be tempted to use a single line for each of these?
ie.
```
it { expect(policy[:definition]['ha-mode']).to eq('nodes') }
```

wyardley: Will do. You think moving the outer bit to a context makes sense here though?

wyardley: Updated, btw, if you want to re-review

PR #604, opened 2017-09-05
Convert to use 'archive' instead of 'staging' for rabbitmqadmin install

wyardley:
This PR switches the module to use the 'archive' module instead of 'staging'.

### DIFF
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -1,6 +1,8 @@
 #
 class rabbitmq::install::rabbitmqadmin {
 
+  require '::archive'
+
   if($rabbitmq::ssl and $rabbitmq::management_ssl) {
     $management_port = $rabbitmq::ssl_management_port
     $protocol        = 'https'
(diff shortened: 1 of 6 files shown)


manifests/install/rabbitmqadmin.pp
```diff
@@ -25,13 +27,14 @@
...
 
-  staging::file { 'rabbitmqadmin':
-    target      => "${rabbitmq::rabbitmq_home}/rabbitmqadmin",
-    source      => "${protocol}://${sanitized_ip}:${management_port}/cli/rabbitmqadmin",
-    curl_option => "-u \"${default_user}:${default_pass}\" -k ${curl_prefix} --retry 30 --retry-delay 6",
-    timeout     => '180',
-    wget_option => '--no-proxy --no-check-certificate',
-    require     => [
+  archive { 'rabbitmqadmin':
+    path           => "${rabbitmq::rabbitmq_home}/rabbitmqadmin",
```
wyardley: Not sure if it's better to update this to just stage it, drop it in its final destination upon successful retrieval, and then clean up the tmp file vs. having the file resource?

bastelfreak: I think the current behavior is totally valid

PR #606, opened 2017-09-05
fix a couple of problems with erl_ssl_path fact

costela:
Hi there,

This small PR fixes two small issues with the `erl_ssl_path` fact.
The first is pretty straightforward: avoid the `.gsub` call if we didn't actually get any data to mangle.
The second is a bit less clear-cut: drop the use of `.with_env` while calling `erl`. Newer versions of facter (>=3) AFAIK stopped supporting it (compare [2.x docs](https://puppetlabs.github.io/facter/2.0.x/Facter/Util/Resolution.html) with [3.x docs](https://puppetlabs.github.io/facter/master/structfacter_1_1ruby_1_1resolution.html)) and it was unclear why we needed it in the first place. If I'm missing something and it *is* needed in some case, I'd be happy to find another workaround.

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,7 +708,7 @@ This module is maintained by [Vox Pupuli](https://voxpupuli.org/). Voxpupuli
 welcomes new contributions to this module, especially those that include
 documentation and rspec tests. We are happy to provide guidance if necessary.
 
-Please see [CONTRIBUTING](.git/CONTRIBUTING.md) for more details.
+Please see [CONTRIBUTING](.github/CONTRIBUTING.md) for more details.
 
 ### Authors
 * Jeff McCune <jeff@puppetlabs.com>


wyardley:
@costela couple things:
1) I should have asked for tests on the first go-round with this new fact. The test should be in `spec/unit/facter/util/fact_erl_ssl_path.rb`. Let me know if you need more help / guidance on this part.
2) Voxpupuli generally likes the commits to be squashed (so rebase, fixup the other commits, and then force-push the branch).

wyardley:
@costela ps - Feel free to chat in #voxpupuli on irc.freenode.net or Puppet Community Slack if you need help w/ any of this. I realize you're not the original person who submitted the change. Let me know if creating the tests will be a problem.

costela:
FYI: I've moved this PR to a different branch, and github doesn't seem to support updating the source branch (just the destination branch). So I'll open a new one shortly, including tests and squashed commits.

PR #613, opened 2017-09-06
Switch to 'contain' vs. anchor pattern, and use Class['foo'] vs Class['::foo']

wyardley:
Update class ordering to follow current best practices.

This also removes some apparently redundant uses of `require => Class['rabbitmq::install']`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -151,45 +151,41 @@ class rabbitmq(
     }
   }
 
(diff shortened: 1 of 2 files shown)


wyardley:
Unless we get rid of `puppet-lint-absolute_classname-check` in `Gemfile`, we'll have to use
contain '::rabbitmq::foo'
vs
contain rabbitmq::foo

PR #617, opened 2017-09-07
Add official support for Ubuntu 16.04

wyardley:
This adds official support for Ubuntu 16.04; best I can tell, it already works.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,7 +2,8 @@ fixtures:
   repositories:
     "stdlib": "git://github.com/puppetlabs/puppetlabs-stdlib.git"
     "apt": "git://github.com/puppetlabs/puppetlabs-apt.git"
-    "archive": "https://github.com/puppet-community/puppet-archive.git"
+    "archive": "https://github.com/voxpupuli/puppet-archive.git"
+    "systemd": "https://github.com/camptocamp/puppet-systemd.git"
     "erlang": "git://github.com/garethr/garethr-erlang.git"
   symlinks:
     "rabbitmq": "#{source_dir}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
   - rvm: 2.1.9
     bundler_args: --without development
     dist: trusty
-    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/ubuntu-14.04 CHECK=beaker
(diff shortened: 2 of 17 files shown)


wyardley:
TODO: switch exec for systemd to https://github.com/camptocamp/puppet-systemd

wyardley:
TODO: Travis tests don't seem to have Python:
```
       	Error: Execution of '/usr/local/bin/rabbitmqadmin declare exchange --vhost=host2 --user=dan --password=bar name=exchange2 type=topic internal=false durable=false auto_delete=false arguments={} -c /etc/rabbitmq/rabbitmqadmin.conf' returned 127: /usr/bin/env: 'python': No such file or directory
```
acceptance tests pass locally in Vagrant, though. Will test locally with Docker.

wyardley:
Adding this makes the Docker acceptance tests pass, though not sure if there's a better / cleaner way to do it (maybe use puppet itself, like I've seen some modules do?)

```
      if fact('os.family') == 'Debian'
        on host, puppet('module', 'install', 'puppetlabs-apt'), acceptable_exit_codes: [0, 1]
        # work around lack of Python package on Ubuntu 16 for now
        on host, 'apt-get install -y python'
```
I think this shouldn't hurt even when the system already has python.

ekohl:
I'd argue that https://github.com/voxpupuli/puppet-rabbitmq/blob/master/manifests/install/rabbitmqadmin.pp should ensure python is installed since that installs it. There might be other Ubuntu 16.04 users that have a very minimal install and could run in the same issue.

wyardley:
I think I'm going to end up breaking this into an issue w/ separate PRs.
@bastelfreak: The camptocamp systemd stuff is going to be more complicated than I originally thought to do properly (see this PR for why), but I think there may be another way to work around it in the meantime.

wyardley:
@ekohl The thing w/ adding a package ensure is that it's never as simple as you hope, so was really hoping to have to do something like
`ensure_packages('python')`
Then we have to keep track of names on any system types where it's different, deal with problems when someone has defined the package in a different way that causes a resources problem with Puppet, etc. etc. (meaning we might end up needing to add a new option controlling whether or not to install it).

It does seem like Python 2.x is _not_ installed on Ubuntu 16 by default (https://www.digitalocean.com/community/tutorials/what-s-new-in-ubuntu-16-04), so maybe the Puppet image bakes it in?

ekohl:
I know of all the issues but it's the right thing to do. The best solution would be a package in a repo with the proper dependencies but that's not something we can do.

And the Puppet image could have a dependency somewhere so they pull it in.

PR #620, opened 2017-09-08
Update 'require' statements and mock types to fix spec tests

wyardley:
This resolves the issues I was having with running the rspec tests locally.


wyardley:
https://stackoverflow.com/questions/7844336/rspec-mustbeconfiguredbeforeexamplegroupserror-on-plain-project describes the reasoning behind this.
removing `require 'mocha'` and `require 'mocha/api` seems to result in the tests still working, so I'm thinking not needed, but welcome feedback.

Issue #618, opened 2017-09-07
Idempotency problems with rabbitmq 3.6.5 and puppet 4.8.2


mwhahaha:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.8.2
- Ruby: ruby-2.0.0.648-29.el7.x86_64
- Distribution: centos 
- Module version: 70ac6c59180a307786db28218d11348acc6d7a42
- Rabbitmq 3.6.5

## How to reproduce (e.g Puppet code you use)
https://github.com/openstack/puppet-openstack-integration/blob/master/manifests/rabbitmq.pp

## What are you seeing
Idempotency problem with the rabbit config
```
2017-09-07 17:11:16 +0000 /Stage[main]/Rabbitmq::Config/File[rabbitmq.config]/content (notice): 
--- /etc/rabbitmq/rabbitmq.config	2017-09-07 16:47:01.384000000 +0000
+++ /tmp/puppet-file20170907-7672-1kuhj4r	2017-09-07 17:11:16.605000000 +0000
@@ -3,9 +3,6 @@
 [
   {rabbit, [
     {tcp_listen_options, [
-         binary,
-         {packet,        raw},
-         {reuseaddr,     true},
          {backlog,       128},
          {nodelay,       true},
          {linger,        {true, 0}},
2017-09-07 17:11:16 +0000 Puppet (info): Computing checksum on file /etc/rabbitmq/rabbitmq.config
2017-09-07 17:11:16 +0000 /Stage[main]/Rabbitmq::Config/File[rabbitmq.config] (info): Filebucketed /etc/rabbitmq/rabbitmq.config to puppet with sum 21f23f84e979e437300cf0f14cecc889
2017-09-07 17:11:16 +0000 /Stage[main]/Rabbitmq::Config/File[rabbitmq.config]/content (notice): content changed '{md5}21f23f84e979e437300cf0f14cecc889' to '{md5}0ef507de9e71257dcc93b8974445296b'
2017-09-07 17:11:16 +0000 /Stage[main]/Rabbitmq::Config/File[rabbitmq.config] (info): Scheduling refresh of Class[Rabbitmq::Service]
2017-09-07 17:11:16 +0000 Class[Rabbitmq::Config] (info): Scheduling refresh of Class[Rabbitmq::Service]
2017-09-07 17:11:17 +0000 Class[Rabbitmq::Service] (info): Scheduling refresh of Service[rabbitmq-server]
[...]

PR #621, opened 2017-09-08
BREAKING: Adds config_ranch parameter (default: true) to suppress config lines (#618)

wyardley:
I haven't added tests yet, but wanted to throw this up to see how this proposed fix looks.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -73,6 +73,7 @@ class rabbitmq::params {
   $config                      = 'rabbitmq/rabbitmq.config.erb'
   $config_cluster              = false
   $config_path                 = '/etc/rabbitmq/rabbitmq.config'
+  $config_ranch                = true
   $config_stomp                = false
   $config_shovel               = false
   $config_shovel_statics       = {}
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -288,6 +288,26 @@ describe 'rabbitmq' do
         }
       end
 
+      describe 'does not contain pre-ranch settings with default config' do
+        it do
+          is_expected.to contain_file('rabbitmq.config'). \
+            without_content(%r{binary,}).                 \
+            without_content(%r{\{packet,        raw\},}). \
+            without_content(%r{\{reuseaddr,     true\},})
+        end
+      end
+
+      describe 'contains pre-ranch settings with config_ranch set to false' do
+        let(:params) { { config_ranch: false } }
+
+        it do
+          is_expected.to contain_file('rabbitmq.config'). \
+            with_content(%r{binary,}).                 \
+            with_content(%r{\{packet,        raw\},}). \
+            with_content(%r{\{reuseaddr,     true\},})
+        end
+      end
+
       context 'configures config_cluster' do
         let(:params) do
           {
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -9,6 +9,7 @@ class rabbitmq::config {
   $config                      = $rabbitmq::config
   $config_cluster              = $rabbitmq::config_cluster
   $config_path                 = $rabbitmq::config_path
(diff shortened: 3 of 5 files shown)


wyardley:
Added tests, which also exposed a critical bug in my earlier changes.

Will solicit some input about the naming of the variable; I'm not super happy about `config_ranch`, but I think it's less misleading than some other options. Are we considering this a breaking change @alexjfisher @ekohl?

README.md
```diff
@@ -50,6 +50,105 @@ These are now documented via [Puppet Strings](https://github.com/puppetlabs/pupp
 
 For convenience, some examples are duplicated here:
 
+## rabbitmq class
```
wyardley: This should not show up once #625 is merged and I rebase

bastelfreak: @wyardley can you rebase? I merged #625

manifests/params.pp
```diff
@@ -73,6 +73,7 @@
   $config                      = 'rabbitmq/rabbitmq.config.erb'
   $config_cluster              = false
   $config_path                 = '/etc/rabbitmq/rabbitmq.config'
+  $config_ranch                = true
```
ekohl: How about ```$config_ranch = !$facts['rabbitmq_version'] || versioncmp($facts['rabbitmq_version'], '3.6') >= 0```

ekohl:
> Will solicit some input about the naming of the variable; I'm not super happy about config_ranch, but I think it's less misleading than some other options.

You could consider ```enable_ranch```.

>Are we considering this a breaking change @alexjfisher @ekohl?

Right now it is and with my suggestion (enable ranch if not present, otherwise detect it) it's also breaking, but less so because a second run should reach the old behaviour. If you want to be compatible you should remove the ```!``` from my suggestion.

wyardley:
> You could consider enable_ranch.

So my original iteration of that had something like that, but I think it's a bad idea for this reason: the parameter doesn't _enable_ ranch, rather, on versions of RabbitMQ that use ranch, it suppresses certain config directives that were necessary before that (I think pre-3.6). The directives are unneeded, though not sure they actually break anything when they're present. But, as best I understand, having them there or not does not enable or disable ranch.

manifests/params.pp
wyardley: will the `versioncmp()` call translate directly into a bool automagically?

wyardley: I like this idea in general, just wondering if we're trying to be too clever and might end up shooting ourself in the foot?
Also, maybe it would be easier to read if we just had an `if {}` block rather than keeping it so terse?

ekohl: > will the versioncmp() call translate directly into a bool automagically?

Maybe you need braces around the comparison, haven't tested it.

spec/classes/rabbitmq_spec.rb
```diff
@@ -288,6 +288,26 @@
         }
       end
 
+      describe 'does not contain pre-ranch settings with default config' do
```
ekohl: If the default is true then you would expect it to be present, correct?

wyardley: Well, I agree that the wording could possibly use some improvement, but no, the setting suppresses certain pre-ranch config settings with newer versions of RabbitMQ.

I guess we could invert the boolean and use `config_pre_ranch`, but that also seems kind of awkward.

ekohl: Oh you're right, I was looking at the tests and misread that they both test the same thing. Now I see one is ```without_content``` and the other ```with_content```

wyardley:
I'm going to merge this as it is now, I think the best bet for now is to set the default and not try to be too clever, though we can revisit later if necessary. I feel bad that we'll have to do another major release so quickly, in the future, it would be nice if there's a way to release RCs in a semi-official way so that we can get some more feedback ahead of a major release.

PR #623, opened 2017-09-08
BREAKING: Ensure python package (adds manage_python option)

wyardley:
This is needed for Ubuntu 16.04+ where only Python 3 is installed by default.

This could cause issues in some cases where people have defined the Python package elsewhere, I'm open to suggestions on how to do this more sanely, though really hoping to avoid an additional parameter.

If RabbitMQ distributed it in the package, the dependency could be handled more cleanly.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -124,6 +124,8 @@
 # @param ldap_use_ssl Set to true to use SSL for the LDAP server.
 # @param ldap_port Numeric port for LDAP server.
 # @param ldap_log Set to true to log LDAP auth.
+# @param manage_python If enabled, on platforms that don't provide a Python 2 package by default, ensure that the python package is
+#  installed (for rabbitmqadmin). This will only apply if `admin_enable` and `service_manage` are set.
 # @param manage_repos Whether or not to manage package repositories.
 # @param management_hostname The hostname for the RabbitMQ management interface.
 # @param management_port The port for the RabbitMQ management interface.
@@ -213,6 +215,7 @@ class rabbitmq(
   Optional[String] $package_provider             = undef,
   Boolean $repos_ensure                          = $rabbitmq::params::repos_ensure,
   $manage_repos                                  = undef,
+  Boolean $manage_python                         = $rabbitmq::params::manage_python,
   $rabbitmq_user                                 = $rabbitmq::params::rabbitmq_user,
   $rabbitmq_group                                = $rabbitmq::params::rabbitmq_group,
   $rabbitmq_home                                 = $rabbitmq::params::rabbitmq_home,
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -3,6 +3,15 @@ class rabbitmq::install::rabbitmqadmin {
 
   require '::archive'
 
+  $python_package = $rabbitmq::params::python_package
+  # Some systems (e.g., Ubuntu 16.04) don't ship Python 2 by default
+  if $rabbitmq::manage_python {
+    ensure_packages([$python_package])
+    $rabbitmqadmin_require = [Archive['rabbitmqadmin'], Package[$python_package]]
+  } else {
+    $rabbitmqadmin_require = Archive['rabbitmqadmin']
+  }
+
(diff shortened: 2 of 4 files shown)


manifests/install/rabbitmqadmin.pp
```diff
@@ -3,6 +3,14 @@
 
   require '::archive'
 
+  # Newer versions of Ubuntu (16.04) ship with Python 3 only by default.
+  if $facts['os']['family'] == 'Debian' {
+    ensure_packages('python')
+    $rabbitmqadmin_require = [Archive['rabbitmqadmin'], Package['python']]
```
wyardley: Not sure if there's a cleaner way to do this

wyardley:
Let me know if this needs to be a parameter, though I'd rather not add another knob / whistle for this.

manifests/install/rabbitmqadmin.pp
wyardley: also, this logic needs to be moved to params, and I think it probably does need to be configurable.

spec/classes/rabbitmq_spec.rb
```diff
@@ -177,9 +177,9 @@
 
         it {
           is_expected.to contain_exec('rabbitmq-systemd-reload').with(
-            'command'     => '/usr/bin/systemctl daemon-reload',
-            'notify'      => 'Class[Rabbitmq::Service]',
-            'refreshonly' => true
+            command: '/usr/bin/systemctl daemon-reload',
```
wyardley: not sure why rubocop wasn't complaining about these before

manifests/init.pp
```diff
@@ -213,6 +215,7 @@
   Optional[String] $package_provider             = undef,
   Boolean $repos_ensure                          = $rabbitmq::params::repos_ensure,
   $manage_repos                                  = undef,
+  Boolean $manage_python                         = $rabbitmq::params::manage_python,
```
bastelfreak: datatypes \o/

manifests/params.pp
```diff
@@ -6,6 +6,7 @@
 
   case $facts['os']['family'] {
     'Archlinux': {
+      $manage_python    = false
```
bastelfreak: do we need python only on debian like systems?

ekohl: I think we can require python on all platforms. It's ```ensure_packages``` so there's less chance for conflicts and otherwise we provide the option to disable it.

manifests/install/rabbitmqadmin.pp
```diff
@@ -3,6 +3,14 @@
 
   require '::archive'
 
+  # Newer versions of Ubuntu (16.04) ship with Python 3 only by default.
+  if $rabbitmq::manage_python {
+    ensure_packages('python')
```
ekohl: https://github.com/puppetlabs/puppetlabs-stdlib#ensure_packages states it takes a list of packages so I'm surprised this works.

manifests/params.pp
wyardley: We can if you think that makes sense. My thought was to take a conservative approach, since a) I don't know for sure which systems don't already have it, and b) I don't know if the python 2 package name is consistent across all those platforms, and we don't do acceptance tests on all of them. Once this is enabled, we could switch the param for other platforms. Thoughts?

manifests/install/rabbitmqadmin.pp
wyardley: Yeah, I think it does work, but I forced it to list context, which I think is cleaner.

wyardley: I can switch it to managing it for all platforms, seems kind of silly for RHEL / CentOS, where I think it's a hard dependency for tools like yum / anaconda.

Do we know what the package is called on SUSE?

wyardley:
Ok, how do we feel about this behavior now, @ekohl @bastelfreak? I don't have a way to test on all of these platforms, though verified that the package is python2 on FreeBSD (so presumably on OBSD as well).

I added the package name in params but didn't make it configurable.

bastelfreak:
looks good to me. @ekohl and objections here?

PR #624, opened 2017-09-08
Add official support for Ubuntu 16.04

wyardley:
I think we can avoid making bigger changes to tests or systemd for now by just hitting `/bin/systemctl` on both Ubuntu and CentOS systems.

This will still depend on #623 (or the equivalent) for acceptance tests to pass in Travis, however, unless we explicitly force the python package in the Docker image.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -241,7 +241,7 @@ class rabbitmq::config {
           notify  => Exec['rabbitmq-systemd-reload'],
         }
         exec { 'rabbitmq-systemd-reload':
-          command     => '/usr/bin/systemctl daemon-reload',
+          command     => '/bin/systemctl daemon-reload',
           notify      => Class['Rabbitmq::Service'],
           refreshonly => true,
         }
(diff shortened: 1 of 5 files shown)


wyardley:
Travis test should pass after we merge #623 or similar.

manifests/config.pp
```diff
@@ -241,7 +241,7 @@
           notify  => Exec['rabbitmq-systemd-reload'],
         }
         exec { 'rabbitmq-systemd-reload':
-          command     => '/usr/bin/systemctl daemon-reload',
+          command     => '/bin/systemctl daemon-reload',
```
ekohl: A bit odd to see a RH fix when adding support for Debian but this is correct.

wyardley: It's because of the spec tests, both systems have it in `/bin/`, so once this change went in, the spec tests ran on Ubuntu 16 and failed because the path didn't match. But `/bin/` works on both systems, whereas `/usr/bin/` exists only on RHEL.

wyardley: @ekohl btw, re: camptocamp-systemd, it's on the roadmap for #617 in theory.... _but_:
a) if we just use it to do the reload, I think it's messy to add in that dependency just to do one exec... on the other hand
b) I investigated using it to manage the only file that's getting updated here, for the limits, and the module (as it exists now) doesn't seem to work in a way that easily supports this use case, because you can only set the number of files limit to an integer, whereas I believe 'infinity' is also a valid value. It also doesn't seem to accept a negative value, not sure if that's also valid, and `0` also didn't work (accepted the value, but the service wouldn't start).

So, it would add an extra dependency, ideally a soft one (which could come with its own headaches), but doesn't really solve our problem correctly yet. If the discussion about Vox taking over the module goes any further, or if Puppet itself starts supporting reloading the unit files (had a discussion with Hunner about making a PR against Puppet for that), that might be more reasonable, but as things are now, I think the module doesn't do what we need.

PR #630, opened 2017-09-12
Move facts back out of 'util' subdir, since they're not in lib/facter/util

wyardley:
I think these files should actually be in `spec/unit/facter`, not `spec/unit/facter/util` (probably my mistake).


wyardley:
```
spec/unit/facter/fact_erl_ssl_path_spec.rb:6:1: C: RSpec/FilePath: Spec path should end with facter/util/fact*_spec.rb.
describe Facter::Util::Fact do
^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/unit/facter/fact_rabbitmq_nodename_spec.rb:7:1: C: RSpec/FilePath: Spec path should end with facter/util/fact*_spec.rb.
describe Facter::Util::Fact do
^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/unit/facter/fact_rabbitmq_version_spec.rb:7:1: C: RSpec/FilePath: Spec path should end with facter/util/fact*_spec.rb.
describe Facter::Util::Fact do
^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Is this really correct? 
https://github.com/puppetlabs/puppetlabs-stdlib/tree/master/spec/unit/facter has the specs for `lib/puppet/facter` in `spec/unit/facter`

wyardley:
It looks like it's parsing out `describe Facter::Util::Fact` at the top; this is how the tests seem to be scoped in most of the Puppet repos (though they don't follow the file convention), but I guess this is probably Ok as it is.

PR #632, opened 2017-09-13
Add additional SSL configuration options (original PR from xepa)

wyardley:
Reworks #563


wyardley:
Also going to run the acceptance tests against CentOS 6 / RabbitMQ 3.1.x to make sure that these options don't cause fatal errors with older rmq versions.

spec/acceptance/class_spec.rb
```diff
@@ -154,6 +154,35 @@ class { 'rabbitmq':
...
+        node_ip_address => '0.0.0.0',
+        ssl_interface   => '0.0.0.0',
+        ssl             => true,
+        ssl_cacert      => '/tmp/cacert.crt',
+        ssl_cert        => '/tmp/rabbitmq.crt',
+        ssl_key         => '/tmp/rabbitmq.key',
+      }
+      EOS
+
+      apply_manifest(pp, catch_failures: true)
```
ekohl: Include an idempotency check?

wyardley: I switched to shared examples and added one on all the ones in the class spec (didn't add it for the other acceptance tests).

Not sure if that's going to slow things down by running the manifests twice in some places where it ran once before, but it is probably sensible. We could easily add the shared example to the other acceptance tests if we decided to later.

I also simplified the spec_helper_acceptance using the module helper.

Issue #640, opened 2017-09-26
Please switch to Package Cloud apt repository; rabbitmq.com's one becomes read-only in a few months


michaelklishin:
This module currently uses Package Cloud for RHEL-based distributions and rabbitmq.com apt repo for Debian-based ones.

Team RabbitMQ [plans to stop publishing new releases to rabbitmq.com in a few months](https://groups.google.com/forum/#!topic/rabbitmq-users/9ASVq55wXnA) (note: existing packages won't be removed). 3.7.0 will
be distributed via [Bintray](https://bintray.com/rabbitmq), [GitHub](https://github.com/rabbitmq/rabbitmq-server/releases), and [Package Cloud](https://packagecloud.io/rabbitmq/). It would be great if this module was updated
to use those three ahead of the 3.7.0 release in late November.

wyardley:
@michaelklishin Thanks for the heads up. I'm happy to update apt to use packagecloud when `repos_ensure` is set (keep in mind that it's disabled by default, so the module's default behavior is to use the vendor package).

Is this how we'd want the resulting repo to look on Ubuntu 16?
```
deb https://packagecloud.io/rabbitmq/rabbitmq-server/ubuntu xenial main
```

Also, will the signing key [for the package itself, vs. the repo, which I think uses https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey] continue to be at:
https://www.rabbitmq.com/rabbitmq-release-signing-key.asc

Thanks for the heads up.

wyardley:
Just in a quick test, seeing this:
```
root@ubuntu-server-1604-x64:/tmp# apt-key list
/etc/apt/trusted.gpg
--------------------
[...]
pub   4096R/6026DFCA 2016-05-17
uid                  RabbitMQ Release Signing Key <info@rabbitmq.com>
sub   4096R/12EBCE19 2016-05-17

pub   4096R/D59097AB 2014-01-13 [expires: 2019-01-12]
uid                  packagecloud ops (production key) <ops@packagecloud.io>
sub   4096R/7AD95B3F 2014-01-13 [expires: 2019-01-12]
```

yet:
```
root@ubuntu-server-1604-x64:/tmp# apt-get install rabbitmq-server
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  socat
The following NEW packages will be installed:
  socat
The following packages will be upgraded:
  rabbitmq-server
[...]

PR #641, opened 2017-09-26
Rework apt to use packagecloud repos as well (#640)

wyardley:
Could use some feedback on this, especially in terms of whether the packagecloud repo key itself needs to be configurable, and whether using the mit pgp server is the best way to get the key (it seems safer than grabbing it from the packagecloud site directly).

I couldn't find an easy way around doing variable interpolation in the repo source in the spec tests, but open to suggestions / examples of how to do it better.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -137,8 +137,8 @@
 # @param package_apt_pin Whether to pin the package to a particular source
 # @param package_ensure Determines the ensure state of the package.  Set to installed by default, but could be changed to latest.
 # @param package_gpg_key RPM package GPG key to import. Uses source method. Should be a URL for Debian/RedHat OS family, or a file name for
-#  RedHat OS family. Set to http://www.rabbitmq.com/rabbitmq-signing-key-public.asc by default. Note, that `key_content`, if specified,
-#  would override this parameter for Debian OS family.
+#  RedHat OS family. Set to https://packagecloud.io/gpg.key by default. Note, that `key_content`, if specified, would override this
+#  parameter for Debian OS family.
 # @param package_name The name of the package to install.
 # @param package_provider What provider to use to install the package.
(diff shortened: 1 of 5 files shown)


bastelfreak:
is packagecloud a third party service that generates the packages, or just a hosting platform used by the rabbitmq team? If it is the first option, the usage should be optional?

wyardley:
@bastelfreak in the case of `repos_ensure` being true (the only time this class should be called), we configure the official repo rather than the OS one or other user-configured ones. The linked issue has more detail, but they were both self-hosting and hosting via Packagecloud for a while (but without a proper yum repo), so we switched yum to use packagecloud when `repos_ensure` is set. Now, they're saying that they'll be dropping the self-hosted packages, so the packages are built (and signed) by RabbitMQ, but pushed to packagecloud, which hosts the repo and signs the repo metadata.

rspec doesn't yet have a `with_server` matcher, so need to check how recently that option was added to `apt::key`. Using a separate source from the download itself seems better to me, security-wise, but not sure what the min required version of apt is to use this style.

wyardley:
From https://github.com/puppetlabs/puppetlabs-apt/blob/master/CHANGELOG.md, it seems as if it used to be called `key_server`, but it has a default, so we should just take that; I'll take out the `server` param entirely in the call.

bastelfreak:
This fixes https://github.com/voxpupuli/puppet-rabbitmq/issues/640

bastelfreak:
@wyardley thanks for the explanation, I didn't see the issue :(

michaelklishin:
A couple of clarifications: Package Cloud is just a package distribution service. Team RabbitMQ publishes packages our build infra produces there. We won't be dropping the existing repo on rabbitmq.com but it will become a read-only archive.

wyardley:
Updated with some feedback from the comments. This seems to work for me, and, as a bonus, I think actually makes more sense as well.

PR #645, opened 2017-10-09
Make ldap_user_dn_pattern optional

sfhardman:
This change makes ldap_user_dn_pattern optional.

Per the [docs](https://www.rabbitmq.com/ldap.html):

> If you set both dn_lookup_attribute and user_dn_pattern then the approaches are combined: the plugin fills out the template and then searches for the DN

Sometimes this behaviour isn't wanted and user_dn_pattern needs to be omitted.


wyardley:
This looks good to me.

PR #648, opened 2017-10-12
Add optional variables for SSL management-console

### LINKED ISSUE #564: add vars to manage management ssl certs separately.
Add some new vars to specify the management ssl certs separately
Then just default the vars to the current ssl_key/cacert/cert options
e.g 
 {cacertfile, "<%= @management_ssl_cacert %>"},
                  <%- end -%>
                  {certfile, "<%= @management_ssl_cert %>"},
                  {keyfile, "<%= @management_ssl_key %>"}

Slm0n87:
Fixes #564

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -47,6 +47,9 @@ class rabbitmq::config {
   $ssl_port                            = $rabbitmq::ssl_port
   $ssl_interface                       = $rabbitmq::ssl_interface
   $ssl_management_port                 = $rabbitmq::ssl_management_port
+  $ssl_management_cacert               = $rabbitmq::ssl_management_cacert
+  $ssl_management_cert                 = $rabbitmq::ssl_management_cert
+  $ssl_management_key                  = $rabbitmq::ssl_management_key
   $ssl_management_verify               = $rabbitmq::ssl_management_verify
   $ssl_management_fail_if_no_peer_cert = $rabbitmq::ssl_management_fail_if_no_peer_cert
   $ssl_stomp_port                      = $rabbitmq::ssl_stomp_port
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -238,6 +238,12 @@
 #   port => undef
 # @param ssl_management_port
 #   SSL management port.
+# @param ssl_management_cacert
+#   SSL management cacert. If unset set to ssl_cacert for backwards compatibility.
+# @param ssl_management_cert
+#   SSL management cert. If unset set to ssl_cert for backwards compatibility.
+# @param ssl_management_key
+#   SSL management key. If unset set to ssl_key for backwards compatibility.
 # @param ssl_port
 #   SSL port for RabbitMQ
 # @param ssl_reuse_sessions
@@ -345,6 +351,9 @@ class rabbitmq(
   Integer[1, 65535] $ssl_port                                                                      = 5671,
   Optional[String] $ssl_interface                                                                  = undef,
   Integer[1, 65535] $ssl_management_port                                                           = 15671,
+  Optional[Stdlib::Absolutepath] $ssl_management_cacert                                            = $ssl_cacert,
+  Optional[Stdlib::Absolutepath] $ssl_management_cert                                              = $ssl_cert,
+  Optional[Stdlib::Absolutepath] $ssl_management_key                                               = $ssl_key,
   Integer[1, 65535] $ssl_stomp_port                                                                = 6164,
   Enum['verify_none','verify_peer'] $ssl_verify                                                    = 'verify_none',
   Boolean $ssl_fail_if_no_peer_cert                                                                = false,
--- a/templates/rabbitmq.config.erb
+++ b/templates/rabbitmq.config.erb
@@ -110,11 +110,11 @@
   <%- end -%>
       {port, <%= @ssl_management_port %>},
       {ssl, true},
-      {ssl_opts, [<%- if @ssl_cacert %>
-                  {cacertfile, "<%= @ssl_cacert %>"},
+      {ssl_opts, [<%- if @ssl_management_cacert %>
+                  {cacertfile, "<%= @ssl_management_cacert %>"},
                   <%- end -%>
-                  {certfile, "<%= @ssl_cert %>"},
-                  {keyfile, "<%= @ssl_key %>"},
+                  {certfile, "<%= @ssl_management_cert %>"},
+                  {keyfile, "<%= @ssl_management_key %>"},
                   {verify,<%= @ssl_management_verify %>},
                   {fail_if_no_peer_cert,<%= @ssl_management_fail_if_no_peer_cert %>}
                    <%- if @ssl_versions -%>
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -1240,6 +1240,39 @@ describe 'rabbitmq' do
         end
       end
 
+      describe 'ssl admin options with dedicated admin-key and -certs' do
+        let(:params) do
+          { ssl: true,
+            ssl_management_port: 3141,
+            ssl_management_cacert: '/path/to/management_cacert',
+            ssl_management_cert: '/path/to/management_cert',
+            ssl_management_key: '/path/to/management_key',
+            admin_enable: true }
+        end
+
+        it 'sets rabbitmq_management ssl options to specified values' do
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{rabbitmq_management, \[})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{listener, \[})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{port, 3141\},})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{ssl, true\},})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{ssl_opts, \[})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{cacertfile, "/path/to/management_cacert"\},})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{certfile, "/path/to/management_cert"\},})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{keyfile, "/path/to/management_key"\}})
+        end
+        it 'sets ssl options in the rabbitmqadmin.conf' do
+          is_expected.to contain_file('rabbitmqadmin.conf').with_content(
+            %r{ssl_ca_cert_file\s=\s/path/to/management_cacert}
+          )
+          is_expected.to contain_file('rabbitmqadmin.conf').with_content(
+            %r{ssl_cert_file\s=\s/path/to/management_cert}
+          )
+          is_expected.to contain_file('rabbitmqadmin.conf').with_content(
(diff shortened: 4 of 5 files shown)


wyardley:
👍 from me, can you squash that second commit?

manifests/init.pp
```diff
@@ -243,7 +246,10 @@
   $ssl_port                                      = $rabbitmq::params::ssl_port,
   Optional[String] $ssl_interface                = undef,
   Integer $ssl_management_port                   = $rabbitmq::params::ssl_management_port,
-  Integer $ssl_stomp_port                                = $rabbitmq::params::ssl_stomp_port,
+  Optional[String]  $ssl_management_cacert       = $ssl_cacert,
```
wyardley: Not sure if this could introduce any weird behavior. This is undef by default, since the params you're referencing are undef, but if someone set `ssl_cacert` and tried to explicitly set `ssl_management_cacert` to `undef`, it wouldn't work. I think this is still probably the best and most logical behavior (to take those as defaults).

Slm0n87:
@wyardley thx for reviewing.  You are right, its was not possible to *not* set an CA Cert for the mgmnt.  
I solved this now by changing the Type to "String or Boolean".  
Now this Option can be disabled by just setting it to "false".

wyardley:
@Slm0n87 maybe you could make it undef, and then, _in_ the manifest, set it to the value of the other one?

I think setting it to false vs. undef is kind of an anti-pattern, @bastelfreak @alexjfisher - any thoughts on the best way to accomplish this?

manifests/init.pp
```diff
@@ -243,7 +247,10 @@
   $ssl_port                                      = $rabbitmq::params::ssl_port,
   Optional[String] $ssl_interface                = undef,
   Integer $ssl_management_port                   = $rabbitmq::params::ssl_management_port,
-  Integer $ssl_stomp_port                                = $rabbitmq::params::ssl_stomp_port,
+  Variant[Undef, Boolean, String] $ssl_management_cacert = $ssl_cacert,
+  Optional[String] $ssl_management_cert          = $ssl_cert,
```
bastelfreak: all params should be alligned at their =

bastelfreak:
@wyardley I agree, this is an antipattern and should be avoided.

wyardley:
I think you could do it this way:

1. default the params to undef
1. if `management_ssl` is true, and the params are undef, set `$_ssl_management_cacert` etc. to `$rabbitmq::ssl_cacert` (in the manifest, not in the class params)
1. if `management_ssl` is true and param is not undef, use that value instead.

That make sense?

wyardley:
See also some changes in #649 that may affect this or require a rebase. Basically, I think the cert params should be `Optional[Stdlib::Absolutepath]` vs just `Optional[String]`

manifests/init.pp
```diff
@@ -243,7 +247,10 @@
   $ssl_port                                      = $rabbitmq::params::ssl_port,
   Optional[String] $ssl_interface                = undef,
   Integer $ssl_management_port                   = $rabbitmq::params::ssl_management_port,
-  Integer $ssl_stomp_port                                = $rabbitmq::params::ssl_stomp_port,
+  Variant[Undef, Boolean, String] $ssl_management_cacert = $ssl_cacert,
```
alexjfisher: I think there's some issue trying to default one parameter based on another like this.  @bastelfreak this came up on IRC IIRC??

alexjfisher: Actually, since puppet 4.4.0 (I've been go back through the IRC logs), this should work. https://github.com/puppetlabs/puppet-specifications/blob/master/language/parameter_scope.md#the-default-expression-evaluation.

alexjfisher:
It took me a little while to figure out why the `false` option for `ssl_management_cacert` was needed, but now I have, I actually think the code is OK.

If I understand correctly, you need to default all the parameters to the existing ones for backwards compatibility and because that's what most people will want to do anyway.
There are 2 ways of doing this.  You either default them to `undef` and do it in the manifest, or since puppet 4.4, you can just default the parameters to the others directly in the declaration.

But you also want to account for users who want *no* ssl_management_cacert file.  If you explicitly try to set the parameter to undef, nothing happens, you just get the parameter default.  So you've got to allow `false` instead.

wyardley:
> It took me a little while to figure out why the false option for ssl_management_cacert was needed, but now I have, I actually think the code is OK.

I think if the boolean `management_ssl` is true, then the various cert files are probably all required, no? So I'm not sure if there's a valid reason to suppress these. But I think in the case where you're doing it in the manifest, it should also be possible to have it allow `undef` and a valid path?

Slm0n87:
reopen Issie (accidentally closed by rebasing my forked master)...

Slm0n87:
I updated my patch once more by moving the overwrites for backwards compatibility into the manifest and set them to undef by default.

templates/rabbitmq.config.erb
```diff
@@ -109,11 +109,11 @@
   <%- end -%>
       {port, <%= @ssl_management_port %>},
       {ssl, true},
-      {ssl_opts, [<%- if @ssl_cacert %>
-                  {cacertfile, "<%= @ssl_cacert %>"},
+      {ssl_opts, [<%- if @_ssl_management_cacert %>
```
wyardley: I think regardless of how it's coded, `ssl_management_cacert` can't actually be undef, right? since it still defaults to `ssl_cacert` if it's not defined?

wyardley:
@Slm0n87 I think this will work, but I think @alexjfisher is right that this still won't allow actually suppressing it, because of the way the defaults work. Given that, I think the bit I commented on in the template is maybe extraneous? Also, the module supports >= 4.7.1 only, so maybe using the default in the params is cleaner after all, though I'm Ok with it as it is here.

Sorry for so much back and forth on this PR, and thanks for all the revisions.

Slm0n87:
@wyardley thx for the feedback - you are welcome. regarding the forth and back, I am aware of the conflict keeping a module most configurablae but backwards compatible :)  
So what do you think how we should finally resolve the ssl_management_cacerts? To make your condition in the template back working?  
Creating a new boolean variable $_ssl_management_cacert_enable which defaults to true?  
I can put the defaults back from the manifest to the declaration for cleaner code.

Slm0n87:
like this https://github.com/Slm0n87/puppet-rabbitmq/commit/3d5ba4b8c78d4d203319dae9e36445177da72290  
if its OK I would properly test it and add it to this PR tomorrow.

wyardley:
@Slm0n87 yeah, I think that makes some sense. Having a param was the way I was thinking of suggesting. The current template doesn't seem to support suppressing the cafile, so I think it could be Ok if that's not supported, but maybe you need that feature?

wyardley:
@Slm0n87 Great, two questions then:

1. Can you add spec tests?
1. Do you think we should mirror the same option for the non-management SSL part? Or is the need to suppress the ca file not applicable there?

Slm0n87:
I am not deep enough into spec tests to be able to write tests for my pullrequest. It would be perfect if someone could overtake this which also gives me a good example.  
2. We dont have the fallback case for the amqp ssl. So setting variable ssl_cacert = undef (which is the default) results into setting no ssl_cacert in the template.

manifests/config.pp
```diff
@@ -3,78 +3,82 @@
...
+  $ssl_only                     = $rabbitmq::ssl_only
+  $ssl_cacert                   = $rabbitmq::ssl_cacert
+  $ssl_cert                     = $rabbitmq::ssl_cert
+  $ssl_key                      = $rabbitmq::ssl_key
+  $ssl_depth                    = $rabbitmq::ssl_depth
+  $ssl_cert_password            = $rabbitmq::ssl_cert_password
+  $ssl_port                     = $rabbitmq::ssl_port
+  $ssl_interface                = $rabbitmq::ssl_interface
+  $ssl_management_port          = $rabbitmq::ssl_management_port
+  $ssl_management_cacert_enable = $rabbitmq::ssl_management_cacert_enable
```
bastelfreak: Can you please add datatypes for the new params?

wyardley: I think this is because they're not class params, but variables in `config.pp`. in `init.pp`, these all have data types. IIRC, you can't specify data types for variables vs. class params. I think they're not settable since they're in the body of the manifest, though we could also add an explicit `assert_private()` too maybe?

wyardley:
> I am not deep enough into spec tests to be able to write tests for my pullrequest. It would be perfect if someone could overtake this which also gives me a good example.

I can work on this at some point if I've got the time and submit a PR to your PR. FWIW, I think this is the section, so it might not be as bad as you think to make cases for the specific use cases. While I don't mind adding them, I think it can actually be a pretty useful exercise to add tests yourself.

https://github.com/voxpupuli/puppet-rabbitmq/blob/master/spec/classes/rabbitmq_spec.rb#L721-L1135
https://github.com/voxpupuli/puppet-rabbitmq/blob/master/.github/CONTRIBUTING.md
also has some info on running tests, and we're happy to help on Slack / IRC.

> 2. We dont have the fallback case for the amqp ssl. So setting variable ssl_cacert = undef (which is the default) results into setting no ssl_cacert in the template.

Ah, you're right, misread that before. So this solves the case of

Slm0n87:
- rebased & fix conflicts
- create spec tests for new options

@wyardley @bastelfreak @alexjfisher can you review again?

manifests/init.pp
```diff
@@ -238,6 +238,12 @@
 #   port => undef
 # @param ssl_management_port
 #   SSL management port.
+# @param ssl_management_cacert
+#   SSL management cacert. If unset set to ssl_cacert for backwards compatibility.
```
wyardley: How does this get defaulted when unset? Doesn’t seem to happen in either the variables or the template?

Slm0n87: **ssl_management_cacert** is set to **ssl_cacert** in init.pp.  
this keeps backwards compatibility and the new option to define an alternative management_cacert.

wyardley: I don’t see it being set further down here in `init.pp`, tho, just as an Optional param (which can’t have a default)

wyardley: Sorry, didn’t look far enough right. I don’t think it can be optional, though, then, because of quirks around how Puppet works with optional params with defaults.

manifests/init.pp
```diff
@@ -345,6 +351,9 @@
   Integer[1, 65535] $ssl_port                                                                      = 5671,
   Optional[String] $ssl_interface                                                                  = undef,
   Integer[1, 65535] $ssl_management_port                                                           = 15671,
+  Optional[Stdlib::Absolutepath] $ssl_management_cacert                                            = $ssl_cacert,
+  Optional[Stdlib::Absolutepath] $ssl_management_cert                                              = $ssl_cert,
+  Optional[Stdlib::Absolutepath] $ssl_management_key                                               = $ssl_key,
```
wyardley: @bastelfreak: can a param referring to another param that defaults to undef like this be optional?

Slm0n87: after the referrenced variables are undef by default, the new variables also has to be optional.  

puppet documentation "Optional":

> The Optional data type wraps one other data type, and results in a data type that matches anything that type would match plus undef.

https://puppet.com/docs/puppet/5.3/lang_data_abstract.html#optional

wyardley: Yeah, I just wasn’t sure if it would chain properly. Should be fine.

PR #651, opened 2017-10-13
Fix 'puppet resource rabbitmq_binding' and add tests (#650)

### LINKED ISSUE #650: Error on using puppet resource rabbitmq_binding
I'm pretty sure this is the same issue as #147, just for another type.
I also believe most types have a similar issue.

- Puppet: master: 4.10.8
- Ruby: 2.3.3p222
- Distribution: Debian stretch
- Module version: 7.1.0

Run command `puppet resource rabbitmq_binding`

## What are you seeing

Error: Could not run: Validation of Rabbitmq_binding[5538f6c889e4583bace88315c5502438ec434fde8e4f0f200a280c439ace909c] failed: Source and destination must both be defined.

Get the list of binding resources applied to the host.

wyardley:
This should fix #650. It also
- Reformats the binding provider spec tests to be a little easier to read
- Adds a basic test for `puppet resource rabbitmq_binding` in the acceptance tests. This way, we should notice if a problem comes up with this command in the future.

Since we got feedback on how to test this in the type rather than in the create method in the provider, I can try to submit a separate PR to fix this for the rabbitmq_user resource.

Some of the other providers don't have instances implemented, so won't work with `puppet resource`; this is a separate issue, though refactoring them would be a good project for someone with the time on their hands 😉

### DIFF
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -3,17 +3,13 @@ require 'puppet'
 require 'digest'
 
 Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
-  if Puppet::PUPPETVERSION.to_f < 3
-    commands rabbitmqctl: 'rabbitmqctl'
-    commands rabbitmqadmin: '/usr/local/bin/rabbitmqadmin'
-  else
-    has_command(:rabbitmqctl, 'rabbitmqctl') do
-      environment HOME: '/tmp'
-    end
-    has_command(:rabbitmqadmin, '/usr/local/bin/rabbitmqadmin') do
-      environment HOME: '/tmp'
-    end
+  has_command(:rabbitmqctl, 'rabbitmqctl') do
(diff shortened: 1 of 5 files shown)


wyardley:
Spot checked some resources, and these look good:

```[root@centos-7-x64 ~]# puppet resource rabbitmq_queue
rabbitmq_queue { 'queue1@host1':
  ensure => 'present',
}
rabbitmq_queue { 'queue2@host2':
  ensure => 'present',
}
[root@centos-7-x64 ~]# puppet resource rabbitmq_plugin
rabbitmq_plugin { 'rabbitmq_federation':
  ensure => 'present',
}
rabbitmq_plugin { 'rabbitmq_federation_management':
  ensure => 'present',
}
rabbitmq_plugin { 'rabbitmq_management':
  ensure => 'present',
}
```

let me know if there any that aren't working for extant resources.

wyardley:
@alexjfisher @ekohl: I think this fixes it in a way that's closer to the original, thanks to @hunner's suggestion.

`if !self[:source] && provider.source` errored for me, but
`if !self[:source] && !defined? provider.source` seems to work.

PR #656, opened 2017-10-23
add ability to have array as package name

tampakrap:
in case the package has subpackages (like it does for the SUSE family
OSes)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -138,7 +138,7 @@
 # @param package_gpg_key RPM package GPG key to import. Uses source method. Should be a URL for Debian/RedHat OS family, or a file name for
 #  RedHat OS family. Set to https://packagecloud.io/gpg.key by default. Note, that `key_content`, if specified, would override this
 #  parameter for Debian OS family.
-# @param package_name The name of the package to install.
+# @param package_name Name(s) of the package(s) to install
 # @param port The RabbitMQ port.
 # @param repos_ensure Ensure that a repo with the official (and newer) RabbitMQ package is configured, along with its signing key.
 #  Defaults to false (use system packages). This does not ensure that soft dependencies (like EPEL on RHEL systems) are present.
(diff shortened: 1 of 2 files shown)


wyardley:
@tampakrap this makes some sense to me, will ask @bastelfreak to review as well.
Would you be able / willing to add / update SUSE support (update metadata, update `params.pp`) in another PR?

How are the packages broken up -- could you give an example of the packages that would make this work on SUSE, and does it vary by version?

tampakrap:
Hello @wyardley thanks for the review.

Regarding packaging, there is a rabbitmq-server package with a -plugins subpackage, see https://build.opensuse.org/package/view_file/network:messaging:amqp/rabbitmq-server/rabbitmq-server.spec?expand=1

Regarding support, there is already support in params.pp and actually I use this module in production for quite some time now. I just noticed that SUSE is missing from metadata.json, I'd be happy to add it there as well in a separate PR

wyardley:
>I just noticed that SUSE is missing from metadata.json, I'd be happy to add it there as well in a separate PR

@tampakrap Yes, there's a section in params already. The module default is now to use vendor packages, so assuming we merge this to add support for multiple packages, I was suggesting that params should then be updated to install the correct packages (especially if the packages are broken up differently in different versions). Acceptance tests would be good too, though it doesn't look like we support SUSE acceptance tests currently (though you could propose the addition of some in https://github.com/voxpupuli/modulesync_config/tree/master/moduleroot/spec/acceptance/nodesets and https://github.com/voxpupuli/modulesync_config/tree/master/moduleroot/spec/acceptance/nodesets/docker)

Also, this PR should probably include a spec test for the multiple packages use case.

wyardley:
@tampakrap Can you update https://github.com/voxpupuli/puppet-rabbitmq/blob/master/manifests/init.pp#L141 ? puppet strings will update the type, but should probably clarify that this can be a package name or a list of package names, especially since the variable is singular.

tampakrap:
I added the subpackage in params.pp as @bastelfreak said which makes sens. @wyardley give me a hint on how this line should look like please and I'll do it as well

wyardley:
@tampakrap sorry for the delayed response. Strings will take care of showing the allowed types, so it would just be to something like this:
`# @param package_name Name(s) of the package(s) to install`

tampakrap:
thanks, I updated it

wyardley:
Merged, thanks for the contribution.

PR #657, opened 2017-10-24
Add options for ssl verify and 'fail_if_no_peer_cert' for the managem…

paebersold:
…ent ssl options


wyardley:
@paebersold This is sensibly implemented, thanks for submitting with strings docs, passing tests, etc.
Quick question - are there many other SSL related params? At some point, it would make more sense to have a generic options hash and pas in options that way.
I also find `ssl_management_fail_if_no_peer_cert` to be _really_ long, not sure if there's a way to make that param shorter but still understandable / consistent...

paebersold:
@wyardley my knowledge of rabbitmq is limited so most of this is based off 
https://www.rabbitmq.com/ssl.html and
https://www.rabbitmq.com/configure.html

At the top scope in the config file there are two options missing from the current rabbitmq.config.erb template
- num_ssl_acceptors
- ssl_handshake_timeout

From the ssl_options scope missing is
- honor_ecc_order

Sorry about the long variable name, getting to be more of a short story.  For clarity I went for keeping it consistent with the existing variable.

wyardley:
> Sorry about the long variable name, getting to be more of a short story. For clarity I went for keeping it consistent with the existing variable.

Yes, I think you're right. Can you update the param documentation following the existing convention?

PR #661, opened 2017-11-02
Allow sensitive type for default_pass parameter

wyardley:
I think this is required for now, at least, to allow setting `default_pass` as a `Sensitive` parameter.
So far, have not found any guidance on allowing this in types / providers without using private classes from Puppet's Ruby code, though we should look into doing it that way if we need to.


wyardley:
Apparently the long-term plan is to allow passing in `Sensitive[T]` or `T` to type `T`, but this is necessary now if you want to pass in type Sensitive.
I don't think this is possible to test yet, until this is handled? https://github.com/rodjek/rspec-puppet/issues/529

wyardley:
Going to close this for now, until the situation is sorted out with how sensitive types are passed in (I'm not as worried about the testing in this case, but if the eventual plan is to allow passing in Sensitive[T] to T, this shouldn't be needed.

PR #669, opened 2017-12-04
Remove `archive require` in rabbitmqadmin class

lzecca78:
due to the dependency of the module from `puppet-archive`, there is no need to make the require. Is also really hard to debug, because the puppet exception doesn't clearly refers to the module requirement (duplicate declaration)

### DIFF
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -7,7 +7,6 @@ class rabbitmq::install::rabbitmqadmin {
       name   => $rabbitmq::rabbitmqadmin_package,
     }
   } else {
-    require archive
 
     $python_package = $rabbitmq::params::python_package
     # Some systems (e.g., Ubuntu 16.04) don't ship Python 2 by default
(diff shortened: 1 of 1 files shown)


wyardley:
[edited your post to include the text from your commit message]
Can you show the error you're getting?

If not using `puppet module install`, the module won't be installed, so I personally think the require is sensible.

Also,  in theory, archive could / should become a  soft dep now that some distributions are packaging it (though since the common distros don't, probably won't do this now).

lzecca78:
@wyardley , i think that is pointless the line `require archive` because the module already declares its dependency from puppet-archive. For this reason, if the module is missing, will broke both the archive{} define and the empy `require archive` as well. On top of this, there is no need to require archive without any further reasons like to call a non-default class parameter (for example: class{'::archive': aws_cli_install => true}). I lost half of my day to understand who was the other resource responsable of the duplicate declaration class, and at the end, i found it. I can understand if it was required for the logic above, but it isn't, because the resource defined will work also without the require.

wyardley:
After some more discussion, it seems clear that `require` doesn't add much here, so I'm going to go ahead and merge this.
Thanks for the contribution.

lzecca78:
thank you  @wyardley ! i think that this will save a lot of troubles to current and future users of this module!

PR #672, opened 2017-12-07
Do not use defaultfor to choose the only existing provider

mbaldessari:
It is sometimes useful to be able to noop (or change) a provider in
certain situations [1] (Think running a temporary container with puppet
tags to create proper config files). Currently, using defaultfor makes
overriding the current provider rabbitmqctl a very hard thing to do.

Let's switch to using the confine mechanism to make sure the provider
runs only on posix OSes while still allowing an operator to override it.

[1] https://github.com/openstack/puppet-tripleo/blob/master/lib/puppet/parser/functions/noop_resource.rb

### DIFF
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -10,7 +10,7 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin) do
     environment HOME: '/tmp'
   end
 
-  confine :feature => :posix
+  confine feature: :posix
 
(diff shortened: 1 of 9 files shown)


lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -10,7 +10,7 @@
     environment HOME: '/tmp'
   end
 
-  defaultfor feature: :posix
+  confine :feature => :posix
```
EmilienM: the linter isn't happy:
https://travis-ci.org/voxpupuli/puppet-rabbitmq/jobs/312947453

Maybe we should do:
`confine feature: :posix`

Issue #671, opened 2017-12-06
not working with rabbitmq-server 3.7 - Cannot parse invalid user line


heap98:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.3.3-1xenial
- Ruby: ruby2.4 (included in puppet package)
- Distribution: Ubuntu 16.04.3 LTS
- Module version: v8.0.0
- Rabbitmq-server : 3.7.0-1

## How to reproduce (e.g Puppet code you use)
just run
puppet agent -t

```puppet
      rabbitmq_user { 'new_user':
                admin    => true,
                password => 'pass',
                provider => 'rabbitmqctl'
        }
```

## What are you seeing
```
Error: Failed to apply catalog: Cannot parse invalid user line: warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell)
```
## What behaviour did you expect instead
something went wrong with name encoding when the module is trying to fetch the result of the command: 
/usr/sbin/rabbitmqctl -q list_users

the function itself is located in this file line 22:
./lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb

## Output log
```
Debug: Executing: '/usr/sbin/rabbitmqctl -q list_users'
Debug: Command succeeded
Debug: Storing state
Debug: Stored state in 0.16 seconds
Error: Failed to apply catalog: Cannot parse invalid user line: warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell)
Debug: Dynamically-bound server lookup failed, falling back to report_server setting
```

## Any additional information you'd like to impart

wyardley:
@heap98:
What's the output of `rabbitmqctl -q list_users`? Does it have that same warning:
`warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell)`?

heap98:
puppet runs as user root. output of the command is following and looks fine.

> root@rabbit01:~# rabbitmqctl -q list_users
guest	[administrator]
> 



`update-locale  LC_ALL=en_US.UTF-8`

doesn't fix the issue.


output von locale:

>[root@rabbit01 ~]$ locale
[...]

PR #674, opened 2017-12-27
Change policy regex to match Rabbitmq v3.7.0 syntax

alexizmailov:
Solves this - https://github.com/voxpupuli/puppet-rabbitmq/issues/671

RMQ < 3.7.0 shows policies as:
`/ ha-all all .* {"ha-mode":"all","ha-sync-mode":"automatic"} 0`

V3.7.0 returns this:
`/ ha-all  .*  all {"ha-mode":"all","ha-sync-mode":"automatic"}  0`

`.*`  and  `all` are swapped and regex needs to be updated.

Also corrected a typo - `vhost` and `name` were swapped, accidentally, I suppose.

### DIFF
--- a/lib/puppet/provider/rabbitmq_policy/rabbitmqctl.rb
+++ b/lib/puppet/provider/rabbitmq_policy/rabbitmqctl.rb
@@ -6,7 +6,7 @@ Puppet::Type.type(:rabbitmq_policy).provide(:rabbitmqctl, parent: Puppet::Provid
   confine feature: :posix
 
   # cache policies
-  def self.policies(name, vhost)
+  def self.policies(vhost, name)
     @policies = {} unless @policies
     unless @policies[vhost]
       @policies[vhost] = {}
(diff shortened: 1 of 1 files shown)


alexizmailov:
No idea why those checks failed, looks like the reason is unrelated.

wyardley:
What about just making a looser regex that matches both ways? I assume the `versioncmp` should already have been computed by the time this can successfully run, but still seems riskier to rely on the fact. Maybe something like (untested): ` /^\s+ha-all.*all.* {"ha-mode":"all","ha-sync-mode":"automatic"} 0`
I think there should probably also be spec tests for both possible formats to avoid a regression later?

I don't think the test failures (at least the 5.0 ones) are unrelated, probably the spec test needs to be adjusted somewhat to pass?

```
  2) Puppet::Type::Rabbitmq_policy::ProviderRabbitmqctl matches policies from list (>=3.2.0)
     Failure/Error: output = rabbitmqctl('-q', 'status')
     
     Mocha::ExpectationError:
       unexpected invocation: rabbitmqctl(-q, status)
```
https://travis-ci.org/voxpupuli/puppet-rabbitmq/jobs/322395359

alexizmailov:
I don't think it would work because the regex is not the only problem. 
If I just expand that regex - `applyto` and `pattern` will get wrong values in v3.7.0. 
They also need to be swapped.

```
# <3.7.0
24: applyto    = Regexp.last_match(4) || 'all'
29: pattern    = Regexp.last_match(3).to_s.gsub(%r{\\\\}, '\\')

# >3.7.0
33: applyto    = Regexp.last_match(3) || 'all'
38: pattern    = Regexp.last_match(4).to_s.gsub(%r{\\\\}, '\\')
```

fatmcgav:
@alexizmailov I took the liberty of tidying up the code and fixing the failing tests. 
My fixes can be found here: https://github.com/fatmcgav/puppet-rabbitmq/commit/c86e325c33fe52bb3a5e83e7ce9a8b8a19d2f379

I can either PR them against your branch, or open a fresh PR with both changes. 

**N.B:** There's one failing test on the following policy at 3.7.0, which I need to validate. Unfortunately I'm struggling to find any documentation on the format of the expected return, so not sure if the exchange policies change at 3.7.0 aswell. 
```
/ test exchanges .* {"ha-mode":"all"} 0
```

alexizmailov:
@fatmcgav Thank you very much. I'm not really proficient in Ruby and I didn't know it's possible to do that trick with array indexes.
Please, do whatever suits you best.

fatmcgav:
@alexizmailov No worries... I've opened #676 which incorporates both sets of changes :)

wyardley:
@alexizmailov thanks so much for figuring out this issue and submitting the initial work. I'm going to close this in favor of #676, which adds a new test case, and passes tests.

PR #678, opened 2018-02-07
Add in download_options to Archive to all passing of extra params suc…

paebersold:
…h as ssl options

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -277,7 +277,7 @@ class rabbitmq(
   Stdlib::Absolutepath $inetrc_config_path                         = $rabbitmq::params::inetrc_config_path,
   Boolean $ssl_erl_dist                                            = $rabbitmq::params::ssl_erl_dist,
   Optional[String] $rabbitmqadmin_package                          = $rabbitmq::params::rabbitmqadmin_package,
-  Optional[Array] mqa_download_opts                                = $rabbitmq::params::mqa_download_opts,
+  Optional[Array] $mqa_download_opts                               = $rabbitmq::params::mqa_download_opts,
 ) inherits rabbitmq::params {
 
   if $ssl_only and ! $ssl {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -277,7 +277,7 @@ class rabbitmq(
   Stdlib::Absolutepath $inetrc_config_path                         = $rabbitmq::params::inetrc_config_path,
   Boolean $ssl_erl_dist                                            = $rabbitmq::params::ssl_erl_dist,
   Optional[String] $rabbitmqadmin_package                          = $rabbitmq::params::rabbitmqadmin_package,
-  [Array] $archive_options                                         = $rabbitmq::params::archive_options,
+  Array $archive_options                                         = $rabbitmq::params::archive_options,
 ) inherits rabbitmq::params {
 
   if $ssl_only and ! $ssl {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -185,6 +185,7 @@
 # @param rabbitmq_group OS dependent, default defined in param.pp. The system group the rabbitmq daemon runs as.
 # @param rabbitmq_home OS dependent. default defined in param.pp. The home directory of the rabbitmq deamon.
 # @param $rabbitmqadmin_package OS dependent. default defined in param.pp. If undef: install rabbitmqadmin via archive, otherwise via package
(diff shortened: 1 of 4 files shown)


paebersold:
Sorry, not sure how to get pass the rubocop test as the new test follows the same format as existing tests.

manifests/init.pp
```diff
@@ -185,6 +185,7 @@
 # @param rabbitmq_group OS dependent, default defined in param.pp. The system group the rabbitmq daemon runs as.
 # @param rabbitmq_home OS dependent. default defined in param.pp. The home directory of the rabbitmq deamon.
 # @param $rabbitmqadmin_package OS dependent. default defined in param.pp. If undef: install rabbitmqadmin via archive, otherwise via package
+# @param $mqa_download_opts. default defined in params.pp.  Extra options to pass to Archive resource to download rabbitmqadmin
```
wyardley: Maybe `archive_options` would be more clear? The strings docs will add the default, so, even while this text is in some of the other lines, that bit is not actually needed.

manifests/init.pp
```diff
@@ -276,6 +277,7 @@
   Stdlib::Absolutepath $inetrc_config_path                         = $rabbitmq::params::inetrc_config_path,
   Boolean $ssl_erl_dist                                            = $rabbitmq::params::ssl_erl_dist,
   Optional[String] $rabbitmqadmin_package                          = $rabbitmq::params::rabbitmqadmin_package,
+  Optional[Array] $mqa_download_opts                               = $rabbitmq::params::mqa_download_opts,
```
wyardley: This should either be set to `undef` and left directly here (vs. coming from params), or shouldn't be optional. Except when using hiera, Puppet won't let you set a default for an optional type for complicated (and historical) reasons, so an Optional param with a default is not a good idea.

spec/classes/rabbitmq_spec.rb
```diff
@@ -287,6 +287,21 @@
             )
           end
         end
+        context 'with service_manage set to true and mqa_download_opts specified' do
+          let(:params) do
+            {
+              admin_enable: true,
+              mqa_download_opts: ['blah', 'blig']
```
wyardley:
```
spec/classes/rabbitmq_spec.rb:301:33: C: Style/WordArray: Use %w or %W for an array of words. (b)
              download_options: ['blah', 'blig']
                                ^^^^^^^^^^^^^^^^
```

You just need to do `%w(blah blig)` instead. You can run rubocop locally with `bundle exec rake rubocop` if you want to make sure rubocop passes before pushing and having Travis run again.

spec/classes/rabbitmq_spec.rb
```diff
@@ -287,6 +287,21 @@
...
+            }
+          end
+
+          it 'we use the correct extra options to download rabbitmqadmin' do
+            is_expected.to contain_archive('rabbitmqadmin').with(
+              source: 'http://127.0.0.1:15672/cli/rabbitmqadmin',
+              download_options: ['blah', 'blig']
```
wyardley: same as above (re: rubocop).

manifests/install/rabbitmqadmin.pp
```diff
@@ -28,6 +28,8 @@
     $default_user = $rabbitmq::default_user
     $default_pass = $rabbitmq::default_pass
     $management_ip_address = $rabbitmq::management_ip_address
+    $extra_opts = $rabbitmq::archive_options
```
wyardley: Here, I would use the convention of keeping the same param name,
`archive_options = $rabbitmq::archive_options` (and then using that below).

manifests/init.pp
wyardley: This needs to be updated too (to match the other places).

manifests/init.pp
```diff
@@ -276,6 +277,7 @@
   Stdlib::Absolutepath $inetrc_config_path                         = $rabbitmq::params::inetrc_config_path,
   Boolean $ssl_erl_dist                                            = $rabbitmq::params::ssl_erl_dist,
   Optional[String] $rabbitmqadmin_package                          = $rabbitmq::params::rabbitmqadmin_package,
+  [Array] $archive_options                                         = $rabbitmq::params::archive_options,
```
wyardley: Take out the leading parens when you get rid of `Optional`, so `Array $archive_options`

paebersold: yep - sorry about that typo

manifests/install/rabbitmqadmin.pp
paebersold: shall do

wyardley:
Continued in #681

PR #679, opened 2018-02-07
Add support for multiple LDAP servers

skrussell:
The RabbitMQ LDAP plug-in supports multiple servers being listed in the the "servers" field, but the puppet module currently only accepts a single String value.

My simple pull request allows an array of servers to be passed, which is then correctly formatted in the rabbitmq.confg template. I additionally attempt to detect an array that has already been flattened to a string, as happens if you configure the ldap_servers variable from hiera by way of sourcing an array variable that is already defined else where in puppet (rabbitmq::ldap_server: "%{ldap::servers}").

Tested and working with a string, an array, and a array flattened to a string.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -157,7 +157,7 @@
 # @param ldap_auth
 #   Set to true to enable LDAP auth.
 # @param ldap_server
-#   LDAP server to use for auth.
+#   LDAP server or servers to use for auth.
 # @param ldap_user_dn_pattern
 #   User DN pattern for LDAP auth.
 # @param ldap_other_bind
@@ -370,7 +370,7 @@ class rabbitmq (
   Array $ssl_ciphers                                                                               = [],
   Boolean $stomp_ensure                                                                            = false,
   Boolean $ldap_auth                                                                               = false,
-  String $ldap_server                                                                              = 'ldap',
+  Variant[String[1],Array[String[1]]] $ldap_server                                                 = 'ldap',
   Optional[String] $ldap_user_dn_pattern                                                           = undef,
   String $ldap_other_bind                                                                          = 'anon',
   Boolean $ldap_use_ssl                                                                            = false,
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -596,6 +596,32 @@ describe 'rabbitmq' do
         end
       end
 
+      describe 'configuring ldap authentication' do
+        let :params do
+          { config_stomp: false,
+            ldap_auth: true,
+            ldap_server: ['ldap1.example.com', 'ldap2.example.com'],
+            ldap_other_bind: 'as_user',
+            ldap_use_ssl: false,
+            ldap_port: 389,
+            ldap_log: true,
+            ldap_config_variables: { 'foo' => 'bar' } }
+        end
+
+        it { is_expected.to contain_rabbitmq_plugin('rabbitmq_auth_backend_ldap') }
+
+        it 'does not set user_dn_pattern when none is specified' do
+          verify_contents(catalogue, 'rabbitmq.config',
+                          ['[', '  {rabbit, [', '    {auth_backends, [rabbit_auth_backend_internal, rabbit_auth_backend_ldap]},', '  ]}',
+                           '  {rabbitmq_auth_backend_ldap, [', '    {other_bind, as_user},',
+                           '    {servers, ["ldap1.example.com", "ldap2.example.com"]},',
+                           '    {use_ssl, false},',
+                           '    {port, 389},', '    {foo, bar},', '    {log, true}'])
+          content = catalogue.resource('file', 'rabbitmq.config').send(:parameters)[:content]
+          expect(content).not_to include 'user_dn_pattern'
+        end
+      end
+
       describe 'configuring auth_backends' do
         let :params do
           { auth_backends: ['{baz, foo}', 'bar'] }
--- a/templates/rabbitmq.config.erb
+++ b/templates/rabbitmq.config.erb
@@ -167,7 +167,11 @@ end
 % Configure the LDAP authentication plugin
   {rabbitmq_auth_backend_ldap, [
     {other_bind, <%= @ldap_other_bind %>},
+<% if @ldap_server.class == Array -%>
+    {servers, <%= @ldap_server %>},
+<% else -%>
     {servers, ["<%= @ldap_server %>"]},
+<% end -%>
 <% if @ldap_user_dn_pattern -%>
     {user_dn_pattern, "<%= @ldap_user_dn_pattern %>"},
 <%- end -%>


wyardley:
I think is a great idea. I'm a bit on the fence about having it allow "array-in-string-form", maybe cleaner to force it to be a hostname or an array of hostnames?

This probably needs to have test coverage added.

templates/rabbitmq.config.erb
```diff
@@ -153,7 +153,11 @@
 % Configure the LDAP authentication plugin
   {rabbitmq_auth_backend_ldap, [
     {other_bind, <%= @ldap_other_bind %>},
+<% if @ldap_server.class == Array or @ldap_server =~ /^\[/  -%>
```
wyardley: I usually see `@foo.is_a?(Array)` instead?
As mentioned in the comments, my personal take is that it's probably better not to support both an array and a string formatted as an array, and then simplify this to a single bit in the template, like
`{servers, [<%= @ldap_server.map { |v| "#{v}" }.join(', ') %>]},`

skrussell: That sounds okay to me, however I will need to retain the "flattened string" detection (and therefore maintain my own fork) for now as our use case is based around the fact that hiera squashes all data types to a string (https://tickets.puppetlabs.com/browse/PUP-7675); though I do appreciate that you won't necessarily want that in a released version. Any advice on how to get around that problem would be greatly appreciated though...

wyardley: @skrussell I've never had problems defining arrays in hiera. Should be just like:
```
rabbitmq::ldap_server:
  - server1.example.com
  - server2.example.com
```

skrussell: Yup, and that's how I've tested it, however, if you have a module that dynamically defines a variable somewhere, like we have, and then you use the heira scope variable lookup, then the array is flattened to a string.

So, if you have this manifest (this is a simplified example, but assume the servers variable is dynamically populated for now):
{code}
class ldap {
  $servers = [ 'svr1.example.com', 'svr2.example.com' ]
}
{code}
and in hiera:
{code}
rabbitmq::ldap_server: %{ldap::servers}
{code}
The value passed to the rabbitmq parameter gets squashed to a string rather than being passed as an array. This is a limitation of the scope() function that is being used by hiera, and will hopefully be fixed in a future release, but until this, this is a pain for us as we want to be able to dynamically allocate (the nearest) ldap servers without statically defining lists per node in Hiera.

vStone: I use `[@ldap_servers].flatten` quite a lot to support both string and array[string]

skrussell: I refactored our code to remove the need to support an array as string, so now we just use a straight forward array (like was originally suggested/requested).

If we could look again at merging this for the next release, that would be great.

wyardley:
Can you rebase?

skrussell:
@wyardley - sure. Done (hopefully correctly).

wyardley:
@skrussell yes, rebase looks good, but can you squash down to 1 commit too? (https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit)

wyardley:
Also, can you add a test case, similar to the ones starting here:
https://github.com/voxpupuli/puppet-rabbitmq/blob/master/spec/classes/rabbitmq_spec.rb#L519
basically copy / paste, but using the array and then verifying that the rendered values are as expected?

manifests/init.pp
```diff
@@ -353,7 +353,7 @@
   Array $ssl_ciphers                                                                               = $rabbitmq::params::ssl_ciphers,
   Boolean $stomp_ensure                                                                            = $rabbitmq::params::stomp_ensure,
   Boolean $ldap_auth                                                                               = $rabbitmq::params::ldap_auth,
-  String $ldap_server                                                                              = $rabbitmq::params::ldap_server,
+  Variant[String,Array[String]] $ldap_server                                                                              = $rabbitmq::params::ldap_server,
```
bastelfreak: Please update it to this:
```suggestion
  Variant[String[1],Array[String[1]]] $ldap_server                                                             = $rabbitmq::params::ldap_server,
```

Also please ensure that the `=` are all aligned

dhoppe: @skrussell Could you please apply the suggestion of @bastelfreak . After that we should be able to merge this pull request.

manifests/init.pp
```diff
@@ -367,7 +367,7 @@
   Array $ssl_ciphers                                                                               = [],
   Boolean $stomp_ensure                                                                            = false,
   Boolean $ldap_auth                                                                               = false,
-  String $ldap_server                                                                              = 'ldap',
+  Variant[String,Array[String]] $ldap_server                                                       = 'ldap',
```
bastelfreak: You can even enforce the minimal length of the string to prohibit empty strings:
```suggestion
  Variant[String[1],Array[String[1]]] $ldap_server                                                 = 'ldap',
```

templates/rabbitmq.config.erb
```diff
@@ -154,7 +154,11 @@
 % Configure the LDAP authentication plugin
   {rabbitmq_auth_backend_ldap, [
     {other_bind, <%= @ldap_other_bind %>},
+<% if @ldap_server.class == Array -%>
```
bastelfreak: can you please add a unit test for this?

bastelfreak:
Hi @skrussell, can you please take a look at the comments and also rebase against our master branch? (don't do a merge commit please).

wyardley:
Hi @skrussell 
Thanks for rebasing this! Can you add a unit test for this (see @bastelfreak's request)?

skrussell:
> Hi @skrussell
> Thanks for rebasing this! Can you add a unit test for this (see @bastelfreak's request)?

I have made an attempt, by simply copying and pasting one of the examples, and editing it to have the multiple 'ldap_server' value. I really have no idea if that if the correct thing to do - I have never added a unit test for anything before, and have no such way to validate or test this.

spec/classes/rabbitmq_spec.rb
```diff
@@ -596,6 +596,32 @@
...
+            ldap_config_variables: { 'foo' => 'bar' } }
+        end
+
+        it { is_expected.to contain_rabbitmq_plugin('rabbitmq_auth_backend_ldap') }
+
+        it 'does not set user_dn_pattern when none is specified' do
+          verify_contents(catalogue, 'rabbitmq.config',
+                          ['[', '  {rabbit, [', '    {auth_backends, [rabbit_auth_backend_internal, rabbit_auth_backend_ldap]},', '  ]}',
+                           '  {rabbitmq_auth_backend_ldap, [', '    {other_bind, as_user},',
+                           '    {servers, ["ldap1.example.com", "ldap2.example.com"]},',
```
wyardley, line 617: I think you need to adjust the servers here to match the value in the array.

wyardley (commented):
Hi @skrussell 
thanks!
This looks about right to me, one minor comment in the review feedback, which should (I think) address the failures in CI. I'd also run a `bundle exec rake rubocop` ahead of pushing up just to make sure it'll pass rubocop checks (you can also run the unit test suite locally as well; how to do so is documented somewhere; it's optional, but may speed up the feedback loop).

Also, feel free to reach out on Voxpupuli's IRC channels if you need any additional help getting the tests to work.

Once the checks are passing in CI, and you've squashed the commits in your branch, I think this should be good to go!

spec/classes/rabbitmq_spec.rb
skrussell: Thanks - my bad… I spend most of my day using that domain name that it was simply a case of my fingers being on autopilot. Corrected and pushed now.

wyardley:
Looks good. Can you squash the two commits? Looks good

kenyon:
> Looks good. Can you squash the two commits?

Squash can be done on merge too.

PR #685, opened 2018-03-07
Fixes for Archlinux and modulesync 1.8

bastelfreak:

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -254,6 +254,27 @@ class rabbitmq::config {
         notify  => Class['Rabbitmq::Service'],
       }
     }
+    'Archlinux': {
+      file { '/etc/systemd/system/rabbitmq.service.d':
+        ensure                  => directory,
+        owner                   => '0',
+        group                   => '0',
+        mode                    => '0755',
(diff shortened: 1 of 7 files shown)


manifests/config.pp
```diff
@@ -254,6 +254,27 @@
         notify  => Class['Rabbitmq::Service'],
       }
     }
+    'Archlinux': {
```
bastelfreak: this isn't the cleanest solution, but it works good. Later on we can think about unifying this setup for all distributions with systemd. Personally I prefer the camptocamp/systemd module for this, but maybe we need to wait for the discussion in https://github.com/voxpupuli/puppet-prometheus/pull/90

alexjfisher: Since this isn't part of the modulesync or 'spec test fixes', does it belong in a separate PR (so that is makes it into the changelog)?

bastelfreak: that's why I didn't label it with modulesync. It will be in the changelog.

alexjfisher: To stop anyone else labelling it later, I've renamed the PR to make it a bit more obvious.

Issue #684, opened 2018-03-07
puppet package install error when running on SLES 12


paebersold:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.1.10
- Ruby: 2.1.9p490
- Distribution: SLES 12sp2
- Module version: v8.1.0

## How to reproduce (e.g Puppet code you use)

Using the default suse values from params.pp

## What are you seeing

Error: Failed to apply catalog: Parameter name failed on Package[rabbitmq-server]: Name must be a String not Array at /etc/puppetlabs/code/environments/modules/rabbitmq/manifests/install.pp:9

## What behaviour did you expect instead

package to be installed

## Output log

as above

## Any additional information you'd like to impart

The params.pp configuration for Suse has the packages defined as an array, not a string (https://github.com/voxpupuli/puppet-rabbitmq/pull/656).  More recent (since 3.4 is appears - see https://groups.google.com/forum/#!topic/puppet-users/gxwDg8XUVqw) versions of puppet no longer allow this behaviour.  

My workaround it to set the package list via hiera, ie
rabbitmq::package_name: 'rabbitmq-server'

Note that I am using the openSUSE rabbitmq-server package from https://www.rabbitmq.com/install-rpm.html.   There is no '-plugins' sub-package.

wyardley:
This was added fairly recently, in #656
Unfortunately, we don't yet have working acceptance tests for SLES / openSUSE, though there was some discussion about it there. I did a quick pass at trying to get it to work a while back, but wasn't able to; the "quirky" versioning also didn't help.

FWIW, the change was added at a point where the module already didn't support Puppet 3.x.  I can reproduce this error in the CentOS Vagrant box used for acceptance testing if I manually edit params to have an array for the package name. 

I'm not sure if there's any reason something like this wouldn't work:
```
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -6,9 +6,8 @@ class rabbitmq::install {
   $package_name     = $rabbitmq::package_name
   $rabbitmq_group   = $rabbitmq::rabbitmq_group
 
-  package { 'rabbitmq-server':
+  package { $package_name:
     ensure => $package_ensure,
-    name   => $package_name,
     notify => Class['rabbitmq::service'],
   }
 ```
From a quick test, that change does break some of the spec tests for Archlinux (because it has a different package name), so fixing this might involve some tweaks with that.

wyardley:
[...]

PR #686, opened 2018-03-07
Install package via title, not name (#684)

wyardley:
This fixes an issue with the package resource accepting an array for `$title`, but not for the name parameter. I'm not sure why it was originally coded this way, but it seems to work with one more minor adjustment to the Arch tests (now that #685 is merged).

It adds support in metadata.json for SLES 11 (only) to test the additional package we expect there. I didn't add support for SLES 12 or openSUSE because I'm not 100% sure about the versioning, and >=12 has systemd, so there are some additional adjustments needed there that should probably be done by someone who can do more testing and who knows the versioning situation (for example, we may not want to just do systemd based on >=12 because of the weird numbering w/ openSUSE).

Additionally, it slightly simplifies the resource requirement for the package(s) and moves them to `init.pp`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -298,9 +298,11 @@ class rabbitmq(
     case $facts['os']['family'] {
       'RedHat': {
         contain rabbitmq::repo::rhel
+        Class['rabbitmq::repo::rhel'] -> Class['rabbitmq::install']
       }
       'Debian': {
         contain rabbitmq::repo::apt
+        Class['rabbitmq::repo::apt'] -> Class['rabbitmq::install']
       }
       default: {
       }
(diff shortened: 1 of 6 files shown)


bastelfreak:
good catch @alexjfisher 

----

@wyardley I think the initial idea was that testing is easier. The unit test only matches for the resource name, but ignores the actual different package name

wyardley:
@alexjfisher Updated; not sure if https://github.com/voxpupuli/puppet-rabbitmq/blob/master/manifests/repo/rhel.pp#L8 / https://github.com/voxpupuli/puppet-rabbitmq/blob/master/manifests/repo/apt.pp#L19 can be left alone (or maybe just handle the ordering in a cleaner way?)
I'd rather make the minimum of structural changes here tho.

alexjfisher:
@wyardley i think it might be neater and easier to understand as eg
`Class[‘rabbitmq::repo::rhel’] -> Class[‘rabbitmq::install’]`
I’d probably move it to init.pp too.

PR #687, opened 2018-03-13
Allow dash as valid character for regex

crazymind1337:
* adds character - to regex for fact rabbitmq_nodename
* adds unit test for changed regex

### DIFF
--- a/spec/unit/facter/util/fact_rabbitmq_nodename_spec.rb
+++ b/spec/unit/facter/util/fact_rabbitmq_nodename_spec.rb
@@ -28,6 +28,16 @@ describe Facter::Util::Fact do
       end
     end
 
+    context 'with dashes in nodename/hostname' do
+      before do
+        allow(Facter::Util::Resolution).to receive(:which).with('rabbitmqctl') { true }
+        allow(Facter::Core::Execution).to receive(:execute).with('rabbitmqctl status 2>&1') { 'Status of node monty-python@rabbit-1 ...' }
+      end
+      it do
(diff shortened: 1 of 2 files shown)


spec/unit/facter/util/fact_rabbitmq_nodename_spec.rb
```diff
@@ -28,6 +28,16 @@
       end
     end
 
+    context 'with dash in username' do
```
wyardley: this should be 'nodename', no?

wyardley:
If I'm understanding correctly, #470 fixed the hostname regex itself, so this is just fixing the nodename fact to match, right?

bastelfreak:
Hi @crazymind1337, thanks for this PR. Can you take a look at the used email address in your commit? It isn't associated with your github account.

spec/unit/facter/util/fact_rabbitmq_nodename_spec.rb
crazymind1337: Yes, thanks for the hint. Will be done soon.

crazymind1337:
Oh :) I have worked on that with my company account but created the pull request with my private account :) I gonna fix it soon.

crazymind1337:
@wyardley, there is only one commit. The message just has more than one line. Do you want to have it also just to be a one-liner?

wyardley:
@crazymind1337 Yes, sorry, I just meant to squash if you made a new commit (which looks like you did). Looks good to me, so let's just see if the author looks right to @bastelfreak now.

wyardley:
Merged, thanks for the contribution.

Issue #689, opened 2018-03-20
rabbitmq_user_permissions timing out


cdenneen:
- Puppet: 4.10.1
- Ruby: 2.1.9 (puppet embedded)
- Distribution: Centos7
- Module version: 7.1.0, 8.1.0
- Rabbitmq: rabbitmq-server-3.3.5-34.el7.noarch

## How to reproduce (e.g Puppet code you use)
```
class { '::rabbitmq':
  environment_variables => {
    'node_ip_address' => '127.0.0.1',
  },
}

# Fixing a broken dependency
Package['rabbitmq-server'] -> Service['rabbitmq-server']

  rabbitmq_user { 'sensu':
    password => 'readonly',
    require  => Service['rabbitmq-server'],
  }
  -> rabbitmq_vhost { '/sensu':
    ensure  => present,
  }
  -> rabbitmq_user_permissions { 'sensu@/sensu':
    configure_permission => '.*',
    read_permission      => '.*',
    write_permission     => '.*',
  }
```
## What are you seeing
Have 2 prompts open to server running apply:

1. puppet apply
```
Debug: Prefetching rabbitmqctl resources for rabbitmq_user
Debug: Executing: '/usr/sbin/rabbitmqctl -q list_users'
Debug: Command succeeded
Notice: /Stage[main]/Profile::Sensu_server_enterprise/Rabbitmq_user[sensu]/ensure: current_value absent, should be present (noop)
Debug: /Stage[main]/Profile::Sensu_server_enterprise/Rabbitmq_user[sensu]: The container Class[Profile::Sensu_server_enterprise] will propagate my refresh event
Debug: Executing: '/usr/sbin/rabbitmqctl -q list_vhosts'
Debug: Command succeeded
Notice: /Stage[main]/Profile::Sensu_server_enterprise/Rabbitmq_vhost[/sensu]/ensure: current_value absent, should be present (noop)
Debug: /Stage[main]/Profile::Sensu_server_enterprise/Rabbitmq_vhost[/sensu]: The container Class[Profile::Sensu_server_enterprise] will propagate my refresh event
Debug: Executing: '/usr/sbin/rabbitmqctl -q list_user_permissions sensu'
Debug: Command failed, retrying
Debug: Executing: '/usr/sbin/rabbitmqctl -q list_user_permissions sensu'
Debug: Command failed, retrying
Debug: Executing: '/usr/sbin/rabbitmqctl -q list_user_permissions sensu'
[...]

PR #690, opened 2018-03-24
Explicitly set LC_ALL to C (#689)

wyardley:
This attempts to work around issues experienced related to LC_ALL not being set in a fairly conservative way, by setting it explicitly in the `$environment_variables` hash.

Not sure if we should do a merge with user-passed in ones, or if there is a better way to format this, but seems like it could help avoid issues? I will see if this runs against Ubuntu 16 in acceptance tests, but would be great if someone could test that this works properly and doesn't cause unintended consequences.

I guess that the main way it could cause issues would be if someone had a resource with 8 bit characters in the name, and didn't explicitly set another env? We could default to UTF-8, but C seems more conservative.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -135,7 +135,7 @@ class rabbitmq::params {
   $ldap_config_variables               = {}
   $wipe_db_on_cookie_change            = false
   $cluster_partition_handling          = 'ignore'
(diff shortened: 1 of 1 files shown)


wyardley:
@fatmcgav: you have any thoughts on this?

PR #695, opened 2018-04-09
Release 8.2.2

wyardley:
Release better bugfix for error with some newer versions of RabbitMQ when locale is not UTF-8.


wyardley:
@bastelfreak @alexjfisher Is the change in the way the Changelog was generated an issue?

PR #699, opened 2018-04-19
Add loopback_users parameter (adds ability to allow guest user to login remotely)

### LINKED ISSUE #698: Allow the remote connections with the guest user

- Puppet: 5.X
- Ruby: 2.4.X
- Distribution: Debian Stretch
- Module version: 8.2.2

The current version of the module doesn't allow to configure the exceptional use case to allow remote connections with the guest user

## What are you seeing
The configuration is not complete

A template to cover this weird use case

## Output log
...

jjuarez:
The aim of this pull request is to give the possibility to allow remote connections with the guest user, of course this is not the default configuration, because it implies some security risks 

Fixes #698

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -279,7 +279,7 @@ class rabbitmq(
   Boolean $ssl_erl_dist                                            = $rabbitmq::params::ssl_erl_dist,
   Optional[String] $rabbitmqadmin_package                          = $rabbitmq::params::rabbitmqadmin_package,
   Array $archive_options                                           = $rabbitmq::params::archive_options,
-  Optional[Array] $loopback_users                                  = $rabbitmq::params::loopback_users,
+  Array $loopback_users                                            = $rabbitmq::params::loopback_users,
 ) inherits rabbitmq::params {
 
   if $ssl_only and ! $ssl {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -186,7 +186,7 @@
 # @param rabbitmq_home OS dependent. default defined in param.pp. The home directory of the rabbitmq deamon.
 # @param $rabbitmqadmin_package OS dependent. default defined in param.pp. If undef: install rabbitmqadmin via archive, otherwise via package
 # @param $archive_options. default defined in param.pp.  Extra options to Archive resource to download rabbitmqadmin file
-# @param $allow_remote_connections. default defined in param.pp. This option allow remote connections using the guest user
+# @param $allow_remote_guest_connections. default defined in param.pp. This option allow remote connections using the guest user
 class rabbitmq(
   Boolean $admin_enable                                            = $rabbitmq::params::admin_enable,
   Enum['ram', 'disk', 'disc'] $cluster_node_type                   = $rabbitmq::params::cluster_node_type,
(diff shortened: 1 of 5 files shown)


jjuarez:
I'm not sure, the semantic of the parameter is not the best definitely, because the aim of this attribute is to allow the remote connections for a group of users, but by default with this configuration:
```
{rabbit, [{loopback_users: []}]}
```
you're allowing the connection to another bind (not only the loopback interface) to the guest user, in my opinion this seems a bit extrange

jjuarez:
All right them, now the new module variable is called *allow_remote_guest_connection* as you suggested @wyardley I think it would be more appropriate

wyardley:
I guess another way to do it would be to make a parameter `loopback_users` with the default value, and let an empty array be passed in to achieve the same result. I'm not sure which design is better.

I'm going to request feedback from @bastelfreak

wyardley:
Thanks @bastelfreak 
@jjuarez - does that make sense to you? I'm really sorry I didn't think of it in that way earlier :/
Either way, thanks for submitting this and for taking the time to write tests.

jjuarez:
Yes, of course, It make sense for me, the problem with this approach is the semantic of this configuration parameter, because I'm not sure what would be the default value for this list, I guest it could be something like that:
```
{loopback_users: ['guest']}
```
but I'm not sure and I didn't find information about this. So let me make a try with this new approach

manifests/init.pp
```diff
@@ -278,6 +279,7 @@
   Boolean $ssl_erl_dist                                            = $rabbitmq::params::ssl_erl_dist,
   Optional[String] $rabbitmqadmin_package                          = $rabbitmq::params::rabbitmqadmin_package,
   Array $archive_options                                           = $rabbitmq::params::archive_options,
+  Optional[Array] $loopback_users                                  = $rabbitmq::params::loopback_users,
```
wyardley: This is a strange quirk of how Puppet works, but a param can't be Optional if it has a default. Well, technically, it can, but it can only be overridden with `undef` using hiera, so it's safer to not make it optional, and pass in an empty array (many people's preferred pattern anyway) to set it to an empty list when necessary.

wyardley:
ps - https://www.rabbitmq.com/configure.html#configuration-file does confirm that 'guest' is the default setting for that parameter, so I think we're good there.

manifests/init.pp
jjuarez: Yep, you're right, I'm working with puppet 4 and the automatic hiera bind then I forgot the "old" params pattern; I think an empty list as default value for this parameter is not the best choice to achieve the behaviour that you could expect, so I could make not optional this parameter a fix the default value to the list with the guest user, let's see

wyardley: Right, I think the right fix is to make the parameter not optional, and if users want to allow the guest user to login remotely as well, they can pass in an empty array. You could explicitly mention that in the param docs if you want to make it more clear.

wyardley: But the default value should be `['guest']` and then just `Array $loopback_users`. I think we're saying the same thing 😃

jjuarez: Yes, we're in the same page

wyardley:
@jjuarez Merged, and thanks for the contribution!

jjuarez:
It has been a pleasure to collaborate, thanks @wyardley

wyardley:
@jjuarez also interesting workaround noted here: https://github.com/voxpupuli/puppet-rabbitmq/pull/480#issuecomment-384705832

Obviously, not needed now, but might be handy if you need to tweak other config variables; I should have thought of that.

jjuarez:
Thanks @wyardley it looks very useful to help me  “temporary” with my configuration issue

PR #705, opened 2018-06-22
Ensure version fact does not throw an error for invalid match

### LINKED ISSUE #704: rabbitmq_version fact fails on Ubuntu 18.04

- Puppet: 5.5.2
- Ruby: 2.4.4p296 (from Puppet)
- Distribution: Ubuntu 18.04, Package rabbitmq-server 3.6.10-1
- Module version: 8.2.2

Add the rabbitmq class on a Ubuntu 18.04 system using the Ubuntu repo and run puppet.
```
class { 'rabbitmq':
  service_manage    => false,
  port              => '5672',
  delete_guest_user => true,
  repos_ensure      => false,
}
```

## What are you seeing
```
Error: Facter: error while resolving custom fact "rabbitmq_version": undefined method `[]' for nil:NilClass
```
This happens because `rabbitmqadmin --version` returns `rabbitmqadmin %%VSN%%` on Ubuntu 18.04. This is a bug in the Ubuntu package (already reported).

Even though this is caused by a bug in the Ubuntu package, this should not cause the fact to error out. I would rather just have the fact be missing.

CTrox:
If the `rabbitmq_version` string returned from `rabbitmqadmin --version` contains invalid characters it throws an error because the array is nil. This will make a empty array out of it so it won't throw an error in that case.

Fixes #704

### DIFF
--- a/lib/facter/rabbitmq_version.rb
+++ b/lib/facter/rabbitmq_version.rb
@@ -2,7 +2,7 @@ Facter.add(:rabbitmq_version) do
   setcode do
     if Facter::Util::Resolution.which('rabbitmqadmin')
       rabbitmq_version = Facter::Core::Execution.execute('rabbitmqadmin --version 2>&1')
-      %r{^rabbitmqadmin ([\w\.]+)}.match(rabbitmq_version)[1]
+      %r{^rabbitmqadmin ([\w\.]+)}.match(rabbitmq_version).to_a[1]
     end
   end
 end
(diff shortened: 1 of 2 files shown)


wyardley:
This seems sensible to me (though keep in mind that the module does use the version fact to determine some things about how it operates, so it may not operate correctly if the version is nil).

Can you add a spec test test for a nil value with an invalid version value?
https://github.com/voxpupuli/puppet-rabbitmq/blob/master/spec/unit/facter/util/fact_rabbitmq_version_spec.rb

wyardley:
@dhollinger from my perspective, good to ship it

PR #709, opened 2018-08-10
Updated comment symbol on inetrc.erb

covidium:
Linting validation fails on this file.

### DIFF
--- a/templates/inetrc.erb
+++ b/templates/inetrc.erb
@@ -1,5 +1,5 @@
-% This file managed by Puppet
-% Template Path: <%= @module_name %>/templates/inetrc
+%% This file managed by Puppet
+%% Template Path: <%= @module_name %>/templates/inetrc
 <%- if @ipv6 -%>
 {inet6, true}.
 <%- end -%>


wyardley:
Can you comment on the reason for this a bit more? And which linting is failing?

Seems like '%% ' should be the correct comment format, no?
http://erlang.org/doc/apps/erts/inet_cfg.html

If this is a recent change, is it backwards compatible?

That said, I think the Travis acceptance test failures would seem to be unrelated to this change, since they're erroring before they even start.

bastelfreak:
Hi @covidium, thanks for the PR. Are you interested in turning this into an epp template? https://puppet.com/docs/puppet/5.5/lang_template_epp.html
 This is the new templating language from Puppet that is more solid than erb. Please rebase after https://github.com/voxpupuli/puppet-rabbitmq/pull/710 got merged, that will fix the broken beaker tests.

wyardley:
@bastelfreak regardless of the template time, I'm still not sure this is a good change to make?

bastelfreak:
@wyardley I've no knowledge about the format, I can't say anything about it. But we should be able to check if it works with an acceptance test?

covidium:
@wyardley you're totally right the comments should use double percent symbols to mark the comments and not a single one. My validation test failed on this erb template as the 2nd line has also embedded ruby code in the comment.

The command is /bin/erb -x -T '-' $file | ruby -c || exit 1;

ERB TEMPLATE - modules/rabbitmq/templates/inetrc.erb
-:3: syntax error, unexpected '<'
 Template Path: <%= @module_name %>/templates/inetrc
                 ^
-:3: unterminated string meets end of file

Once I've doubled the symbol, validation of the erb file was OK.

ERB TEMPLATE - modules/rabbitmq/templates/inetrc.erb
Syntax OK

@bastelfreak I would like to help with the epp template, but at this moment I've zero experience with it

wyardley:
@covidium Looks good to me, thanks for the contribution.

I checked out the PR and double-checked that things look good and that the module path is still rendered correctly.

Since the commits are signed, I don't want to squash-merge -- would you mind squashing down to a single commit / force-pushing in your feature branch?

covidium:
@wyardley done

wyardley:
@covidium Merged, and thanks for the contribution!

PR #714, opened 2018-08-17
Fixing puppet apt module requirement to < 5.1.0

meltingrobot:

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -10,7 +10,7 @@ install_module
 install_module_dependencies
 
 # Install aditional modules for soft deps
-install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 5.0.0') if fact('os.family') == 'Debian'
+install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 5.1.0') if fact('os.family') == 'Debian'
 install_module_from_forge('garethr-erlang', '>= 0.3.0 < 1.0.0') if fact('os.family') == 'RedHat'
 
 RSpec.configure do |c|


spec/spec_helper_acceptance.rb
```diff
@@ -10,7 +10,7 @@
 install_module_dependencies
 
 # Install aditional modules for soft deps
-install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 5.0.0') if fact('os.family') == 'Debian'
+install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 5.1.0') if fact('os.family') == 'Debian'
```
bastelfreak: can you please update this to `< 6.0.0`?  We always try to support the whole major release range.

wyardley:
Puppet librarian shouldn't be looking at the spec_helper for acceptance tests, should it? I don't see an objection to merging this, but I don't see how it can help either. @bastelfreak?

I could see it installing deps from metadata but from spec_helper_acceptance?

wyardley:
@meltingrobot can you squash that commit, if you know how?

wyardley:
I believe that file should be only looking at `metadata.json` and the like, but it will pull in sub dependencies and their dependencies. So it’s possible you have a module that depends on a module that depends on apt (btw, apt is supposed to be a soft dep, and shouldn’t be listed in metadata).

wyardley:
@meltingrobot Awesome, glad we figured that all out. BTW, I think Puppet's recommendation is that things like apt (which affect a specific OS) should be soft deps. So you might propose removing it from `metadata.json` (and adding docs notes) rather than upping the version requirement.

PR #716, opened 2018-08-26
allow puppetlabs/stdlib 5.x, puppet/archive 3.x and puppetlabs/apt 6.x

bastelfreak:


wyardley:
Looks ok to me, but the acceptance test failures do seem related

juniorsysadmin:
Acceptance tests depend on https://github.com/garethr/garethr-erlang which is stale

wyardley:
@juniorsysadmin Is it erlang? I agree that the erlang dep thing is a nightmare and that garethr-erlang is pretty unuseful at this point. That said, isn't the actual error here related to puppet-archive version spec, or is Travis just not printing the useful part?

```
  Last 10 lines of output were:
  	Notice: Preparing to install into /etc/puppetlabs/code/environments/production/modules ...
  	Notice: Downloading from https://forgeapi.puppet.com ...
  	Error: Could not install module 'puppet-archive' (???)
  	  No version of 'puppet-archive' can satisfy all dependencies
  	    Use `puppet module install --ignore-dependencies` to install only this module
```

bastelfreak:
@wyardley old version of archive pulls in an old version of stdlib. I fixed it.

bastelfreak:
gnaaaaaar. It will fail because the erlang module requires stdlib < 5. https://github.com/garethr/garethr-erlang/blob/master/metadata.json#L51

Can we get rid of the dependency to garethr-erlang?

wyardley:
@bastelfreak I think the module suggests multiple ways of satisfying the erlang dep. I think there are some existing issues with garethr/erlang, but I'm not sure we can get new enough Erlang version on all platforms without it or a replacement. Some more discussion in #717

bastelfreak:
Interesting that travis turned green. I restarted the tests.

bastelfreak:
@wyardley I am going to merge this and do a new release. The erlang module on the forge has no version boundaries for stdlib.

PR #720, opened 2018-09-02
Enable acceptance tests; Add debian 9 support

bastelfreak:


bastelfreak:
The tests do not work on ubuntu 18.04, this needs to be investigated. I removed them for now.

PR #736, opened 2018-10-14
Add auto cluster configuration support

vchepkov:
- added clustername fact and spec tests for it
- added rabbitmq_cluster type and provider with tests

### DIFF
--- a/examples/cluster/join_cluster_and_change_name.pp
+++ b/examples/cluster/join_cluster_and_change_name.pp
@@ -1,0 +1,4 @@
+rabbitmq_cluster { 'test_cluster':
+  init_node => 'host1',
+}
+# This will join host2 to host1s cluster and set the cluster name to test_cluster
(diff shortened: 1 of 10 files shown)


bastelfreak:
Hey @vchepkov, thanks for picking this up. Can you take a look at the failing travis jobs?

vchepkov:
I afraid tests are not my forte. I will see if I can copy&paste something, that's an extent of my expertise here :)

KlavsKlavsen:
Really hoping this gets in.. would make cluster setup soo much easier :)

KlavsKlavsen:
I just realized that if you give rabbitmq nodenames in FQDN format - it removes the domain part - and you cannot join - unless you give that name (ie. hostname part only). But in the acceptance test here - init_node is set from $facts['fqdn'] - which won't work..
This should probably strip domain name part of a node name - before trying to join atleast.

Fabian1976:
I ran in to this issue also because i used FQDN as cluster_node names
```
rabbitmq::cluster_nodes:
  - 'local-o-rmq001.example.com'
  - 'local-o-rmq002.example.com'
  - 'local-o-rmq003.example.com'
```

But switching them to hostname only:
```
rabbitmq::cluster_nodes:
  - 'local-o-rmq001'
  - 'local-o-rmq002'
  - 'local-o-rmq003'
```
(and make sure they would resolve in DNS, made the cluster work.

PR #739, opened 2018-11-07
Allow Array values in rabbitmq_parameter

vStone:
Array (lists) are allowed values to use with rabbitmqctl.

For example: src-uri when setting up a shovel.

### DIFF
--- a/spec/unit/puppet/type/rabbitmq_parameter_spec.rb
+++ b/spec/unit/puppet/type/rabbitmq_parameter_spec.rb
@@ -61,7 +61,7 @@ describe Puppet::Type.type(:rabbitmq_parameter) do
 
   it 'does not accept an array for definition' do
     expect do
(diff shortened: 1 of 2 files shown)


spec/unit/puppet/type/rabbitmq_parameter_spec.rb
```diff
@@ -61,7 +61,7 @@
 
   it 'does not accept an array for definition' do
```
wyardley: At a minimum, this part should probably be changed, and / or be tested properly, both here and in the provider?

I would think the right test would now be that a correctly formatted array _is_ accepted, and then maybe a corresponding test in
https://github.com/voxpupuli/puppet-rabbitmq/blob/master/spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_spec.rb to validate that when the value is an array rather than a hash, the correct thing happens?

PR #743, opened 2018-12-03
Update APT GPG Key

### LINKED ISSUE #742: Package cloud APT GPG key changed?

- Distribution: Ubuntu/Debian

Attempt to install rabbitmq-server using apt method 

## What are you seeing

Failure to validate APT gpg key as it's not available

Installation of rabbitmq-server

Our internal overnight build failed last night.  After investigation, this is due to the fact that the GPG key listed in https://github.com/voxpupuli/puppet-rabbitmq/blob/master/manifests/repo/apt.pp is no longer the relevant key for the archive.

Having investigated, it seems that the key needed is `8C695B0219AFDEB04A058ED8F4E789204D206F89`

I've found this via

```
 % wget https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey -qO - | gpg              
gpg: WARNING: no command supplied.  Trying to guess what you mean ...
pub   rsa4096 2018-10-08 [SCEA]
      8C695B0219AFDEB04A058ED8F4E789204D206F89
uid           https://packagecloud.io/rabbitmq/rabbitmq-server (https://packagecloud.io/docs#gpg_signing) <support@packagecloud.io>
sub   rsa4096 2018-10-08 [SEA]
```

I will make a new PR to update this, however, it seems that this should be checked to see if this is a security issue?  Changing the key - (and not having that key available in public keyservers) seems like a MAJOR security issue ?

Mezzle:
Fixes #742

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -534,8 +534,9 @@ Default value: $rabbitmq::params::package_ensure
 Data type: `Optional[String]`
 
 RPM package GPG key to import. Uses source method. Should be a URL for Debian/RedHat OS family, or a file name for
-RedHat OS family. Set to https://packagecloud.io/gpg.key by default. Note, that `key_content`, if specified, would override this
-parameter for Debian OS family.
+RedHat OS family. Set to https://www.rabbitmq.com/rabbitmq-release-signing-key.asc for RedHat OS Family and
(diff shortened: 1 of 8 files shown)


wyardley:
In the meantime, I'll restart those two beaker tests in case that helps.

Mezzle:
Changes apply to any versions on packagecloud

wyardley:
Did you update the docs manually? I think the html ones should be generated with strings, right?

PR #744, opened 2018-12-06
Update REFERENCE.md, remove docs dir from master

wyardley:
    
Update REFERENCE.md, as docs haven't been regenerated for a while.
We should either not push the HTML docs, or put them in gh-pages. This is still
under discussion (https://github.com/voxpupuli/modulesync_config/issues/397). 

I can also update gh_pages branch via the rake task if you'd like.


REFERENCE.md
```diff
@@ -535,8 +543,8 @@ Data type: `Optional[String]`
 
 RPM package GPG key to import. Uses source method. Should be a URL for Debian/RedHat OS family, or a file name for
 RedHat OS family. Set to https://www.rabbitmq.com/rabbitmq-release-signing-key.asc for RedHat OS Family and
-https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey for Debian OS Family by default. Note, that `key_content`, if
-specified, would override this parameter for Debian OS family.
+https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey for Debian OS Family by default.. Note, that `key_content`, if specified, would
```
ekohl: The `..` looks odd. Looks like it's also present in `init.pp`.

REFERENCE.md
```diff
@@ -589,6 +597,10 @@ The name of the service to manage.
 
 Default value: $rabbitmq::params::service_name
 
+##### `$service_restart.`
+
+Default defined in param.pp. Whether to resetart the service on config change.
```
ekohl: resetart -> restart

wyardley:
@ekohl @bastelfreak Fixed some other issues / typos with the source docs.

I can switch it to be a single commit with the changes to source and then the generated docs in another if that's better.

REFERENCE.md
```diff
@@ -35,7 +35,7 @@
 
 rabbitmq
 
- @param $loopback_users. default defined in param.pp. This option configures a list of users to allow access via the loopback interfaces
+ @param loopback_users This option configures a list of users to allow access via the loopback interfaces
```
ekohl: Looks like it still doesn't recognize it as a parameter. Not sure why.

REFERENCE.md
```diff
@@ -35,7 +35,7 @@
 
 rabbitmq
```
ekohl: This comes from the first line. I think it can be removed

REFERENCE.md
wyardley: I see that. Not sure why, but still get the error
```
[warn]: Missing @param tag for parameter 'loopback_users' near manifests/init.pp:192.
```

ekohl: You could try adding a line with just a `#` after is and before the class definition.

wyardley: I tried that, didn't seem to make a difference.

REFERENCE.md
wyardley: removed

REFERENCE.md
wyardley: An empty line with no hash seems to work... I _think_ that's right? The diffs are hard to review now.

REFERENCE.md
```diff
@@ -5,7 +5,7 @@
 
 **Classes**
 
-* [`rabbitmq`](#rabbitmq): A module to manage RabbitMQ
+* [`rabbitmq`](#rabbitmq):
```
wyardley: @ekohl not sure why the `@summary` disappears now? So maybe that leading `# rabbitmq` is necessary?

ekohl:
I *think* `@param name description` is invalid and it needs to be:
```
@param name
  description
```

wyardley:
Argh, now the examples are messed up....

ekohl:
See my changes in https://github.com/ekohl/puppet-rabbitmq/commit/57b0b1740b108b58c508030dc28f396a2804e1d3

I've replaced the spaces and then it does pick it up. The reason is that there's a weird unicode character after the `#` on that line. Using `cat -vet` will show this to you:

```
$ cat -vet manifests/init.pp | grep loopback_users | head -n 1
#M-BM- @param $loopback_users. default defined in param.pp. This option configures a list of users to allow access via the loopback interfaces$
```

wyardley:
Thanks. Do you want me to try and cherry pick that off, or do you want to submit it?

wyardley:
> $ cat -vet manifests/init.pp
On OS X, this renders normally for me too. `od -c` shows:
```
0026020    a   d   m   i   n       f   i   l   e  \n   #      **   @   p
0026040    a   r   a   m       l   o   o   p   b   a   c   k   _   u   s
0026060    e   r   s       T   h   i   s       o   p   t   i   o   n    
```

Pushed up a version without the formatting changes and without some of the other changes, but happy to close this in favor of your version if you want.

ekohl:
> Thanks. Do you want me to try and cherry pick that off, or do you want to submit it?

Feel free to pick it into your branch and improve from there.

wyardley:
Also created https://tickets.puppetlabs.com/browse/PDOC-271

REFERENCE.md
```diff
@@ -5,16 +5,21 @@
...
 * [`rabbitmq::server`](#rabbitmqserver): Class: rabbitmq::server  This module manages the installation and config of the rabbitmq server   it has only been tested on certain version 
-* [`rabbitmq::service`](#rabbitmqservice): Class: rabbitmq::service    This class manages the rabbitmq server service itself.  Parameters:  Actions:  Requires:  Sample Usage:
+
+_Private Classes_
+
+* `rabbitmq::config`: Sets all the configuration values for RabbitMQ and creates the directories for config and ssl.
+* `rabbitmq::install`: Ensures that rabbitmq-server exists
+* `rabbitmq::install::rabbitmqadmin`: Install rabbitmq admin
+* `rabbitmq::params`: OS Specific parameters and other settings
+* `rabbitmq::repo::apt`: requires   puppetlabs-apt   puppetlabs-stdlib
```
ekohl: Probably could use a better summary

PR #745, opened 2018-12-07
Updates to rabbitmq::server docs

wyardley:
Per discussion in #744, document `rabbitmq::server`, just because it's so messy.

That said, I'm thinking we might want to deprecate / remove this compatibility interface?


ekohl:
We could also nuke the class. It's been a compat wrapper since 2014.

wyardley:
@ekohl Yeah, my thoughts too. But since I fixed the docs, let’s go ahead and release w/ this, and kill it on the next major (if we remember)?

PR #746, opened 2018-12-08
Release 8.4.1

wyardley:


ekohl:
Probably because we need to do a puppet-blacksmith release that includes https://github.com/voxpupuli/puppet-blacksmith/commit/64bc5554c33035cf690410f3223187074571585d

wyardley:
> Probably because we need to do a puppet-blacksmith release that includes...

No, I just bumped it wrong. I'm updating and regenerating changelog.

PR #747, opened 2018-12-08
Remove duplicated / outdated docs in README

wyardley:
Since we are generating REFERENCE.md, point users to here for docs.

We can regenerate changelog after this is merged for #746


README.md
```diff
@@ -46,310 +46,12 @@ prior to 2.7.1.
 
 All options and configuration can be done through interacting with the parameters
 on the main rabbitmq class.
-These are now documented via [Puppet Strings](https://github.com/puppetlabs/puppet-strings)
```
ekohl: This could still be a good hint.

wyardley: agreed. I was on the fence about this; put it back in.

PR #755, opened 2018-12-25
modulesync 2.4.0

bastelfreak:
modulesync 2.4.1


.msync.yml
```diff
@@ -1 +1 @@
-modulesync_config_version: '2.3.1'
+modulesync_config_version: '2.4.0'
```
wyardley: should this be 2.4.1?

bastelfreak: ups. this is 2.4.0

PR #758, opened 2019-01-15
Officially support Ubuntu 18.04

### LINKED ISSUE #757: Officially support Ubuntu 18.04

- Puppet: All
- Ruby: All
- Distribution: Ubuntu 18.04
- Module version: Current master

Take a look into metadata.json

## What are you seeing
Ubuntu 18.04 isn't supported

Ubuntu 18.04 is supported

baurmatt:
This PR adds official (metadata.json + Tests) for Ubuntu 18.04.

Fixes #757.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,21 @@ matrix:
     sudo: required
   - rvm: 2.5.1
     bundler_args: --without development release
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_IS_PE=no BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_debug=true BEAKER_setfile=ubuntu1804-64 BEAKER_HYPERVISOR=docker CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.5.1
+    bundler_args: --without development release
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_IS_PE=no BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_debug=true BEAKER_setfile=ubuntu1804-64 BEAKER_HYPERVISOR=docker CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.5.1
+    bundler_args: --without development release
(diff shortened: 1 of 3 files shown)


baurmatt:
Hmm this is strange:

```
    Port "15672"

ubuntu1804-64-1 17:09:00$ /bin/sh -c ss\ -tnl4\ \|\ grep\ --\ \\\ \\\\\\\*:15672\\\

ubuntu1804-64-1 executed in 0.01 seconds
```

```
root@ubuntu1804-64-1:/# netstat -nlp | grep 15672
tcp        0      0 0.0.0.0:15672           0.0.0.0:*               LISTEN      2410/beam.smp
```

The problem seems to be that its trying to grep for `*:15672` due to this line in specinfra https://github.com/mizzy/specinfra/blob/7fcb239eb5b4dca705663ca342c405c41a5ffc0c/lib/specinfra/command/module/ss.rb#L17

I guess the output off ss changed:
```
root@ubuntu1804-64-1:/# ss -tnl4 | grep 15672
LISTEN   0         128                 0.0.0.0:15672            0.0.0.0:*
```

```
root@ubuntu1804-64-1:/# which ss
/bin/ss
root@ubuntu1804-64-1:/# dpkg -S /bin/ss
iproute2: /bin/ss
root@ubuntu1804-64-1:/# dpkg -l | grep iproute2
ii  iproute2                    4.15.0-2ubuntu1                   amd64        networking and traffic control tools
```

baurmatt:
This is most certainly the reason for this https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/commit/?id=1267c0b924de0297510efd48f8f849af774be369

wyardley:
Good find. It would be better if we can avoid just marking the tests of the service listening as pending — want to try and get this fixed upstream?

baurmatt:
@wyardley Just opened a issue for that ;) Take a look, perhaps you have a good idea of how we can fix this :)

baurmatt:
Oh really? Since when is that? Worked for me a couple of weeks ago pretty well.

Travis should update the specinfra gem as we do not pin it directly.

wyardley:
@baurmatt Travis also seems to be having some issues right now, not running on any of the new PRs :/

You're right about specinfra; forgot that `Gemfile.lock` wasn't committed when I looked locally.

Also, seems like the build is passing.

wyardley:
@baurmatt: can you open a separate PR to add it to docker sets in https://github.com/voxpupuli/puppet-rabbitmq/blob/master/.sync.yml ?

baurmatt:
I keep forgetting about modulesync :( Openend up #767 to make it persistent.

PR #765, opened 2019-01-16
Allow offline and online mode for plugins

ahmet2mir:
Hello Pupuli!

In RabbitMQ we could enable a plugin offline or online.
Some plugins should be enabled offline.

See https://www.rabbitmq.com/plugins.html#ways-to-enable-plugins
man rabbitmq-plugins

### DIFF
--- a/lib/puppet/type/rabbitmq_plugin.rb
+++ b/lib/puppet/type/rabbitmq_plugin.rb
@@ -9,6 +9,12 @@ manages rabbitmq plugins
  rabbitmq_plugin {'rabbitmq_stomp':
    ensure => present,
  }
+
+@example Ensure a rabbitmq_plugin offline resource (with RabbitMQ version >=3.4.0)
+ rabbitmq_plugin {'rabbitmq_stomp':
+   ensure => present,
+   mode   => 'offline',
+ }
 DESC
 
   ensurable do
@@ -24,6 +30,12 @@ DESC
   newparam(:name, namevar: true) do
     desc 'The name of the plugin to enable'
     newvalues(%r{^\S+$})
+  end
+
+  newparam(:mode) do
+    desc 'Define how the plugin should be enabled regarding node status.'
+    newvalues(:online, :offline, :best)
+    defaultto(:best)
   end
 
   newparam(:umask) do
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -52,6 +52,13 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
     rabbitmqctl("list_#{resource}", *list_opts, *opts)
   end
 
+  def self.rabbitmq_running
+    rabbitmqctl('-q', 'status')
+    return true
+  rescue Puppet::ExecutionFailure, Timeout::Error
+    return false
+  end
+
   # Retry the given code block 'count' retries or until the
   # command succeeds. Use 'step' delay between retries.
   # Limit each query time by 'timeout'.
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -1,3 +1,5 @@
+require 'puppet/util/package'
+
 require File.expand_path(File.join(File.dirname(__FILE__), '..', 'rabbitmq_cli'))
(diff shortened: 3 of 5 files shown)


lib/puppet/type/rabbitmq_plugin.rb
```diff
@@ -26,6 +32,15 @@
     newvalues(%r{^\S+$})
   end
 
+  newparam(:mode) do
```
wyardley: `mode` seems a little potentially ambiguous with file modes, but also, I think an `offline` Boolean would probably be cleaner?

ahmet2mir: First I used a bool but why offline => true and not online => false
It's not like the flag durable which should be true or false.
Then I used a free param "options" where I just put "--offline".
Finally I found that specify offline or online explicitly is easier. 

If you think that it's more clean and simple for a user, why not :)

ahmet2mir:
Offline or online ? At least 2 years z years ago.
But I found a recent commit in rabbit  https://github.com/rabbitmq/rabbitmq-cli/commit/09466e342ae9196d64116e790fa69a63980d7c66

I was using Rabbit 3.5 and just migrated to 3.7.10.
So since 3.6 it seems that thoses flags had 3 values: offline, online or "best" if nothing specified. 

I will add "best" to other options and Boolean can't work here. The name used in rabbit is "mode".

wyardley:
Is it supported in 3.3.x, though? Otherwise you’ll need to wrap it in version specific logic.

wyardley:
Acceptance tests run against vendor version, which is 3.3.x still on some platforms. Eventually, it may make sense to do a breaking chance and move back to using the official RMQ packages by default (which would also involve handling Erlang deps differently), but so far, no one has created a PR to propose that change.

ahmet2mir:
Arf. Ok I will check it tomorrow (hello from France).

wyardley:
Thanks! Getting the acceptance tests running locally can be super helpful for testing (you can play around with different params / versions too).

I personally think it would be cleaner to decouple the param value and the CLI flags a bit more.

wyardley:
The rubocop and spec tests can also be run locally.

ahmet2mir:
@wyardley added tests, move to provider and check version.

Found from which version thoses flags exists
```bash
docker run --rm -it centos:7 bash
yum install -y https://www.rabbitmq.com/releases/erlang/erlang-19.0.4-1.el7.centos.x86_64.rpm
yum install -y https://www.rabbitmq.com/releases/rabbitmq-server/v3.3.5/rabbitmq-server-3.3.5-1.noarch.rpm
rabbitmq-plugins --help # ko
yum install -y https://www.rabbitmq.com/releases/rabbitmq-server/v3.4.0/rabbitmq-server-3.4.0-1.noarch.rpm
rabbitmq-plugins --help # ok
```

To get the version, rabbit should running. So I added a function to check it.
If it's not running, this option doesn't matter.

bastelfreak:
@wyardley are you fine with this as well?

wyardley:
Seems ok to me from a quick look, just wondering if it makes since to wait til after we cut the release and do modulesync?

ahmet2mir:
@wyardley squashed. Thanks for your time :)

wyardley:
@ahmet2mir Can you rebase again? We've skipped the puppet6-nightly stuff in Travis, at least for now.

ahmet2mir:
@wyardley  done.

wyardley:
Looks like you merged vs rebased?

ahmet2mir:
Arf, I "pull -r" (wrong reflex), fixed with a simple rebase.

wyardley:
That worked! These aren’t the nightly build failure issue, so can you take a look at the Travis beaker failures?

ahmet2mir:
It's something else.
I've got frozen string error. All other tests fails.

```
/home/travis/build/voxpupuli/puppet-rabbitmq/spec/spec_helper_acceptance.rb:6: warning: already initialized constant UNSUPPORTED_PLATFORMS
/home/travis/build/voxpupuli/puppet-rabbitmq/spec/spec_helper_acceptance.rb:6: warning: previous definition of UNSUPPORTED_PLATFORMS was here
An error occurred while loading ./spec/acceptance/clustering_spec.rb.
Failure/Error: run_puppet_install_helper
FrozenError:
  can't modify frozen String
```

wyardley:
@ahmet2mir I'm getting that error on the modulesync PR as well (even though last run on master went Ok), so I think this must be unrelated to changes here.

wyardley:
Thanks for your patience and for the contribution @ahmet2mir

PR #764, opened 2019-01-16
Update fallback path for CLI commands

wyardley:
#### Pull Request (PR) description
This should resolve an idempotency issue, at least on certain platforms, by using the version of `rabbitmqctl` and `rabbitmq-plugins` in `/usr/sbin` (the wrapper) vs. the one in `/usr/lib/rabbitmq/bin`. Open to suggestions of a cleaner fix for this

Possibly related to #748 as well

cc: @JayH5 

#### This Pull Request (PR) fixes the following issues
    Fixes #763 
    Fixes #748

PR #766, opened 2019-01-16
Override the PATH for providers to include /usr/lib/rabbitmq/bin

### LINKED ISSUE #763: rabbitmqctl -q status failes

- Puppet: 5.3.2
- Ruby: 
- Distribution: Centos 7
- Module version: no version, since the latest PR isn't released yet in a version

```puppet
  class { '::rabbitmq':
    package_ensure    => present,
    interface         => '192.168.1.150',
    port              => 5672,
    admin_enable      => true,
    management_port   => 15672,
    delete_guest_user => true,
  }
```

## What are you seeing

```
Notice: /Stage[main]/Rabbitmq::Service/Service[rabbitmq-server]/ensure: ensure changed 'stopped' to 'running'
Debug: /Stage[main]/Rabbitmq::Service/Service[rabbitmq-server]: The container Class[Rabbitmq::Service] will propagate my refresh event
Info: /Stage[main]/Rabbitmq::Service/Service[rabbitmq-server]: Unscheduling refresh on Service[rabbitmq-server]
Debug: Class[Rabbitmq::Service]: The container Stage[main] will propagate my refresh event
Debug: Class[Rabbitmq::Service]: The container Class[Rabbitmq] will propagate my refresh event
Debug: Prefetching rabbitmqctl resources for rabbitmq_user
Debug: Executing: '/usr/lib/rabbitmq/bin/rabbitmqctl -q status'
Debug: Command failed, retrying
```

A Puppet apply without errors.

## Output log

In /var/log/rabbitmq/log/crash.log I get these lines:

```
2019-01-16 07:28:40 =ERROR REPORT====
** Connection attempt from disallowed node 'rabbitmqcli-2401-rabbit@d7b1151hi4b7ae2' **
```

Did a little digging, it maybe that this is partly a RabbitMQ problem, when on the command-prompt and I run the command manually I get this:

```
[root@d7b1151hi4b7ae2 log]# /usr/lib/rabbitmq/bin/rabbitmqctl -q status
Error: unable to perform an operation on node 'rabbit@d7b1151hi4b7ae2'. Please see diagnostics information and suggestions below.

Most common reasons for this are:

 * Target node is unreachable (e.g. due to hostname resolution, TCP connection or firewall issues)
 * CLI tool fails to authenticate with the server (e.g. due to CLI tool's Erlang cookie not matching that of the server)
 * Target node is not running

In addition to the diagnostics info below:

 * See the CLI, clustering and networking guides on http://rabbitmq.com/documentation.html to learn more
 * Consult server logs on node rabbit@d7b1151hi4b7ae2

DIAGNOSTICS
===========

attempted to contact: [rabbit@d7b1151hi4b7ae2]

rabbit@d7b1151hi4b7ae2:
  * connected to epmd (port 4369) on d7b1151hi4b7ae2
  * epmd reports node 'rabbit' uses port 25672 for inter-node and CLI tool traffic
  * TCP connection succeeded but Erlang distribution failed

  * Authentication failed (rejected by the remote node), please check the Erlang cookie


Current node details:
 * node name: 'rabbitmqcli-2876-rabbit@d7b1151hi4b7ae2'
 * effective user's home directory: /root
 * Erlang cookie hash: 7ZzWTe+DWf70zSiDpO3ZYw==
```

But when I run the same commando from '/usr/sbin':

```
[root@d7b1151hi4b7ae2 ~]# /usr/sbin/rabbitmqctl -q status
[{pid,1555},
 {running_applications,
     [{rabbitmq_management,"RabbitMQ Management Console","3.7.10"},
      {rabbitmq_management_agent,"RabbitMQ Management Agent","3.7.10"},
      {rabbitmq_web_dispatch,"RabbitMQ Web Dispatcher","3.7.10"},
      {rabbit,"RabbitMQ","3.7.10"},
...
```

When running Puppet a second time after this error, it picks up the /usr/sbin version and works !

```
Debug: Executing: '/usr/bin/systemctl is-active rabbitmq-server'
Debug: Executing: '/usr/bin/systemctl is-enabled rabbitmq-server'
Debug: Prefetching rabbitmqctl resources for rabbitmq_user
Debug: Executing: '/usr/sbin/rabbitmqctl -q status'
Debug: Executing: '/usr/sbin/rabbitmqctl list_users -q --no-table-headers'
Debug: Command succeeded
Debug: /Stage[main]/Rabbitmq::Management/Rabbitmq_user[guest]: Nothing to manage: no ensure and the resource doesn't exist
```

This default comes from: './lib/puppet/provider/rabbitmq_cli.rb', changing it to '/usr/sbin'  solves the run, but don't know if this is wise for other OS then RedHat/CentOS?!?!

JayH5:

This attempts to rework the resolution of RabbitMQ CLI executables. Some discussion is in #764. This is not a _great_ solution because it modifies the environment (`PATH`) which reflects everywhere else in the process (and child processes). But I've dug through the Puppet code and there doesn't seem to be a way to adjust the `PATH` specifically for when Puppet is searching for an executable for a specific command, short of some meta-programming.

Fixes #763


wyardley:
Does this work in beaker?

Possibly slightly less elegant workaround (that does seem to work for me on CentOS 6 and 7 acceptance tests) in #764

lib/puppet/provider/rabbitmq_cli.rb
```diff
@@ -1,19 +1,36 @@
...
+  # specified by name (instead of absolute path), Puppet searches the PATH
+  # before each execution of the command. Append /usr/lib/rabbitmq/bin to the
+  # end of the PATH so that Puppet will look there last at the time a command is
+  # executed. This is the best I can come up with short of fragile meta-
+  # programming with Puppet internals.
+  append_to_path '/usr/lib/rabbitmq/bin'
 
   def self.home_tmp_command(name, path)
     has_command name, path do
       environment HOME: '/tmp'
```
wyardley: Can't you just set PATH in `environment`, or does Puppet override it?

JayH5: Would be nice, but the `which` call is separate 😞 

https://github.com/puppetlabs/puppet/blob/111441de51724ba054cb122aece8324d173c3a8e/lib/puppet/provider/command.rb#L19-L24

The `environment` gets passed through `@options`.

(Don't know why the GitHub code embedding in comments is not working today)

JayH5:
There's nothing to do with rehashing here. The binaries in the `PATH` are not cached by Puppet and every time a command is called, Puppet re-searches the `PATH` directories, unless you give Puppet an absolute path, which is what we were doing.

This means we don't try to look up the full path to an executable at all until a command is actually invoked, by which time RabbitMQ should be installed.

> ps - I'm not sure what's up with Travis not running.

Embedded code references are not working either so I think something in GitHub may be broken.

wyardley:
You were doing `|| real path`, but if the entire code runs every time, shouldn't it pick up the bin in `PATH` first once it exists?

Travis is working again, should run if you force-push.

JayH5:
> You were doing `|| real path`, but if the entire code runs every time, shouldn't it pick up the bin in `PATH` first once it exists?

I'm not sure I understand what you are saying 😞

wyardley:
> > You were doing || real path, but if the entire code runs every time, shouldn't it pick up the bin in PATH first once it exists?

> I'm not sure I understand what you are saying 😞

https://github.com/voxpupuli/puppet-rabbitmq/blob/f6f7b6e83f9dfb52a8b982096d95d7415652d32f/lib/puppet/provider/rabbitmq_cli.rb#L5
If this is running every time, and it's not a hashing issue, shouldn't the result of `Puppet::Util.which(binary)` take precedence over the hard-coded path you mentioned?

JayH5:
> If this is running every time, and it's not a hashing issue, shouldn't the result of `Puppet::Util.which(binary)` take precedence over the hard-coded path you mentioned?

It's not running at every command execution, that's the thing. It's called here, once, I think each time a provider is loaded by Puppet. Not on command execution:
https://github.com/voxpupuli/puppet-rabbitmq/blob/f6f7b6e83f9dfb52a8b982096d95d7415652d32f/lib/puppet/provider/rabbitmq_cli.rb#L15

Yay Travis + code snippets working again

PR #774, opened 2019-02-07
Allow Array values in rabbitmq_parameter

vStone:
Array (lists) are allowed values to use with rabbitmqctl.

For example: src-uri when setting up a shovel.

@wyardley could you re-review this, had to create a new pr since I have left the vrtdev organisation since the previous one :)

### DIFF
--- a/spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_spec.rb
+++ b/spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_spec.rb
@@ -52,11 +52,11 @@ describe Puppet::Type.type(:rabbitmq_parameter).provider(:rabbitmqctl) do
 
   it 'matches parameters from list' do
     provider.class.expects(:rabbitmqctl_list).with('parameters', '-p', '/').returns <<-EOT
-shovel  documentumShovel  {"src-uri":"amqp://","src-queue":"my-queue","dest-uri":"amqp://remote-server","dest-queue":"another-queue"}
+shovel  documentumShovel  {"src-uri":["amqp://cl1","amqp://cl2"],"src-queue":"my-queue","dest-uri":"amqp://remote-server","dest-queue":"another-queue"}
 EOT
     expect(provider.exists?).to eq(component_name: 'shovel',
                                    value: {
-                                     'src-uri' => 'amqp://',
+                                     'src-uri' => ['amqp://cl1', 'amqp://cl2'],
                                      'src-queue'  => 'my-queue',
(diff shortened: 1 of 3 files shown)


spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_spec.rb
```diff
@@ -52,11 +52,11 @@
 
   it 'matches parameters from list' do
     provider.class.expects(:rabbitmqctl_list).with('parameters', '-p', '/').returns <<-EOT
-shovel  documentumShovel  {"src-uri":"amqp://","src-queue":"my-queue","dest-uri":"amqp://remote-server","dest-queue":"another-queue"}
+shovel  documentumShovel  {"src-uri":["amqp://cl1","amqp://cl2"],"src-queue":"my-queue","dest-uri":"amqp://remote-server","dest-queue":"another-queue"}
 EOT
     expect(provider.exists?).to eq(component_name: 'shovel',
                                    value: {
-                                     'src-uri' => 'amqp://',
+                                     'src-uri' => ['amqp://cl1', 'amqp://cl2'],
```
wyardley: Unless the output is always a list here, rather than change the existing test, I think we need a test for both string and list output formats?

vStone: Dont we already test for strings in the dest-uri. This isnt hard-coded on the src-uri so this example already covers both

wyardley: Hm. Ok.
I'm still a little confused about where it's a string representing an array and where it's an actual array. Presumably we provide it to Puppet as an array, but when it's passing it to the command line, it's a string formatted like an array, right?

vStone: That's all in the provider which is using json (parse and to_json). I'm not sure I should test if the json conversion works like it should.

https://github.com/voxpupuli/puppet-rabbitmq/blob/48e8723c55d019ad419a7430700d034a6843fa6a/lib/puppet/provider/rabbitmq_parameter/rabbitmqctl.rb#L20
https://github.com/voxpupuli/puppet-rabbitmq/blob/48e8723c55d019ad419a7430700d034a6843fa6a/lib/puppet/provider/rabbitmq_parameter/rabbitmqctl.rb#L78

wyardley:
@vStone Given that there's a single acceptance test for this already, I'm going to assume that the new use case won't be different, hopefully that's a valid assumption.

PR #778, opened 2019-02-13
Add a custom fact for rabbitmq's plugins folder.

TomRitserveldt:
If you want to add third party plugins with puppet you need to know
where the plugins folder is located.
This can differ depending on rabbitmq version or OS.


wyardley:
@TomRitserveldt: I'll restart the failed beaker job, but if it fails again, can you look into it?

PR #784, opened 2019-03-25
Don't crash on rabbitmq_plugins_dirs fact if rabbitmqctl is not present

### LINKED ISSUE #783: Fact rabbitmq_plugins_dirs can crash the puppet run

- Puppet: 5.5.6
- Ruby: 2.5.3
- https://github.com/voxpupuli/puppet-rabbitmq/blob/5477e69eccc009e3685d8eb7715e6dcae29966e6/lib/facter/rabbitmq_plugins_dirs.rb#L4-L5

Use the module on a system which doesn't have rabbitmqctl present. E.g. when we only want to generate configs but not manage RabbitMQ on that host.

## What are you seeing

"Error: Facter: error while resolving custom fact \"rabbitmq_plugins_dirs\": undefined method `[]' for nil:NilClass",

This happens because:
```ruby
      rabbitmq_pluginsdirs_env = Facter::Core::Execution.execute("rabbitmqctl eval 'application:get_env(rabbit, plugins_dir).'")
```
returns nil, and then we try to access index on that nil with `[1]`.

Ignore the absence of rabbitmqctl, return empty list for rabbitmq_plugins_dirs fact.

jistr:

Before trying to access the contents of a regexp match, see if the
match was successful. On machines without rabbitmqctl this match will
return nil. If we try to index the nil with `[1]`, it will crash.

Fixes #783

### DIFF
--- a/lib/facter/rabbitmq_plugins_dirs.rb
+++ b/lib/facter/rabbitmq_plugins_dirs.rb
@@ -2,8 +2,8 @@ Facter.add(:rabbitmq_plugins_dirs) do
   setcode do
     if Facter::Util::Resolution.which('rabbitmqctl')
       rabbitmq_pluginsdirs_env = Facter::Core::Execution.execute("rabbitmqctl eval 'application:get_env(rabbit, plugins_dir).'")
-      rabbitmq_plugins_dirs = %r{^\{ok\,\"(\/.+\/\w+)}.match(rabbitmq_pluginsdirs_env)[1]
-      rabbitmq_plugins_dirs.split(':')
+      rabbitmq_plugins_dirs_match = %r{^\{ok\,\"(\/.+\/\w+)}.match(rabbitmq_pluginsdirs_env)
+      rabbitmq_plugins_dirs_match[1].split(':') if rabbitmq_plugins_dirs_match
     end
   end
 end


lib/facter/rabbitmq_plugins_dirs.rb
```diff
@@ -2,7 +2,8 @@
   setcode do
     if Facter::Util::Resolution.which('rabbitmqctl')
       rabbitmq_pluginsdirs_env = Facter::Core::Execution.execute("rabbitmqctl eval 'application:get_env(rabbit, plugins_dir).'")
-      rabbitmq_plugins_dirs = %r{^\{ok\,\"(\/.+\/\w+)}.match(rabbitmq_pluginsdirs_env)[1]
+      rabbitmq_plugins_dirs_match = %r{^\{ok\,\"(\/.+\/\w+)}.match(rabbitmq_pluginsdirs_env)
+      rabbitmq_plugins_dirs = rabbitmq_plugins_dirs_match ? rabbitmq_plugins_dirs_match[1] : ''
```
wyardley: This looks basically right, I wonder if it can / should be nil vs an empty string? Also maybe not using the ternary operator would be a little easier to read?

jistr: Thanks for the review :).

Re nil vs. string: i think it should be an empty string, because on the line below we call .split(':') on it, which produces an empty array as the final fact value, which i think is correct. `nil` would crash on the `split` call.

Re ternary: I can change it, what would be your preference? Something like this?
```ruby
rabbitmq_plugins_dirs = if rabbitmq_plugins_dirs_match then rabbitmq_plugins_dirs_match[1] else '' end
```

wyardley: I think that would look good, maybe not all on the same line?

I'm pretty sure the fact itself can return an empty string, so can we return `nil` _after_ trying to split?

https://puppet.com/docs/facter/3.6/fact_overview.html

jistr: Ok i see, previously i didn't understand that you suggested to return `nil` instead of `[]` from the fact block itself. I can make it so :). I'll amend the patch for another review round.

lib/facter/rabbitmq_plugins_dirs.rb
```diff
@@ -2,8 +2,12 @@
   setcode do
     if Facter::Util::Resolution.which('rabbitmqctl')
       rabbitmq_pluginsdirs_env = Facter::Core::Execution.execute("rabbitmqctl eval 'application:get_env(rabbit, plugins_dir).'")
-      rabbitmq_plugins_dirs = %r{^\{ok\,\"(\/.+\/\w+)}.match(rabbitmq_pluginsdirs_env)[1]
-      rabbitmq_plugins_dirs.split(':')
+      rabbitmq_plugins_dirs_match = %r{^\{ok\,\"(\/.+\/\w+)}.match(rabbitmq_pluginsdirs_env)
+      if rabbitmq_plugins_dirs_match then
+        rabbitmq_plugins_dirs_match[1].split(':')
+      else
```
wyardley: Sorry, seems like I like this style but rubocop doesn’t (see Travis failures). I think you can just return `nil` if the condition doesn’t match with the if.

jistr: Yea if i follow the rubocop recommendation and just remove the `else` branch, it should work fine, the `if` should return `nil` in that case anyway.

jistr:
```ruby
Offenses:
lib/facter/rabbitmq_plugins_dirs.rb:6:7: C: Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||. (https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier)
      if rabbitmq_plugins_dirs_match
      ^^
```
:grin:

jistr:
The test for `rabbitmqctl` missing is in the specs already. So i wonder if the test maybe doesn't work as a real environment would... Not sure though, i'm not so much of a puppet/facter expert.

wyardley:
Thanks, yeah, just noticed that
https://github.com/voxpupuli/puppet-rabbitmq/blob/5477e69eccc009e3685d8eb7715e6dcae29966e6/spec/unit/facter/util/fact_rabbitmq_plugins_dirs_spec.rb#L32-L35
That's good enough for me, thanks for the contribution.

jistr:
Thanks for the review!

PR #785, opened 2019-03-28
Update the ssl_ciphers parameter to support the OpenSSL style

jamgregory:
Currently, this module only supports the old Erlang style, but RabbitMQ post-3.7.9 (rabbitmq/rabbitmq-server#1712) supports the OpenSSL style.

This change allows the rabbitmq.config file to determine which style is being used, and apply that when defining SSL cipher suites.

n/a

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -775,8 +775,12 @@ Default value: `undef`
 
 Data type: `Array`
 
-Support only a given list of SSL ciphers. Example: `['dhe_rsa,aes_256_cbc,sha','dhe_dss,aes_256_cbc,sha', 'ecdhe_rsa,aes_256_cbc,sha']`. Supported ciphers in your install can be listed with: rabbitmqctl eval 'ssl:cipher_suites().'
-Functionality can be tested with cipherscan or similar tool: https://github.com/jvehent/cipherscan.git
+Support only a given list of SSL ciphers, using either the Erlang or OpenSSL styles.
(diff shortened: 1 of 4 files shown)


jamgregory:
> Can you also update the puppet strings docs:
> 
> [puppet-rabbitmq/manifests/init.pp](https://github.com/voxpupuli/puppet-rabbitmq/blob/7613f08657d6640b22ec0f4cb55cf5b3c2f2e397/manifests/init.pp#L256)
> 
> Line 256 in [7613f08](/voxpupuli/puppet-rabbitmq/commit/7613f08657d6640b22ec0f4cb55cf5b3c2f2e397)
> 
>  # @param ssl_ciphers

Thanks @wyardley - I've updated that now.

jamgregory:
> Would it be hard to preserve the sorted ordering? If we release this, it may cause a reload for existing configs, no?

Ah, good point. The reason I removed that was because sorting them like that may cause less secure algorithms to float to the top above more secure ones (for example, DHE would come before ECDHE, but ECDHE is more secure).

I'm happy to put something in that causes the order to be preserved if the `ssl_honor_cipher_order` is set to true, otherwise it just sorts them - that should avoid existing configs being re-ordered?

wyardley:
Mm - I see what you’re saying. I don’t know if this warrants a new parameter, may be better to just release it like this (it makes the code easier to read too).

Let’s see if @bastelfreak thinks that should make this a breaking change.

PR #796, opened 2019-05-02
Fix permissions, based on scheme introduced by RabbitMQ 3.7.*

### LINKED ISSUE #703: Owner of /etc/rabbitmq

- Puppet: any
- Ruby: any
- Distribution: Gentoo
- Module version: Latest

Default code.

## What are you seeing

Wrong permissions on /etc/rabbitmq or NOT forced.

Permissions of /etc/rabbitmq should be force to be owned by `$rabbitmq_user`. Because, it's required by : `/usr/sbin/rabbitmq-plugins enable rabbitmq_management` which creates `enabled_plugins` file in config dir under rabbitmq_user.

## Output log

```
 # /usr/sbin/rabbitmq-plugins enable rabbitmq_management
Error: {cannot_read_en
```

dhoppe:
Fixes #703

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -144,14 +144,14 @@ class rabbitmq::config {
   file { '/etc/rabbitmq':
     ensure => directory,
     owner  => '0',
-    group  => '0',
-    mode   => '0755',
+    group  => $rabbitmq_group,
+    mode   => '2755',
   }
 
   file { '/etc/rabbitmq/ssl':
     ensure => directory,
     owner  => '0',
-    group  => '0',
+    group  => $rabbitmq_group,
     mode   => '0755',
   }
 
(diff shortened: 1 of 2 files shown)


manifests/config.pp
```diff
@@ -161,7 +161,7 @@
     content => template($config),
     owner   => '0',
     group   => $rabbitmq_group,
-    mode    => '0640',
+    mode    => '0644',
```
wyardley: If the directories themselves are readable by the group, as are the files, I think these files can (and should?) stay 0640?

dhoppe: I have my doubts, too. On the one hand, SSL certificates should not be readable by everyone, but on the other hand these defaults are given by the SGID bit via /etc/rabbitmq. 

Maybe this question can be answered by someone who has more experience with RabbitMQ?

wyardley: What do you mean given by the sgid bit on the directory? You can have files with mode 0640 in a directory that’s world readable or executable w sgid bit (say 2755 or 2751). Sgid bit just affects default group ownership of files created in the directory afaik

wyardley:
@dhoppe hope you don't mind - took a pass at a slightly narrower version of this in #835

PR #797, opened 2019-05-03
Use data-in-modules instead of params.pp

dhoppe:


dhoppe:
Do we really need to keep the ```server.pp```? It seems to be some kind of wrapper for backward compatibility and this will be a major release.

wyardley:
I think it was already deprecated. Should be fine to pull it out.

dhoppe:
@wyardley Yes, I wanted to make sure that the default values are fetched correctly.

wyardley:
@dhoppe Yeah, forgot there isn't really a task to do that automagically yet... :/

kobybr:
This should have been tagged as a breaking change.  Previous parameter references have been used in control repos.
 For example $rabbitmq::params::ssl_management_port is now $rabbitmq:::ssl_management_port.

wyardley:
Not much that can be done about it at this point even if that were the case, but I think params was always supposed to be private 

And the class was marked as private well before this version was released

https://github.com/voxpupuli/puppet-rabbitmq/blob/9924ca115bb0aecaa0f8b4158d23d44fee3b8610/manifests/params.pp#L3

kobybr:
Yes, I see that was added in 8.4.1 - however, we implemented this module prior to that.   As a volunteer effort I pointed this out not to complain, but to help for future releases.   Anyways, thanks for responding it is much appreciated.

Issue #781, opened 2019-03-15
Fetch signing keys from GitHub or Bintray


michaelklishin:
This module downloads signing key from rabbitmq.com. Team RabbitMQ deprecated downloads from rabbitmq.com a couple of years ago (see [Signatures](https://www.rabbitmq.com/signatures.html)). Why? We don't want to be distributing artifacts, there are services that do it better.

The key is available from [Bintray](https://dl.bintray.com/rabbitmq/Keys/rabbitmq-release-signing-key.asc) and [GitHub](https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc). Please switch to one of those locations.

The current key isn't going to be removed from rabbitmq.com but when it's time to renew, it may or may not make the cut.

wyardley:
Are there any releases for which `F4E789204D206F89` is still valid, and if so, is that accessible from anywhere?

michaelklishin:
@wyardley we've migrated to the new key [about 3 years ago](https://www.rabbitmq.com/key-transition-0xF7B8CEA6056E8E56.txt) ([mailing list announcement](https://groups.google.com/forum/#!msg/rabbitmq-users/BO5cmEsdEhc/Jupz1_Q4AwAJ)). 3.5.8 and early 3.6.x releases were all signed with the new key. We re-signed even legacy apt repositories on rabbitmq.com IIRC.

The old key is [available from Bintray](https://bintray.com/beta/#/rabbitmq/Keys/SigningKey/1.0.0?tab=files). I honestly thing we can consider it to be irrelevant.

wyardley:
@michaelkishlin I have to double check - module’s default behavior actually ships 3.3.x (from the vendors’ repos) on certain platforms, tho I guess in that case it will probably be signed with the vendor’s key? I feel like there’s a reason that we have the old key imported in one or two places, but could be wrong.

juniorsysadmin:
Ideally the public key should be included with this module and not fetched as well.

wyardley:
@juniorsysadmin I agree that that's probably the most secure way. Do you have time / inclination to throw up a PR to switch it to this pattern?

juniorsysadmin:
@wyardley I have sadly not much time for this at the moment.

PR #798, opened 2019-05-03
Refactor manifests

dhoppe:
The following classes still have to be refactored:
- [ ] config.pp
- [ ] server.pp
- [ ] repo/apt.pp
- [ ] repo/rhel.pp

The classes ```apt.pp``` and ```rhel.pp``` depend on pr #797. 

I also need to know which Mirror we want to use as default and change the GPG key. I would like to use Bintray, because it provides packages for Erlang. 

Because of the many changes to the code we will have to release a major release anyway.


wyardley:
#781 has some more info about the keys. I like the suggestion to just bake the files directly into the module; this is probably the most secure option.

wyardley:
Also, I don’t have a preference on mirror, but packagecloud has their erlang packages too:
 https://packagecloud.io/rabbitmq/erlang

dhoppe:
> Also, I don’t have a preference on mirror, but packagecloud has their erlang packages too:
> https://packagecloud.io/rabbitmq/erlang

Yes, but they only provide Erlang packages for RedHat/CentOS.

dhoppe:
> #781 has some more info about the keys. I like the suggestion to just bake the files directly into the module; this is probably the most secure option.

Yes, but I do not know if we just need to provide the latest GPG key. My understanding was that they resigned all packages with the current GPG key.

bastelfreak:
@dhoppe can you please rebase?

wyardley:
@dhoppe do you want to work on this some more? I probably should cut a major release soon, and it would be nice to include these changes if you're able to update them.

wyardley:
@dhoppe I'm going to close this one for now; feel free to reopen / revisit when you're ready to work on this again.

PR #810, opened 2019-08-14
Release 9.1.0

wyardley:
* Update CHANGELOG, REFERENCE and metadata for 9.1.0 release

Note: I'm not 100% sure whether the update to data-in-modules is breaking. Since #798 is not updated, and there don't _seem_ be be breaking changes based on PR descriptions, I think we can make this 9.1.0, but open to suggestions / corrections here.

n/a


bastelfreak:
I will do the release now

PR #814, opened 2019-10-10
Extend version regex for RabbitMQ 3.8

codeinthehole:
Where the output from `rabbitmqctl -q status` isn't printed as an Erlang
property list.

Without this change, you get a crash with error:

> Failed to apply catalog: undefined method `scan' for nil:NilClass


codeinthehole:
@wyardley Good point. Have fixed the tests and force-pushed back up.

bastelfreak:
@codeinthehole thanks for the PR. Can you add an acceptance test that verifies this works?

wyardley:
@bastelfreak unless they’ve been reworked, I think the acceptance tests still are only using the old version of RMQ and erlang?

I can double check, but if so, it’s actually a pretty non-trivial task to add.

bastelfreak:
ah, right.

wyardley:
Maybe an additional unit test would have been good though?

PR #815, opened 2019-10-19
Clean up acceptance spec helper

ekohl:


spec/spec_helper_acceptance.rb
```diff
@@ -3,20 +3,21 @@
...
   c.formatter = :documentation
+
+  # Configure all nodes in nodeset
   c.before :suite do
+    install_module
+    install_module_dependencies
+
+    # Install aditional modules for soft deps
+    install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 8.0.0') if fact('os.family') == 'Debian'
+    install_module_from_forge('garethr-erlang', '>= 0.3.0 < 1.0.0') if fact('os.family') == 'RedHat'
```
wyardley: FWIW, this one isn’t well maintained and may not work with newer Erlang that’s required by newer RMQ versions. Long term, we should either adopt and fix it, or use one of the alternative ways suggested to satisfy the Erlang dep.

PR #819, opened 2019-12-01
Fix no returning rabbitmq_version and add default version?

### LINKED ISSUE #817: Version release
Do you think a new version release would be possible soon?
We are interested particularly in https://github.com/voxpupuli/puppet-rabbitmq/pull/814 since the module is currently not working with RabbitMQ version 3.8.0.

- Puppet: 4.10.12
- Ruby: -
- Distribution: CentOS 7
- Module version: 9.1.0

Include module

## What are you seeing

```Failed to apply catalog: undefined method `scan' for nil:NilClass```

No error

### PR DESCRIPTION
**Pull request by justintime4tea**
Signed-off-by: Justin Gross <justin@jgross.tech>

I'm installing RabbitMQ 3.8.1 using rabbitmq module params as such:
```
management_enable          => true
package_name                    => rabbitmq-server-3.8.1
service_name                      => rabbitmq-server
use_config_file_for_plugins => true
```

I attempt to add user:
```
 rabbitmq_user { 'SOME_USER':
    password => 'SOME_PASS',
 }
```

A check is done to see what version rabbit is so to know which command line options to use when running the rabbitmq CLI (-q --no-table-headers, or just -q) to determine if the user already exists or not.

The self.rabbitmq_version function returns a version if @rabbitmq_version is declared. 

**When not declared** however it apparently tries to declare it (not a ruby developer) but then afterwords never returns it so whomever has called it is receiving a NilClass? (at least the very first time? presumably, because it sets the @rabbitmq_server?) 

The error produced is as such: 
"Failed to apply catalog: undefined method `scan' for nil:NilClass" because rabbitmq_version never returns anything unless it is globally defined (@rabbitmq_version)

This error is referenced by this issue:
https://github.com/voxpupuli/puppet-rabbitmq/issues/817

### DIFF
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -40,6 +40,8 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
     output = rabbitmqctl('-q', 'status')
     version = output.match(%r{RabbitMQ.*?([\d\.]+)})
     @rabbitmq_version = version[1] if version
+    return version if version
+    return '3.8.1'
   end
 
   def self.rabbitmqctl_list(resource, *opts)
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -40,7 +40,7 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
     output = rabbitmqctl('-q', 'status')
     version = output.match(%r{RabbitMQ.*?([\d\.]+)})
     @rabbitmq_version = version[1] if version
-    return version if version
+    return version[1] if version
     return '3.8.1'
   end
 

### CODE REVIEW

File: lib/puppet/provider/rabbitmq_cli.rb [hunk 1]
```diff
@@ -40,6 +40,8 @@ def self.rabbitmq_version
     output = rabbitmqctl('-q', 'status')
     version = output.match(%r{RabbitMQ.*?([\d\.]+)})
     @rabbitmq_version = version[1] if version
+    return version[1] if version
+    return '3.8.1'
```
wyardley: The module’s default behavior is to install 3.6.x, IIRC. I’m not sure if line 43 is necessary, but pretty sure this is the wrong behavior.

Do you get this issue with the most recent release, or with git master?

wyardley:
Have you verified if #814 resolves this?

krjensen:
> Have you verified if #814 resolves this?

I have just tested this on Ubuntu 18.04 with puppet 3.8.1 and it solved my problem.

wyardley:
> I have just tested this on Ubuntu 18.04 with puppet 3.8.1 and it solved my problem.

“This” meaning the latest release? I.e., can we close this PR?

krjensen:
I checked #814 directly - not the latest release, but I will do that first thing next week.

PR #820, opened 2019-12-02
Release 10.0.0 + modulesync

### LINKED ISSUE #817: Version release

wyardley:
## Pull Request (PR) description
BREAKING CHANGE: Drop Ubuntu 14.04 support (#818)
- Extend version regex for RabbitMQ 3.8 (#814)
- Clean up acceptance spec helper (#815)

Fixes #817

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+## [v10.0.0](https://github.com/voxpupuli/puppet-rabbitmq/tree/v10.0.0) (2019-12-02)
+
+[Full Changelog](https://github.com/voxpupuli/puppet-rabbitmq/compare/v9.1.0...v10.0.0)
+
+**Breaking changes:**
+
+- drop Ubuntu 14.04 support [\#818](https://github.com/voxpupuli/puppet-rabbitmq/pull/818) ([bastelfreak](https://github.com/bastelfreak))
+- Update the ssl\_ciphers parameter to support the OpenSSL style [\#785](https://github.com/voxpupuli/puppet-rabbitmq/pull/785) ([jamgregory](https://github.com/jamgregory))
+
+**Implemented enhancements:**
+
+- Extend version regex for RabbitMQ 3.8 [\#814](https://github.com/voxpupuli/puppet-rabbitmq/pull/814) ([codeinthehole](https://github.com/codeinthehole))
+
+**Merged pull requests:**
+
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -158,7 +175,6 @@ These should not affect the functionality of the module.
 **Fixed bugs:**
 
 - rabbitmq\_user\_permissions timing out [\#689](https://github.com/voxpupuli/puppet-rabbitmq/issues/689)
-- Set default LC\_ALL =\> en\_US.UTF-8 \(\#671, \#689\) [\#694](https://github.com/voxpupuli/puppet-rabbitmq/pull/694) ([wyardley](https://github.com/wyardley))
```
ekohl: That this moves around could be a caching issue. Can you do a modulesync and use our fork to generate it?

bastelfreak: a modulesync from the master branch of modulesync_config should fix it

ekohl:
Given this is breaking, should it also merge https://github.com/voxpupuli/puppet-rabbitmq/pull/796 to avoid doing too many breaking changes? Perhaps there are more.

CHANGELOG.md
wyardley: Doing it, one sec

wyardley:
> Given this is breaking, should it also merge #796 to avoid doing too many breaking changes? 
> Perhaps there are more.

I'm not sure #796 is breaking, but I also don't think it's the right behavior still.

CHANGELOG.md
wyardley: Is this right now?

wyardley:
Since modulesync isn't in its own PR, I shouldn't add the modulesync label, right?

wyardley:
@ekohl @bastelfreak I never got feedback on #785, but I think that one would be backwards-incompatible... from a security standpoint, this one is probably a good one to release, even though it changes behavior... however, because of the change of order, I think it could cause a reload of configs?

bastelfreak:
I approved #785. Please also run `bundle exec rake reference` to ensure that the REFERENCE.md is up2date.

wyardley:
Thanks @bastelfreak!

wyardley:
@ekohl @bastelfreak: Updated changelog. Reference seems up-to-date.

PR #826, opened 2020-01-03
Add support for RHEL/CentOS 8

### LINKED ISSUE #816: Add support for RHEL/CentOS 8
The module is currently not supported on RHEL/CentOS 8. Please add support for those os versions.

dhoppe:
Fixes #816

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '2.10.0'
+modulesync_config_version: '2.10.1'
--- a/.sync.yml
+++ b/.sync.yml
@@ -8,6 +8,7 @@
     - set: ubuntu1804-64
     - set: centos6-64
     - set: centos7-64
+    - set: centos8-64
 appveyor.yml:
   delete: true
 spec/spec_helper.rb:
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,14 @@ matrix:
     bundler_args: --without development release
     env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_debug=true BEAKER_setfile=centos7-64 BEAKER_HYPERVISOR=docker CHECK=beaker
     services: docker
+  - rvm: 2.5.3
+    bundler_args: --without development release
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_debug=true BEAKER_setfile=centos8-64 BEAKER_HYPERVISOR=docker CHECK=beaker
+    services: docker
+  - rvm: 2.5.3
+    bundler_args: --without development release
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_debug=true BEAKER_setfile=centos8-64 BEAKER_HYPERVISOR=docker CHECK=beaker
+    services: docker
 branches:
   only:
   - master
--- a/metadata.json
+++ b/metadata.json
@@ -12,14 +12,16 @@
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
         "6",
-        "7"
+        "7",
+        "8"
       ]
     },
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
         "6",
-        "7"
+        "7",
+        "8"
       ]
     },
     {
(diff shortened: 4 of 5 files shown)


bastelfreak:
This depends on https://github.com/camptocamp/facterdb/pull/137 :(. We need to wait with the merge of this PR or we pin facterdb to the commit with CentOS 8 factsets.

dhoppe:
@bastelfreak I know that camptocamp takes some time to merge pull requests, but pinning facterdb to the commit ID feels dirty, too.

bastelfreak:
I think pinning it is totally fine as a workaround. We already did that in a few other modules as well. I also pinged Raphael again to kick of a facterdb release.

ghoneycutt:
This seems to be failing because it is trying to download puppet for fedora 31 which does not exist. http://yum.puppetlabs.com/puppet5-release-fedora-30.noarch.rpm is the latest release.

dhoppe:
@ghoneycutt This is not related to Fedora 31. The build jobs for CentOS 8 fail. 

By default the parameter ```repos_ensure``` is set to false, but the module ```garethr/erlang``` will enable the EPEL repository, which is used to install the package ```rabbitmq-server```. But the EPEL repository does not contain a ```rabbitmq-server``` package for CentOS 8. 

Does anybody know, why this module does not use the Packagecloud repository to install RabbitMQ?

wyardley:
@dhoppe originally, the RMQ packages were the default. Quite a while back, when Puppet was still maintaining this module, I talked with the folks there and decided to rework it to default to using vendor packages vs packagecloud. I don’t remember all of the details, probably may be some more in the PR comments (#493).

At that time, I don’t think the OS vendor packages were so far out of date, and seemed like a sane default.

As far as the erlang dep, the erlang module isn’t really maintained now, so adjusting the strategy of which erlang version to get from where may also make some sense.

However, I haven’t had the time, need, or inclination to do the structural work involved in redoing the erlang installation, default package version, and integration tests.

As far as erlang dep goes, I think one challenge is making the module have somewhat sane defaults but also deal with cases where erlang is already present from something else the (module) user has done.

nmaludy:
Restarting build job as https://github.com/camptocamp/facterdb/pull/137 was merged

dhoppe:
Travis CI still fails, because stahnma/puppet-module-epel#61 has not been merged.

wyardley:
@dhoppe let me know if you want any more input / background on this. I don't have the energy / need to make changes, but happy to review, and I do think it does make sense to do a major where we switch back to the default of using the RabbitMQ repos.

The thing (maybe mentioned in some other threads) that's also tricky is the different possible methods of satisfying the erlang deps, and the fact that, I think, the puppet erlang module isn't really maintained. So any updates would probably need to accommodate a couple different ways of satisfying the erlang deps.

nmaludy:
I've been looking into this because a module i maintain `stackstorm/st2` depends on this module for RabbitMQ. I think the current piece being held up is that the current erlang module is no longer maintained and EPEL for EL8 doesn't contain the erlang dependencies we need.

I'll see if next week i can take a swing a making a new erlang module and donating it to Vox so we can maintain this as a community going forward.

Thoughts?

nmaludy:
Sorry, i take that back, it looks like erlang installs correctly, but EPEL no longer includes `rabbitmq-server`. We may need to rely on the PackageCloud or Bintray repositories on EL8 to get rabbitmq-server: https://www.rabbitmq.com/install-rpm.html

wyardley:
That would be great (new erlang module).

wyardley:
@nmaludy it may work already if you set `repos_ensure` to `true`

nmaludy:
@wyardley OK, i'll give that a shot on my local box. We may have to adjust the acceptance tests for el8 to set that as default, then probably set it up that way in `params.pp` and/or document it in the README.

nmaludy:
@wyardley I was able to get this working on a local CentOS 8 Vagrant box with the following:
```
class { 'erlang': epel_enable => true, }
~> class { 'rabbitmq' : }
```

But, i also had to modify `hiera.yaml` to add a new data layer:
```yaml
  - name: 'Operating System Family Version'
    path: "family/%{facts.os.family}-%{facts.os.release.major}.yaml"
```

And then add the following hiera data file: `data/family/RedHat-8.yaml`
```yaml
---
rabbitmq::python_package: 'python3'
rabbitmq::repos_ensure: true
```

Going to create a PR now.

nmaludy:
New PR is available here: #842

PR #831, opened 2020-04-06
Use voxpupuli-acceptance

ekohl:


spec/spec_helper_acceptance.rb
```diff
@@ -116,3 +105,4 @@
     apply_manifest(pp, catch_changes: true)
   end
 end
+>>>>>>> acceptance-copy
```
wyardley, line 108: @ekohl is this supposed to be here, or from conflict resolution?

ekohl: So I didn't properly restore it. I'll have to look into this.

PR #833, opened 2020-04-10
Fix issue #832 - removing federation upstream

### LINKED ISSUE #832: Problem removing (ensure => 'absent') federation upstream

- Puppet: puppet-agent 5.5.19-1stretch
- Ruby: ruby 1:2.3.3
- Distribution: debian stretch
- Module version: Version 10.0.0 of https://forge.puppet.com/puppet/rabbitmq

1. create federation upstream (ensure: present) ... upstream gets created:
```
rabbitmq_parameter { 'nonprod_settlement_develop@/develop':
  component_name => 'federation-upstream',
  ensure => 'present',
  value => {
    'uri' => 'amqps://fed_user:fed_pass@upstream_server:5672/%2fdevelop',
    'ack-mode' => 'on-confirm',
    'max-hops' => '10',
    'trust-user-id' => false,
  },
}
```
2. remove federation upstream (ensure: absent) ... you receive error removing inexistent shovel:
```
rabbitmq_parameter { 'nonprod_settlement_develop@/develop':
  component_name => 'federation-upstream',
  ensure => 'absent',
  value => {
    'uri' => 'amqps://fed_user:fed_pass@upstream_server:5672/%2fdevelop',
    'ack-mode' => 'on-confirm',
    'max-hops' => '10',
    'trust-user-id' => false,
  },
}
```

## What are you seeing
It looks like the removal operation tries to remove **shovel** instead of federation upstream.

I expected federation upstream named 'nonprod_settlement_develop' inside vhost '/develop' to be removed from RabbitMQ cluster... and not observing an attempt to remove shovel with the same name inside the same vhost.

## Output log
Error: Execution of '/usr/sbin/rabbitmqctl clear_parameter -p /develop shovel nonprod_settlement_develop' returned 69: Clearing runtime parameter "nonprod_settlement_develop" for component "shovel" on vhost "/develop" ...
Error:
Parameter does not exist
Error: /Stage[main]/Messagebroker::Rabbitmq/Rabbitmq_parameter[nonprod_settlement_develop@/develop]/ensure: change from 'present' to 'absent' failed: Execution of '/usr/sbin/rabbitmqctl clear_parameter -p /develop shovel nonprod_settlement_develop' returned 69: Clearing runtime parameter "nonprod_settlement_develop" for component "shovel" on vhost "/develop" ...
Error:
Parameter does not exist

I have not one shovel defined anywhere. We use RabbitMQ 3.8.3 with Erlang 22.0.7.

fiksn:

I just fixed the bug, there should be some better unit test as well.

Fixes #832

### DIFF
--- a/spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_federation_spec.rb
+++ b/spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_federation_spec.rb
@@ -1,0 +1,74 @@
+require 'spec_helper'
+
+provider_class = Puppet::Type.type(:rabbitmq_parameter).provider(:rabbitmqctl)
+describe provider_class do
+  let(:resource) do
+    Puppet::Type.type(:rabbitmq_parameter).new(
+      name: 'documentumFederation@/',
+      component_name: 'federation',
+      value: {
+        'uri'      => 'amqp://',
+        'expires'  => '360000'
+      }
+    )
+  end
+  let(:provider) { provider_class.new(resource) }
+
+  after do
+    described_class.instance_variable_set(:@parameters, nil)
+  end
+
+  describe '#prefetch' do
+    it 'exists' do
+      expect(described_class).to respond_to :prefetch
+    end
+
+    it 'matches' do
+      provider_class.expects(:rabbitmqctl_list).with('vhosts').returns <<-EOT
+/
+EOT
+      provider_class.expects(:rabbitmqctl_list).with('parameters', '-p', '/').returns <<-EOT
+federation  documentumFederation  {"uri":"amqp://","expires":360000}
+EOT
+      provider_class.prefetch('documentumFederation@/' => resource)
+    end
+  end
+
+  describe '#instances' do
+    it 'exists' do
+      expect(described_class).to respond_to :instances
+    end
+
+    it 'fail with invalid output from list' do
+      provider_class.expects(:rabbitmqctl_list).with('vhosts').returns <<-EOT
+/
+EOT
+      provider.class.expects(:rabbitmqctl_list).with('parameters', '-p', '/').returns 'foobar'
+      expect { provider_class.instances }.to raise_error Puppet::Error, %r{cannot parse line from list_parameter}
+    end
+
+    it 'return no instance' do
+      provider_class.expects(:rabbitmqctl_list).with('vhosts').returns <<-EOT
+/
+EOT
+      provider_class.expects(:rabbitmqctl_list).with('parameters', '-p', '/').returns ''
+      instances = provider_class.instances
+      expect(instances.size).to eq(0)
+    end
+  end
+
+  describe '#create' do
+    it 'create parameter' do
+      provider.expects(:rabbitmqctl).with('set_parameter', '-p', '/', 'federation', 'documentumFederation',
+                                          '{"uri":"amqp://","expires":360000}')
+      provider.create
+    end
+  end
+
+  describe '#destroy' do
+    it 'destroy parameter' do
+      provider.expects(:rabbitmqctl).with('clear_parameter', '-p', '/', 'federation', 'documentumFederation')
+      provider.destroy
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


bastelfreak:
@fiksn thanks for fixing this, can you somehow add a test to ensure this works now as expected?

fiksn:
@bastelfreak: Yeah, adding tests would be my wish too, but unfortunately I am not that fluent with Ruby (yet). For now the test on my side is referencing my repo with the change in Puppetfile.

fiksn:
I've added some unit tests and ran the stuff with `docker build .` It fails with the old version of code and passes with the new. So I suppose it should be good enough 🙂

wyardley:
First glance, LGTM, but haven’t been doing much Puppet or Ruby development recently, so probably good to get some eyes on it, especially in terms of whether the tests test the right things.

Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ruby:2.5.3
 
 WORKDIR /opt/puppet
+ENV LC_ALL=C.UTF-8
```
wyardley, line 4: Is this needed or just a byproduct of something you added for testing locally? Not sure if this file comes from modulesync

spec/unit/puppet/provider/rabbitmq_parameter/rabbitmqctl_federation_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    described_class.instance_variable_set(:@parameters, nil)
+  end
+
+  describe '#prefetch' do
+    it 'exists' do
+      expect(described_class).to respond_to :prefetch
+    end
+
+    it 'matches' do
+      provider_class.expects(:rabbitmqctl_list).with('vhosts').returns <<-EOT
```
ekohl, line 27: In our modules we've been moving away from setting `expects` before the test but rather use `allow()`, then the test and then `expect().to have_received()`.
https://relishapp.com/rspec/rspec-mocks/docs/basics/spies#spy-on-a-method-on-a-partial-double

wyardley: Makes sense; given that the submitter said they’re not very experienced with tests, and that this matches the style of the rest of the tests, I’m inclined to say let’s accept this, and then someone can refactor the tests to the newer style later if it makes sense?

ekohl: I'm fine with that.

Dockerfile
fiksn: Apparently the default locale is set to "POSIX" and when I run the test through docker container rake fails with "ArgumentError: invalid byte sequence in US-ASCII". I agree that this might be a bit offtopic for this issue, but fact is that without that line `docker build .` fails. Of course if I installed and ran everything locally stuff still passes.

Not sure what should be done here. Perhaps I can squash my two commits that deal with this bug and then create a separate commit / pull request for the issue that test failure when using docker. Perhaps there is even some nicer way (like that it is expected to pass some -e LC_ALL.. flag, but didn't find any documentation about it).

fiksn:
I am more than happy to contribute. Sorry for not being responsive lately.

Removed the part from Dockerfile now and squashed the two commits (fix and unit tests). Passing environment variables won't work as easy as I thought since this is the "build" and not "run" phase (you can just add --build-arg). Do you believe it makes sense to open another issue for that one-liner?

wyardley:
@fiksn so, I'm not doing too much work on Vox modules anymore, but the `Dockerfile` comes from https://github.com/voxpupuli/modulesync_config, so we can't make modifications directly to it in this module vs. upstream.

We have seen a few issues with locale settings like `LANG` or `LC_ALL` in this module, and there are some existing workarounds (e.g., 6c7a1467370ac926b48d6aa3c7aed4a78301b175) related to the CLI tools needing those to be set. These issues would only come up on systems where it wasn't set system-wide, so this may be a similar kind of issue.

It may be that more of those need to be made, vs. adjusting the environment in the test env, as this could actually be reflecting a real problem in cases where that env isn't set at system level. As the tests work in CI, I think we can assume the module is working, for the most part.

PR #835, opened 2020-04-26
[fix] ownership and permissions on conf files

### LINKED ISSUE #703: Owner of /etc/rabbitmq

### LINKED ISSUE #813: breaks /etc/rabbitmq ownership under ubuntu
Puppet: 5.4.0
    Ruby: 2.5.1
    Distribution: Ubuntu 18.04
    Module version: 9.1.0

I found that the puppet-rabbitmq does not work when used repeatedly under Ubuntu, since puppet-rabbitmq changes the ownership of /etc/rabbitmq to root, while under Ubuntu (although I agree that this is poor design) it is supposed to be rabbitmq.rabbitmq, since rabbitmq-plugins want's to write /etc/rabbitmq/enabled_plugins and needs to temporarily create /etc/rabbitmq/enabled_plugins.tmp, therefore needs write access to /etc/rabbitmq.

wyardley:
* Set owner of config directories to rabbitmq user rather than root
* Set group of config directories explicitly to rabbitmq, and set setgid bit
* Set more restrictive permissions on the SSL directory
* Update expectations, and test file permissions in acceptance tests

This is a narrower version of #796 (@dhoppe)
This does have some security implications, since when the file is owned by root, if the application is compromised, the rabbitmq user still may not be able to modify configuration files; with these changes, it would.

I _believe_ that changing the ownership this way should be safe even on older RabbitMQ versions. However, would appreciate some eyes and / or actual testing of this, as I'm not in a good position to do extensive validations beyond the acceptance tests.

Fixes #703
Closes #796
Fixes #813

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -37,6 +37,19 @@ describe 'rabbitmq class:' do
       it { is_expected.to be_enabled }
       it { is_expected.to be_running }
     end
+
+    describe file('/etc/rabbitmq') do
+      it { is_expected.to be_directory }
+      it { is_expected.to be_owned_by 'rabbitmq' }
+      it { is_expected.to be_grouped_into 'rabbitmq' }
+    end
+
+    describe file('/etc/rabbitmq/ssl') do
+      it { is_expected.to be_directory }
+      it { is_expected.to be_owned_by 'rabbitmq' }
+      it { is_expected.to be_grouped_into 'rabbitmq' }
+      it { is_expected.not_to be_readable.by('others') }
+    end
   end
 
   context 'disable and stop service' do
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -146,23 +146,23 @@ class rabbitmq::config {
 
   file { '/etc/rabbitmq':
     ensure => directory,
-    owner  => '0',
-    group  => '0',
-    mode   => '0755',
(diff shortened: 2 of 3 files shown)


spec/acceptance/class_spec.rb
```diff
@@ -37,6 +37,19 @@ class { 'erlang': epel_enable => true}
       it { is_expected.to be_enabled }
       it { is_expected.to be_running }
     end
+
+    describe file('/etc/rabbitmq') do
+      it { is_expected.to be_directory }
+      it { is_expected.to be_owned_by 'rabbitmq' }
+      it { is_expected.to be_grouped_into 'rabbitmq' }
+    end
```
wyardley, line 45: _think_ this is the right place for these to live?
seems sensible to test these in the acceptance tests one way or another, I think, right?

Issue #839, opened 2020-06-04
rabbitmq_user resource displays password when needed changed in noop


kobybr:
When running puppet agent in --noop mode, rabbitmq_user will display the 'should be' password if it needed changed.

i.e.
Rabbitmq_user[testuser]/password: current_value , should be somerandompassowrd (noop)

PR #840, opened 2020-06-22
Hide user password

tobias-urdin:
Override the is_to_s and should_to_s
methods in the puppet property to make
sure it's not shown in outputs.

Closes: #839


wyardley:
> I haven't done too much with this but I think this is correct. Perhaps @alexjfisher knows better.

Thanks. From memory, I think this is the right fix as well. Not sure about the specific wording, but seems fine.

PR #841, opened 2020-06-23
Don't use RABBITMQ_SERVER_ERL_ARGS

jeckersb:
Instead, use RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS

Quoted from https://www.rabbitmq.com/configure.html#supported-environment-variables:

"
RABBITMQ_SERVER_ERL_ARGS - Standard parameters for the erl command
used when invoking the RabbitMQ Server. This should be overridden for
debugging purposes only. Overriding this variable replaces the default
value.
"

The default value of RABBITMQ_SERVER_ERL_ARGS on *nix is:

+P 1048576 +t 5000000 +stbt db +zdbbl 128000

So currently, if ipv6/tls is enabled, and the caller does not provide
its own default value of RABBITMQ_SERVER_ERL_ARGS, it will end up
overwriting those server defaults.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -125,7 +125,7 @@ class rabbitmq::config {
       $proto_dist = 'inet6_tcp'
       $ssl_path = ''
     }
-    $ipv6_or_tls_env = ['SERVER', 'CTL'].reduce({}) |$memo, $item| {
+    $ipv6_or_tls_env = ['SERVER_ADDITIONAL', 'CTL'].reduce({}) |$memo, $item| {
       $orig = $_environment_variables["RABBITMQ_${item}_ERL_ARGS"]
       $munged = $orig ? {
         # already quoted, keep quoting
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -1326,7 +1326,7 @@ describe 'rabbitmq' do
         context 'without other erl args' do
           it 'enables inet6 distribution' do
             is_expected.to contain_file('rabbitmq-env.config'). \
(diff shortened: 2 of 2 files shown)


jeckersb:
@mbaldessari

wyardley:
@jeckersb If this is a result of changes, do you know which version the change happened? Does this PR risk breaking things on older RabbitMQ versions (and if so, which ones)?

wyardley:
ah, I see, appending vs. overriding (https://www.rabbitmq.com/runtime.html)
Assuming this works on all older versions, seems good to me.

jeckersb:
> I don't know too much about RabbitMQ but does this remove the option to use `RABBITMQ_SERVER_ERL_ARGS`?

No, if you want to set RABBITMQ_SERVER_ERL_ARGS you can still pass it in as part of the environment_variables hash.  This just changes the behavior when the ipv6 or ssl_erl_dist flags are enabled, since they require munging the server cmdline to enable the right flags.  With this patch the munging is done against the RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS variable so the rabbitmq system defaults are preserved.

jeckersb:
> @jeckersb If this is a result of changes, do you know which version the change happened? Does this PR risk breaking things on older RabbitMQ versions (and if so, which ones)?

The ADDITIONAL variable was added here - https://github.com/rabbitmq/rabbitmq-server/commit/9447629717e621454ee2f8735f2772f8455390e5

Which is part of RabbitMQ v3.4.0, and per https://www.rabbitmq.com/versions.html the 3.3 series went out of support on 31 March 2015.  So in theory it could break someone who insists on using a RabbitMQ version that has been out of support for over 5 years now, but really insists on the latest puppet-rabbitmq :)

PR #842, opened 2020-06-29
[WIP] Add CentOS/RHEL 8 Support

### LINKED ISSUE #816: Add support for RHEL/CentOS 8

nmaludy:
This PR adds support for installing RabbitMQ on CentOS 8 / RHEL 8.
Also added back into the README information to the end-user about the need to use an external `erlang` module on RHEL systems.

Fixes #816 

Closes #826

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -8,6 +8,8 @@
     - set: ubuntu1804-64
     - set: centos6-64
     - set: centos7-64
+    # RabbitMQ fails to start via systemd unless /sys/fs/cgroup is mounted through from to base host
+    - set: centos8-64{mount_folders={sysfscgroup={host_path=/sys/fs/cgroup\,container_path=/sys/fs/cgroup\,opts=ro}}}
 appveyor.yml:
   delete: true
 spec/spec_helper.rb:
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@
 4. [Usage - Configuration options and additional functionality](#usage)
 5. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
 5. [Limitations - OS compatibility, etc.](#limitations)
-   * [RedHat module dependencies](#redhat-module-dependecies)
 6. [Development - Guide for contributing to the module](#development)
 
 ## Overview
@@ -53,6 +52,38 @@ You can view example usage in [REFERENCE](REFERENCE.md).
 ## Reference
 
 See [REFERENCE](REFERENCE.md).
+
+## Limitations
+
+If running CentOS/RHEL, ensure the `epel` repo, or another repo containing a
+suitable Erlang version, is present. On Debian systems, puppetlabs/apt
+(>=2.0.0 < 5.0.0) is a soft dependency.
+
+To have a suitable erlang version installed on RedHat and Debian systems,
+you have to install another puppet module from https://forge.puppetlabs.com/garethr/erlang with:
+
+    puppet module install garethr-erlang
+
+This module handles the packages for erlang.
+To use the module, add the following snippet to your site.pp or an appropriate profile class:
+
+For RedHat systems:
+```puppet
+include 'erlang'
+class { 'erlang':
+  epel_enable => true,
+}
+```
+
+For Debian systems:
+```puppet
+include 'erlang'
+package { 'erlang-base':
+  ensure => 'latest',
+}
+```
+
+This module also depends on voxpupuli/archive to install rabbitmqadmin.
 
 ## Development
 
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,0 +1,3 @@
+# Vox Pupuli Security Policy
+
+Our vulnerabilities reporting process is at https://voxpupuli.org/security/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+os: linux
 dist: bionic
 language: ruby
 cache: bundler
@@ -7,7 +8,7 @@ before_install:
   - bundle --version
 script:
   - 'bundle exec rake $CHECK'
-matrix:
+jobs:
   fast_finish: true
   include:
   - rvm: 2.4.4
@@ -70,6 +71,14 @@ matrix:
     bundler_args: --without development release
     env: BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_setfile=centos7-64 CHECK=beaker
     services: docker
+  - rvm: 2.5.3
+    bundler_args: --without development release
+    env: BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_setfile=centos8-64{mount_folders={sysfscgroup={host_path=/sys/fs/cgroup\,container_path=/sys/fs/cgroup\,opts=ro}}} CHECK=beaker
+    services: docker
+  - rvm: 2.5.3
+    bundler_args: --without development release
+    env: BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_setfile=centos8-64{mount_folders={sysfscgroup={host_path=/sys/fs/cgroup\,container_path=/sys/fs/cgroup\,opts=ro}}} CHECK=beaker
+    services: docker
 branches:
   only:
   - master
@@ -84,7 +93,7 @@ notifications:
       - "chat.freenode.org#voxpupuli-notifications"
 deploy:
   provider: puppetforge
-  user: puppet
+  username: puppet
   password:
     secure: "Ojc0h2mbt9Y+eCyiKD+x1iYNONOP27Me63hjo9jo2v1bSs3aiM7djlcpz/sG+jRJ7JQoUyaGzHSn+gvxwWqdagFfFgDmipMKD0OXQinq7upRaG2hR+akKo0jllq9zLjJGBDoxurioKfOzPGlt2bX3UYY5KyeJ3AIM4dwCGVtSh4="
   on:
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
 end
--- a/metadata.json
+++ b/metadata.json
@@ -12,14 +12,16 @@
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
         "6",
-        "7"
+        "7",
+        "8"
       ]
     },
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
         "6",
-        "7"
+        "7",
+        "8"
       ]
     },
     {
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -185,6 +185,19 @@ describe 'rabbitmq' do
 
         context 'with service_manage set to true' do
           let(:params) { { admin_enable: true, management_ip_address: '1.1.1.1', service_manage: true } }
+          let(:python_package) do
+            if %w[Debian SUSE].include?(facts[:os]['family'])
+              'python'
+            elsif %w[FreeBSD OpenBSD].include?(facts[:os]['family'])
+              'python2'
+            elsif facts[:os]['family'] == 'RedHat'
+              if facts[:os]['release']['major'] == '8'
+                'python3'
+              else
+                'python'
+              end
+            end
+          end
 
           context 'with rabbitmqadmin_package set to blub' do
             let(:params) { { rabbitmqadmin_package: 'blub' } }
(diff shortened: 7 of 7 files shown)


nmaludy:
@ekohl Maybe i'm doing something wrong, but i did the following:
```
git clone https://github.com/voxpupuli/modulesync_config
export GIT_TAG=$(git tag --list | sort -V | tail -n1)
echo $GIT_TAG
# this printed 2.12.0
git checkout $GIT_TAG
bundle install
```

Get my module fork:
```
mkdir -p modules/nmaludy
cd modules/nmaludy
git clone git@github.com:nmaludy/puppet-rabbitmq.git
cd ../..
```

Run modulesync:
```
bundle exec msync update -n nmaludy -f puppet-rabbitmq --noop
```

Is this correct?

nmaludy:
@ekohl FYI looks like my modulesync changes are consistent with the `2.12.0` tag https://github.com/voxpupuli/modulesync_config/blob/2.12.0/config_defaults.yml#L7-L16

But they are different on the `master` branch: https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L7-L16

I can run modulesync again against master if you prefer?

wyardley:
@nmaludy yeah, the modulesync PRs should be in a separate commit, at least. I can run one against the repo and then you can rebase / squash your branch

wyardley:
Hrm, actually, yeah, since it is already synced against 2.12.0, should just leave it out.
If you need something that's more recent than that for this PR to work, we may need to have someone cut a newer modulesync tag.

nmaludy:
@wyardley i was using modulesync to add in the new beaker acceptance sets, i can just do it manually for now since these modulesync changes are not desired, standby for an updated commit

ekohl:
Looks like this module is actually on master. We've been making changes to simplify things using voxpupuli-acceptance but looks like we didn't do that via a real tag.

@nmaludy for now, I think you can use master to modulesync (or manual).

@bastelfreak should we do a 2.13.0 where we introduce voxpupuli-acceptance? Then later we can do the big breaking one wiht whitespace.

wyardley:
I can make that... just thought generally we're only supposed to sync off a tag.

wyardley:
@nmaludy can you rebase and then squash down into logical commits?

nmaludy:
Looks like acceptance tests are hanging when trying to start the `rabbitmq-server` service inside the docker container. This worked perfectly in a CentOS 8 Vagrant VM, so i'm guessing the failure/hang is a docker-ism. Going to look further now.

nmaludy:
So, this indeed failed for me when using Beaker on my local box. I tried to reproduce with a standard docker setup using Vagrant and this setup worked correctly:

```
docker_registry = ENV['DOCKER_REGISTRY']
docker_image = ENV['DOCKER_IMAGE'] ? ENV['DOCKER_IMAGE'] : 'centos:8'
docker_image_full = if docker_registry
                      "#{docker_registry}/#{docker_image}"
                    else
                      docker_image
                    end

dockerfile_path = ENV['DOCKERFILE_PATH'] ? ENV['DOCKERFILE_PATH'] : '.'
dockerfile = ENV['DOCKERFILE'] ? ENV['DOCKERFILE'] : 'Dockerfile'
cmd = ['/sbin/init']
volumes = ['/sys/fs/cgroup:/sys/fs/cgroup:ro']

Vagrant.configure('2') do |config|
  config.vm.synced_folder '.', '/ci/'

  config.vm.provider 'docker' do |d|
    if docker_image
      # pull container from registry
      d.image = docker_image_full
    else
      # build container from our Dockerfile
      d.build_dir = dockerfile_path
      d.dockerfile = dockerfile
    end
    # tell the container to stay running, even when /bin/bash exits (default command)
    d.create_args = ['-t', '--privileged']
    # mount volumes needed for systemd
    d.volumes = volumes
    # run as init
    d.cmd = cmd
  end

  config.vm.provision 'docker' do |d|
    d.pull_images docker_image_full
    d.run 'default'
  end
end
```

Running the commands:
```
vagrant up
docker ps -a
docker exec -it puppet-rabbitmq_default_1593453936 bash
curl -sSL https://raw.githubusercontent.com/nmaludy/puppet-install-shell/master/install_puppet_6_agent.sh | bash -s
yum -y install git
echo "puppetlabs/stdlib puppetlabs/apt puppet/archive garethr/erlang camptocamp/systemd puppetlabs/yumrepo_core" | xargs -n 1 /opt/puppetlabs/puppet/bin/puppet module install
cp -r /ci /etc/puppetlabs/code/environments/production/modules/rabbitmq
# copy code
cat << "EOF" > ./node.pp
class { 'rabbitmq': }
if $facts['os']['family'] == 'RedHat' {
  class { 'erlang': epel_enable => true}
  Class['erlang'] -> Class['rabbitmq']
}
EOF
/opt/puppetlabs/puppet/bin/puppet apply node.pp
```

Trying to figure out what is different from the container i created an the one that Beaker created? Maybe it isn't running in `privileged` mode, or doesn't have the right volumes mounted?

nmaludy:
Think i found the problem... my container mounts the volume `/sys/fs/cgroup:/sys/fs/cgroup:ro` but the beaker container does not.

I was able to find this by doing:

```shell
$ docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                  NAMES
81b21fb99204        89c3755d835c        "/sbin/init"        3 minutes ago       Up 2 minutes        0.0.0.0:4088->22/tcp   sweet_lehmann
888c591e1b22        centos:8            "/sbin/init"        20 minutes ago      Up 20 minutes                              puppet-rabbitmq_default_1593455879

$ docker inspect --format='{{.Mounts}}' 888c591e1b22
[{bind  /sys/fs/cgroup /sys/fs/cgroup  ro false rprivate} {bind  /home/nmaludy/storage/home/src/git/nmaludy/puppet-rabbitmq /ci   true rprivate}]

$ docker inspect --format='{{.Mounts}}' 81b21fb99204
[]
```

nmaludy:
I've figured out how to add this volume mount to the nodeset in beaker and modified the travis config to match. This test round _should_ get further than last time (at least the service should start)!

nmaludy:
Volume mounting worked in Travis, got passed the issue with systemd service restarts.

Ran into an issue with plugin idempotency and found a bug where we were only listing Explicitly enabled plugins. Since our test installs `rabbitmq_plugin { [ 'rabbitmq_federation_management', 'rabbitmq_federation' ]:` and `rabbitmq_federation` is a dependency of `rabbitmq_federation_management` it was being implicitly enabled. `rabbitmq-plugins` CLI doesn't list out implicitly enabled plugins unless you pass another flag. So, without this flag Puppet kept on thinking that the `rabbitmq_federation` plugin was not installed. I've fixed the bug by passing in the `-e` flag to list out implicitly enabled plugins within our `instances` function of the provider.

wyardley:
> I've fixed the bug by passing in the -e flag to list out implicitly enabled plugins within our instances function of the provider.

Thanks for digging into that. That one for sure should be a separate issue / PR, so that it's visible in the changelog etc, if you wouldn't mind cherry-picking that change separately (and maybe filing an issue for it).

Why wouldn't that issue come up with the integration tests for other versions?

nmaludy:
@wyardley looks like this was introduced in RabbitMQ 3.6.7 https://github.com/rabbitmq/rabbitmq-cli/commit/e0871712254f3310ac456e9cd3cf9c155d93ee87

I'll add a version switch in the code, it looks like we'll want to pass in `-E` for older version then _just_ `-e` for newer versions because `-e` displays all implicitly and explicitly enabled plugins where `-E` restricts it to _just_ explicitly enabled plugins (even if both are passed).

data/family/RedHat-8.yaml
```diff
@@ -0,0 +1,3 @@
+---
+rabbitmq::python_package: 'python3'
+rabbitmq::repos_ensure: true
```
wyardley, line 3: I see why you're doing this here, but I think probably we shouldn't use non-default params for one specific OS version for this.

If you want to switch the default to repos_ensure as a major (breaking) change, I'm onboard with that, though it may be a bit of an involved process.

wyardley:
> @wyardley looks like this was introduced in RabbitMQ 3.6.7 rabbitmq/rabbitmq-cli@e087171

Got it. yeah, the integration tests just run with the default version. I'm Ok with a couple of options:

1) Integration test with both `repos_ensure: true` and `repos_ensure: false` for all supported OS versions
2) Assuming there is a vendor package, keep `repos_ensure: false` (default) for this OS version
3) Make a separate (major / breaking) PR to change the default behavior back to `repos_ensure: true` for everyone.

#3 probably makes the most sense, but is also potentially a lot of work.

I take responsibility for the current behavior, introduced back in
https://github.com/voxpupuli/puppet-rabbitmq/pull/493
see my earlier comments here about this topic.

wyardley:
> I'll add a version switch in the code

If someone ever has the time / inclination, I think hitting the APIs vs. using the CLI would also be better. Someone started down that path, but don't think anything is happening in that regard these days.

nmaludy:
@wyardley I like your suggestions, let me get the build passing then we can figure out how we want to cherry pick and split it up. Does that sound ok? Just not sure how many more "gotchas" i'm going to run into :) 

Also, thanks so much for your feedback!!!

wyardley:
> @wyardley I like your suggestions, let me get the build passing then we can figure out how we want to cherry pick and split it up. Does that sound ok?

Absolutely!

spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
```diff
@@ -9,17 +9,84 @@
...
+  describe '#instances' do
+    it 'exists' do
+      expect(provider_class).to respond_to :instances
+    end
+
+    # rubocop:disable RSpec/MultipleExpectation
+    it 'retrieves instances' do
+      provider.class.expects(:plugin_list).returns(%w[foo bar])
+      instances = provider_class.instances
+      expect(instances.size).to eq(2)
```
ekohl, line 27: Is this expectation really that useful? If it's not 2, it's hard to debug because you don't get the actual values. Below you compare names so you at least know more.

README.md
```diff
@@ -54,6 +53,34 @@ You can view example usage in [REFERENCE](REFERENCE.md).
...
 See [REFERENCE](REFERENCE.md).
 
+## Limitations
+
+If running CentOS/RHEL, ensure the `epel` repo, or another repo containing a
+suitable Erlang version, is present. On Debian systems, puppetlabs/apt
+(>=2.0.0 < 5.0.0) is a soft dependency.
+
+To have a suitable erlang version installed on RedHat and Debian systems,
+you have to install another puppet module from http://forge.puppetlabs.com/garethr/erlang with:
```
ekohl, line 63: In general I prefer https links
```suggestion
you have to install another puppet module from https://forge.puppetlabs.com/garethr/erlang with:
```

README.md
```diff
@@ -54,6 +53,34 @@ You can view example usage in [REFERENCE](REFERENCE.md).
...
+
+    puppet module install garethr-erlang
+
+This module handles the packages for erlang.
+To use the module, add the following snippet to your site.pp or an appropriate profile class:
+
+For RedHat systems:
+
+    include 'erlang'
+    class { 'erlang': epel_enable => true}
```
ekohl, line 73: You can get syntax highlighting
```suggestion
```puppet
include 'erlang'
class { 'erlang': epel_enable => true}
```

(Note Github doesn't allow me to suggest the last closing backticks)

The same goes below for the Debian suggestion.

.sync.yml
```diff
@@ -8,6 +8,8 @@
     - set: ubuntu1804-64
     - set: centos6-64
     - set: centos7-64
+    # RabbitMQ fails to start via systemd unless /sys/fs/cgroup is mounted through from obase host
```
ekohl, line 11: obase?

spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
nmaludy: Agreed and fixed!

nmaludy:
Doing a lot more digging and was having issues trying to do consistent realtime version checks when listing plugins. Checking with older versions of RabbitMQ and the oldest one i could get my hands on was 3.5.1 and `rabbitmq-plugin` in that version supports the `-e` option for listing implicitly enabled plugins. I'm not sure how this code has been working so long without running into this problem.

Fighting with this all day and i think the best option is to change `-E` to `-e` for listing plugins. Proof is in the pudding, switching from `-E` to `-e` for listing plugins allowed all acceptance tests to pass!

nmaludy:
@wyardley ok, i'm ready to talk about splitting this up into different PRs now!

wyardley:
> Proof is in the pudding, switching from -E to -e for listing plugins allowed all acceptance tests to pass!

if they pass across all supported versions, I _think_ we should be good. You could also run the beaker set for the default RMQ version and manually validate that it works there. As long as it works with our min current supported version, should be Ok.

As far as breaking up, here's what I propose:
* issue / PR for `rabbitmq_plugin` provider arguments (we could maybe cut a release for this before the breaking changes too)
* PR flagged as "backwards-incompatible" / breaking to change `repos_ensure` default to `true`, with any docs updates needed in README or generated docs (watch out for puppet-strings, though, as you don't want to update the generated docs, just the sources) and any adjustments needed to make current matrix pass for this.
* new enhancement PR (or repurpose this one) adding CentOS / RHEL 8 support

After all that, I can make a release PR with the updated docs, changelog, etc., targeting a major version.

nmaludy:
`rabbitmq_plugin` PR is here: https://github.com/voxpupuli/puppet-rabbitmq/pull/844

nmaludy:
`repos_ensure => true` PR is here: https://github.com/voxpupuli/puppet-rabbitmq/pull/845

nmaludy:
@wyardley I've squashed this PR down and removed the extra changes caused by `repos_ensure`. So, this is currently blocked by:
- #844 
- #845 

Once we merge those two PRs, this is ready to go after a quick rebase.

PR #845, opened 2020-07-01
Change repos_ensure default from false to true

nmaludy:
As part of the work i was doing to enable CentOS/RHEL 8 support in #842 i ran into an issue where EPEL no longer ships RabbitMQ packages. In that PR i had made some hacky fixes to enable the RabbitMQ repos, via the `repos_ensure` parameter, by default on that OS. After some discussions it was decided that we should instead, simply change the default value of `repos_ensure` to `true` on all OSes.

**Note: This is a breaking change!**

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -29,7 +29,7 @@ rabbitmq::package_gpg_key: ~
 rabbitmq::package_name: 'rabbitmq'
 rabbitmq::package_source: ~
 rabbitmq::package_provider: ~
-rabbitmq::repos_ensure: false
+rabbitmq::repos_ensure: true
 rabbitmq::manage_python: true
 rabbitmq::python_package: 'python'
 rabbitmq::rabbitmq_user: 'rabbitmq'
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -202,7 +202,7 @@
 #   Name of the package required by rabbitmqadmin.
 # @param repos_ensure
 #   Ensure that a repo with the official (and newer) RabbitMQ package is configured, along with its signing key.
-#   Defaults to false (use system packages). This does not ensure that soft dependencies (like EPEL on RHEL systems) are present.
+#   Defaults to true (use RabbitMQ repos). This does not ensure that soft dependencies (like EPEL on RHEL systems) are present.
 #   It also does not solve the erlang dependency.  See https://www.rabbitmq.com/which-erlang.html for a good breakdown of the
 #   different ways of handling the erlang deps.  See also https://github.com/voxpupuli/puppet-rabbitmq/issues/788
 # @param service_ensure
@@ -329,7 +329,7 @@ class rabbitmq(
   Variant[String, Array] $package_name                                                             = 'rabbitmq',
   Optional[String] $package_source                                                                 = undef,
   Optional[String] $package_provider                                                               = undef,
-  Boolean $repos_ensure                                                                            = false,
+  Boolean $repos_ensure                                                                            = true,
   Boolean $manage_python                                                                           = true,
   String $python_package                                                                           = 'python',
   String $rabbitmq_user                                                                            = 'rabbitmq',
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -27,21 +27,6 @@ describe 'rabbitmq' do
       end
 
       context 'with default params' do
-        it { is_expected.not_to contain_class('rabbitmq::repo::apt') }
-        it { is_expected.not_to contain_apt__source('rabbitmq') }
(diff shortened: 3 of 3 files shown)


nmaludy:
@wyardley Looking at the build failures here and we may have opened a can of worms.

Debian: https://travis-ci.org/github/voxpupuli/puppet-rabbitmq/jobs/703736212#L1799-L1837
CentOS: https://travis-ci.org/github/voxpupuli/puppet-rabbitmq/jobs/703736222#L1701-L1707

In both cases, the repo enabled by the `garethr/erlang` module has such an old version of Erlang, that the modern version of `rabbitmq-server` can't be used.

I think if we want this to be the default, we'll have to craft our own Erlang module that has flags to support enabling the PackageCloud/Bintray repos.

Thoughts?

wyardley:
@nmaludy yeah, tried to warn you that this was a big can of worms 😆 

A new erlang module would be ideal, and if we can get that stood up (could either just start a brand new one or fork garethr). Additionally, we could consider a bool (defaulting to false) to use the RabbitMQ provided erlang packages from their repos? This won't work for everyone's use case, though, and even if we had such an option, if it defaults to `false`, wouldn't be ideal to use it in our own integration tests.

Since we document that the erlang dep has to be satisfied, you could use some other method of bootstrapping erlang in the acceptance test helpers, but that could get kind of hairy across the different things we support.

nmaludy:
@wyardley I've started on my own implementation of an erlang module. My plan is to donate this to Vox once i complete it, hopefully in the next few days.

ekohl:
> In both cases, the repo enabled by the `garethr/erlang` module has such an old version of Erlang, that the modern version of `rabbitmq-server` can't be used.

Given garethr/erlang used to be approved, would it make sense to try and transfer it to VP?

wyardley:
I feel like I might have tried to ask him about it ages ago, but could be wrong.

nmaludy:
I don't have bandwidth to look at this right now.

But from some initial debugging, it looks like RabbitMQ is hanging when starting up in this configuration.

- `apt-get install rabbitmq-server` is failing:
```shell
  Setting up erlang-examples (1:21.3-1) ...                                                                                                                                       
  Setting up logrotate (3.8.7-1+b1) ...                                                                                                                                           
  Setting up socat (1.7.2.4-2) ...                                                                                                                                                
  Setting up rabbitmq-server (3.8.5-1) ...                                                                                                                                        
  Adding group `rabbitmq' (GID 117) ...                                                                                                                                           
  Done.                                                                                                                                                                           
  Adding system user `rabbitmq' (UID 110) ...                                                                                                                                     
  Adding new user `rabbitmq' (UID 110) with group `rabbitmq' ...                                                                                                                  
  Not creating home directory `/var/lib/rabbitmq'.                                                                                                                                
  Job for rabbitmq-server.service failed. See 'systemctl status rabbitmq-server.service' and 'journalctl -xn' for details.                                                        
  invoke-rc.d: initscript rabbitmq-server, action "start" failed.                                                                                                                 
  dpkg: error processing package rabbitmq-server (--configure):                                                                                                                   
   subprocess installed post-installation script returned error exit status 1                                                                                                     
  Processing triggers for systemd (215-17+deb8u13) ...                                                                                                                            
  Errors were encountered while processing:                                                                                                                                       
   rabbitmq-server                                                                                                                                                                
  E: Sub-process /usr/bin/dpkg returned an error code (1)   
```
- Trying to start rabbitmq-server manually hangs

Some next steps are:
- Attempt to use the "out of box" Erlang repos to narrow down if it's an Erlang problem or a RabbitMQ problem
- Identify what is causing the rabbitmq-server install to error
- Identify what is causing rabbitmq-server service to hang

nmaludy:
@wyardley this PR is still a struggle for me... are you OK with us getting RHEL8 support in there with the one override for `repos_ensure` on RHEL8, then i can come back and revisit this making it the default on all platforms at a later time?

wyardley:
@nmaludy Anything I can help with? and is the puppet-erlang module released yet? I still don't see a tag cut there.

My thought is that once that's released, it should be relatively easy to get this working and release it as a major / breaking release..

I don't think it's ideal to have RHEL8 support a different set of configurations from the default (though you could override that setting in your local config, presumably)

nmaludy:
@wyardley the integration tests are failing because rabbitmq fails to start on the hosts with the new repo, haven't been able to figure out why it's failing to start and/or how to fix it.

wyardley:
Can we get a tag cut of the vox erlang mofule you wrote?

Can you try adding your erlang stuff as needed replacing the garethr stuff here
https://github.com/voxpupuli/puppet-rabbitmq/blob/b05b6b24fc09591cf033ea841d844f7f78653c59/spec/spec_helper_acceptance.rb#L8
And then take it out of the spec manifests for now?

Issue #846, opened 2020-07-09
Cannot set delivery-limit policy


philomory:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.5.18
- Ruby: 2.4.4
- Distribution: CentOS 7
- Module version: 10.0.0

## How to reproduce (e.g Puppet code you use)
```
rabbitmq_policy { 'example1@/':
  pattern    => "^example1$",
  priority   => 50,
  applyto    => 'queues',
  definition => {
    'delivery-limit' => 3,
  },
}

rabbitmq_policy { 'example2@/':
  pattern    => "^example2$",
  priority   => 50,
  applyto    => 'queues',
  definition => {
    'delivery-limit' => '3',
  },
}
```

## What are you seeing
For the integer delivery-limit: 
```
Error: Failed to apply catalog: Parameter definition failed on Rabbitmq_policy[example1]: Invalid definition, value 3 is not a string 
```

For the string delivery-limit:
```
Error:
Validation failed

<<"3">> is not a valid delivery limit
```

## What behaviour did you expect instead
I should be able to set a delivery limit via policy with the RabbitMQ module.

## Any additional information you'd like to impart
I can see in the code that the module expects all policy values to be Strings as they come in from the manifest, and the Type munges a select whitelist of values to integers; obviously `delivery-limit` could be added to this whitelist, but, I wonder if it wouldn't make more sense to instead stop trying to enforce stringiness in the first place - example 1 should be allowed to 'just work', without the module needing to be _aware_ that the `delivery-limit` policy takes an integer.

PR #847, opened 2020-07-09
Allow delivery-limit policy to be set

philomory:
This will allow `delivery-limit` to be set via policy.

Fixes: #846

### DIFF
--- a/lib/puppet/type/rabbitmq_policy.rb
+++ b/lib/puppet/type/rabbitmq_policy.rb
@@ -128,10 +128,16 @@ DESC
         raise ArgumentError, "Invalid shards-per-node value '#{shards_per_node_val}'"
       end
     end
-    if definition.key? 'ha-sync-batch-size' # rubocop:disable Style/GuardClause
+    if definition.key? 'ha-sync-batch-size'
       ha_sync_batch_size_val = definition['ha-sync-batch-size']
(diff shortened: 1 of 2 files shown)


philomory:
I think the "must be a string" validation probably dates back to when communication between Puppet and Ruby didn't actually support any data types other than String, Array and Hash; so checking "is it a String?" was really just a way to ensure it _wasn't_ a Hash or Array. But, at this point, that chunk of code really only accomplishes four things:

1. You can't accidentally feed an Array or Hash to a policy that accepts only a single value (there are less invasive ways to accomplish this),
2. New policies that accept anything other than a string require updates to the module to support them
3. Attempts to pass e.g. an Integer from the manifest for a policy that takes an Integer fail; you counter-intuitively must pass a String an allow the module to convert it to an integer for you.
4. Because the module is converting everything from Strings, it has to manually check that `value.to_i.to_s == value` because `"hello".to_i == 0`

wyardley:
@philomory if you want to propose a broader PR to do less validation, happy to take a look; I'm not working with Puppet or with RabbitMQ these days, so we may need to get some broader feedback as well.

philomory:
@wyardley I'll look into doing that, although at the moment the current PR addresses my immediate issues.

PR #849, opened 2020-07-11
Remove facter rabbitmq_nodename error message

mbaldessari:
Currently when facter checks for rabbitmq_nodename and rabbitmq
is not available due to a number of reasons. We get the following
facter error displayed:
Error: Facter: error while resolving custom fact "rabbitmq_nodename": undefined method `[]' for nil:NilClass

Since facter errors are non fatal anyway let's be a little friendlier.

Before this change:
$ puppet apply foo.pp
Error: Facter: error while resolving custom fact "rabbitmq_nodename": undefined method `[]' for nil:NilClass
Notice: Compiled catalog for messaging-0.redhat.local in environment production in 0.02 seconds
Notice: foo
Notice: /Stage[main]/Main/Notify[foo]/message: defined 'message' as 'foo'

After this change:
$ puppet apply foo.pp
Notice: Compiled catalog for messaging-0.redhat.local in environment production in 0.02 seconds
Notice: foo
Notice: /Stage[main]/Main/Notify[foo]/message: defined 'message' as 'foo'

### DIFF
--- a/lib/facter/rabbitmq_nodename.rb
+++ b/lib/facter/rabbitmq_nodename.rb
@@ -2,7 +2,11 @@ Facter.add(:rabbitmq_nodename) do
   setcode do
     if Facter::Util::Resolution.which('rabbitmqctl')
       rabbitmq_nodename = Facter::Core::Execution.execute('rabbitmqctl status 2>&1')
-      %r{^Status of node '?([\w\.\-]+@[\w\.\-]+)'?}.match(rabbitmq_nodename)[1]
+      begin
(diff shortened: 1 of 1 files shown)


mbaldessari:
I think returning nil when rabbitmqctl status without throwing an error when it did not return anything useful is a sensible thing to do. 
Background for this is that I get pinged quite often in the openstack world because this error leads people to believe that rabbitmq is having issues, so it seemed worthwhile to not error out in the perfectly normal situation where status is not up and running.
How about I add an error only when running in debug mode, would that be better?

wyardley:
That seems like a good way to do it!

PR #853, opened 2020-07-25
modulesync 3.0.0 & puppet-lint updates

bastelfreak:
modulesync 3.0.0


bastelfreak:
i messed this modulesync run up a bit :(
feel free to recreate it if you like, otherwise I will work on it in the next days (but there are some more repos that require work..)

Issue #836, opened 2020-05-30
Not idempotent again on RHEL based platforms since move to systemd module


tobias-urdin:
This commit [1] some years ago fixed so that this module on RHEL based platforms was idempotent again. Since the move to the systemd module the same issue has occured again.

Need to implement a way to set selinux_ignore_defaults in systemd::service_limits (and thus in systemd::dropin_file) and add that usage in this module.

[1] https://github.com/voxpupuli/puppet-rabbitmq/commit/0227f762070ffbbea3c28d6a60174de98fa4cc1c

tobias-urdin:
On CentOS 8:
2020-05-29 10:52:28 +0000 /Stage[main]/Rabbitmq::Config/Systemd::Service_limits[rabbitmq-server.service]/Systemd::Dropin_file[rabbitmq-server.service-90-limits.conf]/File[/etc/systemd/system/rabbitmq-server.service.d]/seltype (notice): seltype changed 'rabbitmq_unit_file_t' to 'systemd_unit_file_t'

tobias-urdin:
Adding support for selinux_ignore_defaults to systemd module here [1].

[1] https://github.com/camptocamp/puppet-systemd/pull/151

tobias-urdin:
We can then edit this line [1] in this module and add the selinux_ignore_defaults parameter for the systemd::service_limits usage when handling RHEL based operating systems.

[1] https://github.com/voxpupuli/puppet-rabbitmq/blob/master/manifests/config.pp#L234

adeleglise:
Is it possible that this commit added a breaking change ?

I can observe this error when trying to use the module v11.0.0 on a Ubuntu 20.04 box:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Systemd::Service_limits[rabbitmq-server.service]: has no parameter named 'selinux_ignore_defaults' (file: /etc/puppetlabs/code/environments/staging/modules/rabbitmq/manifests/config.pp, line: 236) on node
```

wyardley:
@adeleglise can you see if you have a version of the systemd module that’s in bounds of the lower bound that the module requests? Whether this new parameter being required should have been a breaking change I can’t say (I would lean towards no), but feel free to open a new issue if you see this error even with the updated version of the systemd module.

[...]

PR #856, opened 2020-08-23
Ignore SELinux defaults for systemd on RHEL based

tobias-urdin:
It makes sure it's idempotent on RHEL based, needs an increase in systemd dependency to a minimum
versions of 2.10.0 to include the changes there.

Fixes: #836

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -231,9 +231,10 @@ class rabbitmq::config {
 
   if $facts['systemd'] { # systemd fact provided by systemd module
     systemd::service_limits { "${service_name}.service":
-      limits          => { 'LimitNOFILE' => $file_limit },
+      selinux_ignore_defaults => ($facts['os']['family'] == 'RedHat'),
+      limits                  => { 'LimitNOFILE' => $file_limit },
       # The service will be notified when config changes
-      restart_service => false,
+      restart_service         => false,
     }
   }
 
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -148,8 +148,11 @@ describe 'rabbitmq' do
           end
 
           if facts[:systemd]
+            selinux_ignore_defaults = facts[:os]['family'] == 'RedHat'
+
             it do
               is_expected.to contain_systemd__service_limits("#{name}.service").
+                with_selinux_ignore_defaults(selinux_ignore_defaults).
                 with_limits('LimitNOFILE' => value).
                 with_restart_service(false)
             end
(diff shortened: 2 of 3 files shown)


wyardley:
@tobias-urdin can you rebase / squash, and resolve the rubocop failure?

Are there any security implications of this change?

tobias-urdin:
There is no security implications, this simply makes sure the systemd module doesn't overwrite the SELinux contexts the unit file that the installed package drops.

manifests/config.pp
```diff
@@ -230,10 +230,16 @@
   }
 
   if $facts['systemd'] { # systemd fact provided by systemd module
+    $selinux_ignore_defaults_real = $::os['family'] ? {
+      'RedHat' => true,
+      default  => false,
+    }
```
ekohl, line 236: It's better to use `$facts`. I think you can also simplify it further:
```suggestion
    $selinux_ignore_defaults_real = $facts['os']['family'] == 'RedHat'
```

spec/classes/rabbitmq_spec.rb
```diff
@@ -148,8 +148,15 @@
           end
 
           if facts[:systemd]
+            selinux_ignore_defaults = if facts[:os]['family'] == 'RedHat'
+              true
+            else
+              false
+            end
```
ekohl, line 155: This can be simplified:
```suggestion
            selinux_ignore_defaults = facts[:os]['family'] == 'RedHat'
```

ekohl (changes requested):
So I guess this is due to https://tickets.puppetlabs.com/browse/PUP-10548. It would be good to refer to that if that's the case.

tobias-urdin:
Fixed, the bug you linked is not really related. The issue occurs because the RabbitMQ packages built by the CentOS Messaging SIG drops the systemd unit file with the rabbitmq_unit_t context, the systemd modules then tries to overwrite that to systemd_unit_t.

ekohl:
So the RabbitMQ packages don't even provide the content so a restorecon would also revert it back to `systemd_unit_t`? That sounds like bad packaging. It would be good to at least mention that in the commit message, but ideally also some URL to a CentOS bug that describes it.

Other than that the PR looks good to me.

tobias-urdin:
It's not really a package bug, it's more a regression in puppet-rabbitmq that was introduced when moving to the systemd module. See the commit message on the fix back in 2015 [1].

[1] https://github.com/voxpupuli/puppet-rabbitmq/commit/0227f762070ffbbea3c28d6a60174de98fa4cc1c

tobias-urdin:
Anything missing here or are we good?

igalic:
an approving review (from somebody competent in SELinux / RHEL)

ekohl (approved):

PR #865, opened 2020-11-25
Add new parameter autoconvert for rabbitmq_parameter

joec4i:
Add a new parameter `autoconvert` for `rabbitmq_parameter` to allow users to disable auto type conversion from numeric strings to integers if needed.

We have several `x-content-hash` exchanges which use a numeric string as the routing key. Due to the `munge_value` [implementation](https://github.com/voxpupuli/puppet-rabbitmq/blob/2bcfb0e/lib/puppet/type/rabbitmq_parameter.rb#L82), the routing key would be converted to an integer. 

As a result, a valid payload to create a shovel as below
```json
{
  "component": "shovel",
  "vhost": "foo",
  "name": "test",
  "value": {
    "src-uri": "amqps://user:[redacted]@rabbit.service.consul:5671/foo",
    "src-exchange-key": "5",
    "dest-uri": "amqps://user:[redacted]@rabbit.service.consul:5671/bar",
    "dest-exchange-key": "5",
    "ack-mode": "on-confirm",
    "src-protocol": "amqp091",
    "dest-protocol": "amqp091",
    "src-exchange": "indexer_document_requests",
    "src-delete-after": "never",
    "dest-exchange": "indexer_document_requests",
    "dest-add-forward-headers": false
  }
}
```
will be converted to 
```json
{
  "component": "shovel",
  "vhost": "foo",
  "name": "test",
  "value": {
    "src-uri": "amqps://user:[redacted]@rabbit.service.consul:5671/foo",
    "src-exchange-key": 5,
    "dest-uri": "amqps://user:[redacted]@rabbit.service.consul:5671/bar",
    "dest-exchange-key": 5,
    "ack-mode": "on-confirm",
    "src-protocol": "amqp091",
    "dest-protocol": "amqp091",
    "src-exchange": "indexer_document_requests",
    "src-delete-after": "never",
    "dest-exchange": "indexer_document_requests",
    "dest-add-forward-headers": false
  }
}
```
Because RabbitMQ doesn't accept exchange keys to be integers, it'll throw out validation errors:
```
Error:
Validation failed
dest-exchange-key should be binary, actually was 5
src-exchange-key should be binary, actually was 5
```

This PR would allow the auto type conversion to be disabled and therefore numeric strings to be included in the RabbitMQ parameter payload if needed. The users would be responsible to ensure the right date types are used if `autoconvert` is set to `false`.

### DIFF
--- a/lib/puppet/type/rabbitmq_parameter.rb
+++ b/lib/puppet/type/rabbitmq_parameter.rb
@@ -19,6 +19,18 @@ Type for managing rabbitmq parameters
          'expires' => '360000',
      },
    }
+   rabbitmq_parameter { 'documentumShovelNoMunging@/':
+     component_name => '',
+     value          => {
+         'src-uri'    => 'amqp://',
+         'src-exchange'  => 'my-exchange',
+         'src-exchange-key' => '6',
+         'src-queue'  => 'my-queue',
+         'dest-uri'   => 'amqp://remote-server',
+         'dest-exchange' => 'another-exchange',
+     },
+     autoconvert   => false,
+   }
 DESC
 
   ensurable do
@@ -50,6 +62,12 @@ DESC
     end
   end
 
+  newparam(:autoconvert) do
+    desc 'whether numeric strings from `value` should be converted to int automatically'
+    newvalues(:true, :false)
+    defaultto(:true)
+  end
+
   newproperty(:value) do
     desc 'A hash of values to use with the component name you are setting'
     validate do |value|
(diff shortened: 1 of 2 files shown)


wyardley (commented):
Looks reasonable to me, provided it’s used carefully. Would it be good to make the parameter refer to autoconvert vs munge?

joec4i:
> Looks reasonable to me, provided it’s used carefully. Would it be good to make the parameter refer to autoconvert vs munge?

Good point @wyardley .  I've renamed the parameter to `autoconvert` which defaults to `true`. Please have another look. Thanks!

wyardley (approved):
LGTM. Would you mind squashing down to one commit in your branch?

joec4i:
Thanks, Will! It's in one commit now.

wyardley:
Cool. Going to leave this open for a day or two in case anyone else wants to review / comment. I’ll try to cut a new release after merging this.

nmaludy (approved):
LGTM

PR #867, opened 2020-11-30
Ensure :autoconvert is initialized before :value for rabbitmq_parameter

joec4i:
## Pull Request (PR) description

Make sure `:autoconvert` is set before `:value` for `rabbitmq_parameter` since `:value` depends on `:autoconvert`.

This was done by overriding `set_parameters` for the type. It's hacky but there doesn't seem to be a better way to put put a parameter before a property since puppet lets `properties` take precedence over `parameters` -  https://github.com/puppetlabs/puppet/blob/7ef0811/lib/puppet/type.rb#L118-L124 . 

Note: my first attempt was to change `:autoconvert` to a property but that doesn't really fit. Because we don't store its value anywhere, having `:autoconvert` as a property will lead puppet to believe that there is always a change to the `:autoconvert`
```
  Notice: /Stage[main]/Main/Rabbitmq_parameter[documentumFed@fedhost]/autoconvert: defined 'autoconvert' as 'true'
```

My previous PR #865 has a test case where `parameter[:autoconvert] = false` is set explicitly before `value` , so it passes as expected. However, it fails to check the case where a `rabbitmq_parameter` is defined with a hash. e.g.

```puppet
rabbitmq_parameter { 'documentumShovel@/':
  component_name => 'shovel',
  autoconvert    => false,
  value          => { 'myparameter' => '1800000' },
}
```
Both cases are supported with this PR.

Please review. Thanks!

### DIFF
--- a/lib/puppet/type/rabbitmq_parameter.rb
+++ b/lib/puppet/type/rabbitmq_parameter.rb
@@ -82,6 +82,12 @@ DESC
     [self[:name].split('@')[1]]
   end
 
+  def set_parameters(hash) # rubocop:disable Style/AccessorMethodName
+    # Hack to ensure :autoconvert is initialized before :value
+    self[:autoconvert] = hash[:autoconvert] if hash.key?(:autoconvert)
+    super
+  end
+
   def validate_component_name(value)
     raise ArgumentError, 'component_name must be defined' if value.empty?
   end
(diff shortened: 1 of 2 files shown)


wyardley (approved):

wyardley:
Thanks @joec4i 

Any interest in doing any other work for this module?

joec4i:
> Thanks @joec4i
> 
> Any interest in doing any other work for this module?

Thanks @wyardley for the prompt review. Also, as a user of the module, thank you for your work maintaining the module:) I'd love to help where I can although I'm not sure if I'd have enough time. Feel free to ping me if there's anything I can help with though. Cheers!

PR #866, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0

PR #868, opened 2020-12-01
Release 10.3.0

wyardley:
Release 10.3.0:

* Add new parameter autoconvert for rabbitmq_parameter (@joec4i) (#865)
* Ensure :autoconvert is initialized before :value for rabbitmq_parameter (@joec4i) (#867)
* modulesync 3.1.0 (@bastelfreak) (#864)
* modulesync 4.0.0 (@bastelfreak) (#866)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,19 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
-## [v10.2.0](https://github.com/voxpupuli/puppet-rabbitmq/tree/v10.2.0) (2020-10-12)
+## [v10.3.0](https://github.com/voxpupuli/puppet-rabbitmq/tree/v10.3.0) (2020-12-01)
+
+[Full Changelog](https://github.com/voxpupuli/puppet-rabbitmq/compare/v10.2.0...v10.3.0)
+
+**Implemented enhancements:**
+
+- Add new parameter autoconvert for rabbitmq\_parameter [\#865](https://github.com/voxpupuli/puppet-rabbitmq/pull/865) ([joec4i](https://github.com/joec4i))
+
+**Fixed bugs:**
+
+- Ensure :autoconvert is initialized before :value for rabbitmq\_parameter [\#867](https://github.com/voxpupuli/puppet-rabbitmq/pull/867) ([joec4i](https://github.com/joec4i))
+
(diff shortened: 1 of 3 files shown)


bastelfreak:
please also update the changelog

ekohl (commented):
I'm not a fan of the `chore: ` prefix in commit messages.

wyardley:
Oops, yeah, forgot the changelog with all the fuss getting the strings to update (still doesn’t work with latest puppet-strings)

wyardley:
> I'm not a fan of the chore: prefix in commit messages

If Vox ever decides on a convention, I’ll be happy to follow it.

ekohl:
> If Vox ever decides on a convention, I’ll be happy to follow it. For now, there’s no convention, so using the conventional commits / angular style I’m used to seems reasonable enough.

What I've seen the most is to write only `Release x.y.z`. IMHO that's a convention.

wyardley:
> What I've seen the most is to write only Release x.y.z. IMHO that's a convention.

It doesn't matter too much to me, but what I meant was, an overall convention like https://www.conventionalcommits.org/en/v1.0.0-beta.2/ or any competing standard (whether it's Angular style or not) to get the meaning / commit type from _any_ commit (not just release commits). While it would be potentially hard to get people to update their commits, there are tools to enforce conventions in CI, and regardless of how you feel about it aesthetically, it really simplifies versioning, changelog generation, etc. much more so than the existing approach with labels.

So, Vox may have a de facto convention that applies to release PRs / commits only, but AFAICT but it's not mentioned in https://voxpupuli.org/docs/releasing_version/, and it's not an overarching convention in terns of how commit messages should be written more generally.

ekohl (approved):

ekohl:
> It doesn't matter too much to me, but what I meant was, an overall convention like https://www.conventionalcommits.org/en/v1.0.0-beta.2/ or any competing standard (whether it's Angular style or not) to get the meaning / commit type from _any_ commit (not just release commits). While it would be potentially hard to get people to update their commits, there are tools to enforce conventions in CI, and regardless of how you feel about it aesthetically, it really simplifies versioning, changelog generation, etc. much more so than the existing approach with labels.

I strongly dislike that style. Every time I see it, it annoys me. I'd oppose adopting that for VP.

Today we already have a different method of generating the changelog and versioning. While they're not the best, they at least make it easy to adjust later on. If it's in git, that means it must be reviewed as part of the PR. That's easy to forget. It also increases the burden to contribute.

PR #869, opened 2020-12-23
Add optional variables to support SSL CRL check configuration

dimonzozo:
Add support for SSL CRL (Certificate Revocation List) check options.

This option is not documented in official documentation but RabbitMQ supports these options because they are implemented in underlying [OTP SSL module](https://erlang.org/doc/man/ssl.html#type-crl_cache_opts).

There are 3 main use cases:

* Enable CRL check in `best_effort` mode (if certificate revocation status cannot be determined it will be accepted as valid) and without any cache

In this case, result configuration should look like this:

```
[
  {rabbit, [
     {ssl_listeners, [5671]},
     {ssl_options, [{cacertfile,"/path/to/testca/cacert.pem"},
                    {certfile,"/path/to/server/cert.pem"},
                    {keyfile,"/path/to/server/key.pem"},
                    {verify,verify_peer},
                    {fail_if_no_peer_cert,false}]},
                    {crl_check, best_effort}
   ]}
].
```

* Enable CRL check in any of 3 modes and use internal CRL cache

```
[
  {rabbit, [
     {ssl_listeners, [5671]},
     {ssl_options, [{cacertfile,"/path/to/testca/cacert.pem"},
                    {certfile,"/path/to/server/cert.pem"},
                    {keyfile,"/path/to/server/key.pem"},
                    {verify,verify_peer},
                    {fail_if_no_peer_cert,false}]},
                    {crl_check, true},
                    {crl_cache, {ssl_crl_cache, {internal, [{http, 5000}]}}}
   ]}
].
```

* Do not perform any HTTP request and just use local directory to perform CRL check

_See [ssl_crl_hash_dir](https://erlang.org/doc/man/ssl.html#type-crl_cache_opts) implementation docs for more details._

```
[
  {rabbit, [
     {ssl_listeners, [5671]},
     {ssl_options, [{cacertfile,"/path/to/testca/cacert.pem"},
                    {certfile,"/path/to/server/cert.pem"},
                    {keyfile,"/path/to/server/key.pem"},
                    {verify,verify_peer},
                    {fail_if_no_peer_cert,false}]},
                    {crl_check, true},
                    {crl_cache, {ssl_crl_hash_dir, {internal, [{dir, "/path/to/crl/cache/"}]}}}
   ]}
].
```

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -65,9 +65,6 @@ rabbitmq::ssl_reuse_sessions: true
 rabbitmq::ssl_honor_cipher_order: true
 rabbitmq::ssl_dhfile: ~
 rabbitmq::ssl_ciphers: []
-rabbitmq::ssl_crl_check: 'false'
-rabbitmq::ssl_crl_cache_hash_dir: ~
-rabbitmq::ssl_crl_cache_http_timeout: ~
 rabbitmq::stomp_ensure: false
 rabbitmq::ldap_auth: false
 rabbitmq::ldap_server: 'ldap'
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -65,6 +65,9 @@ rabbitmq::ssl_reuse_sessions: true
 rabbitmq::ssl_honor_cipher_order: true
 rabbitmq::ssl_dhfile: ~
 rabbitmq::ssl_ciphers: []
+rabbitmq::ssl_crl_check: 'false'
+rabbitmq::ssl_crl_cache_hash_dir: ~
+rabbitmq::ssl_crl_cache_http_timeout: ~
 rabbitmq::stomp_ensure: false
 rabbitmq::ldap_auth: false
 rabbitmq::ldap_server: 'ldap'
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -378,7 +378,7 @@ class rabbitmq (
   Optional[Stdlib::Absolutepath] $ssl_dhfile                                                       = undef,
   Array $ssl_ciphers                                                                               = [],
   Enum['true','false','peer','best_effort'] $ssl_crl_check                                         = 'false',
-  Optional[String] $ssl_crl_cache_hash_dir                                                         = undef,
+  Optional[Stdlib::Absolutepath] $ssl_crl_cache_hash_dir                                                         = undef,
   Optional[Integer] $ssl_crl_cache_http_timeout                                                    = undef,
   Boolean $stomp_ensure                                                                            = false,
   Boolean $ldap_auth                                                                               = false,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -269,6 +269,15 @@
 #   Functionality can be tested with cipherscan or similar tool: https://github.com/mozilla/cipherscan
 #   * Erlang style: `['ecdhe_rsa,aes_256_cbc,sha', 'dhe_rsa,aes_256_cbc,sha']`
 #   * OpenSSL style: `['ECDHE-RSA-AES256-SHA', 'DHE-RSA-AES256-SHA']`
+# @param ssl_crl_check
+#   Perform CRL (Certificate Revocation List) verification
+#   Please see the [Erlang SSL](https://erlang.org/doc/man/ssl.html#type-crl_check) module documentation for more information.
+# @param ssl_crl_cache_hash_dir
+#   This setting makes use of a directory where CRLs are stored in files named by the hash of the issuer name.
+#   Please see the [Erlang SSL](https://erlang.org/doc/man/ssl.html#type-crl_cache_opts) module documentation for more information.
+# @param ssl_crl_cache_http_timeout
+#   This setting enables use of internal CRLs cache and sets HTTP timeout interval on fetching CRLs from distributino URLs defined inside certificate.
+#   Please see the [Erlang SSL](https://erlang.org/doc/man/ssl.html#type-crl_cache_opts) module documentation for more information.
 # @param stomp_port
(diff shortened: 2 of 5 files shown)


REFERENCE.md
```diff
@@ -834,6 +834,33 @@ Functionality can be tested with cipherscan or similar tool: https://github.com/
 
 Default value: []
 
+##### `ssl_crl_check`
```
wyardley, line 837: it's Ok to add these here, but they _should_ get regenerated when we build the strings before release.

spec/classes/rabbitmq_spec.rb
```diff
@@ -1161,6 +1161,101 @@
...
+            ssl_port: 3141,
+            ssl_cacert: '/path/to/cacert',
+            ssl_cert: '/path/to/cert',
+            ssl_key: '/path/to/key',
+            ssl_crl_cache_hash_dir: '/path/to/crl_cache/dir'
+        end
+
+        it 'sets ssl crl check setting to specified value' do
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{crl_check,true})
+          is_expected.to contain_file('rabbitmq.config').with_content(%r{\[crl_cache,[[:space:]]+{ssl_crl_hash_dir,[[:space:]]+{internal,[[:space:]]+[{dir, "/path/to/crl_cache/dir"}]}}\]})
```
wyardley, line 1191: nit, but I think `\s` might be more consistent than `[[:space:]]` with the rest of the file possibly?

wyardley (commented):
Overall, looks reasonable to me, without knowing a lot of detail about this specific setting. Thanks for adding tests. One minor comment inline, but overall, LGTM

REFERENCE.md
dimonzozo: didn't notice that this is generated, removed this change

spec/classes/rabbitmq_spec.rb
dimonzozo: fixed.

dimonzozo:
Thank you for review! Specs failing in master and I'm not sure if I did everything correctly. Tomorrow I'll take a look and maybe I can fix them.

wyardley:
> Thank you for review! Specs are failing in master and I'm not sure if I did everything correctly. Tomorrow I'll take a look and maybe I can fix them.

Not sure if you had it checked out before, but for me, I had to update the fixtures, or I'd get an error from the old fixtures version of the systemd module not having a newly added parameter (`Systemd::Service_limits[rabbitmq-server.service]: has no parameter named 'selinux_ignore_defaults'`). so you may want to try and see if `bundle exec rake spec_clean` and `bundle exec rake rake spec_prep` helps.

wyardley:
I see also this warning in CI from validation checks:
```
manifests/init.pp:428:double_quoted_strings:WARNING:double quoted string containing no variables
manifests/init.pp:438:double_quoted_strings:WARNING:double quoted string containing no variables
manifests/init.pp:447:double_quoted_strings:WARNING:double quoted string containing no variables
```

manifests/init.pp
```diff
@@ -425,7 +425,7 @@
     }
   }
 
-  if $ssl_crl_check != "false" {
+  if $ssl_crl_check != 'false' {
```
wyardley, line 428: I think these can just be 
```puppet
if $ssl_crl_check {
```

data/common.yaml
```diff
@@ -65,6 +65,9 @@ rabbitmq::ssl_reuse_sessions: true
 rabbitmq::ssl_honor_cipher_order: true
 rabbitmq::ssl_dhfile: ~
 rabbitmq::ssl_ciphers: []
+rabbitmq::ssl_crl_check: 'false'
```
bastelfreak, line 68: you don't need to add data to common.yaml since you already added them to the init.pp. only one of those two places is required (and init.pp is the preferred way)

manifests/init.pp
bastelfreak: I don't think so. The enum will always be true, unless it's undef or `false`. In this case we've not `false` as a boolean but the string `'false'` which evaluates to `true`. tl;dr I think the current implementation is correct

manifests/init.pp
```diff
@@ -368,6 +377,9 @@
   Boolean $ssl_honor_cipher_order                                                                  = true,
   Optional[Stdlib::Absolutepath] $ssl_dhfile                                                       = undef,
   Array $ssl_ciphers                                                                               = [],
+  Enum['true','false','peer','best_effort'] $ssl_crl_check                                         = 'false',
+  Optional[String] $ssl_crl_cache_hash_dir                                                         = undef,
```
bastelfreak, line 381: what about this?
```suggestion
  Optional[Stdlib::Absolutepath] $ssl_crl_cache_hash_dir                                           = undef,
```

manifests/init.pp
wyardley: Oh right. Forgot that it’s an enum.

data/common.yaml
dimonzozo: fixed

manifests/init.pp
dimonzozo: Done.

bastelfreak (approved):

towo (approved):
Looks good to me. Failing CentOS CI jobs are technically irrelephant, but we could wait on #870 to have a formally green CI.

wyardley:
Thanks! Would you mind squashing the commits in your branch if possible?

kenyon:
> Thanks! Would you mind squashing the commits in your branch if possible?

@wyardley you can also squash on merge if you just want to squash down to one commit.

PR #870, opened 2020-12-29
Remove support for CentOS 6

towo:

Support for CentOS 6 is removed from `metadata.json`. There was no stale code to refactor.

CentOS 6 support is making the CI a bad shade of red, and CentOS 6 is EOL, so it should be removed.

Additionally remove a leftover `.nodesets.yml`.

n/a


wyardley:
Do you have a link to failed runs? IIRC, Vox hasn’t broadly dropped support though I know there was some discussion of it, and fairly recent PRs have been green IIRC

towo:
@wyardley Sure, #869 has failed CentOS 6 runs where everything else is green. Problem is that the repositories aren't available anymore to pull packages from.

wyardley:
@bastelfreak do we flag this as backwards-incompatible?

bastelfreak:
yes. just backwards-incompatible. otherwise it might confuse the changelog generator.

PR #874, opened 2021-02-24
make sure the rabbitmq_version method actually returns the version.

TomRitserveldt:
with the previous code there were scenarios where version on line 46 was empty and the following if did not work as expected.

### DIFF
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -40,6 +40,7 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
     output = rabbitmqctl('-q', 'status')
     version = output.match(%r{RabbitMQ.*?([\d\.]+)})
     @rabbitmq_version = version[1] if version
+    @rabbitmq_version
   end
 
   def self.rabbitmqctl_list(resource, *opts)


wyardley:
Will this also solve https://github.com/voxpupuli/puppet-rabbitmq/pull/872 ?

Is there a way to unit test this?

Also, can you double check it your commit email address is tied to your GH account?

TomRitserveldt:
that email should be linked now as well.
I will look into expanding the unit tests for this.
as for the other PR, I think it's related for sure, his problem sounds really similar.

I'm wondering now if it wouldnt be better to just use the rabbitmq_version fact that is created [here](https://github.com/voxpupuli/puppet-rabbitmq/blob/master/lib/facter/rabbitmq_version.rb) using rabbitmqadmin --version, instead of using a regex that could break every time rabbitmqctl --status output changes..

wyardley:
> I'm wondering now if it wouldnt be better to just use the rabbitmq_version fact that is created here using rabbitmqadmin --version

I think that may not be available early enough or as consistently, or else may not be as reliably there, since the module installs it in a weird way, but could be wrong.

TomRitserveldt:
That's correct, using the fact was not consistent enough.
My attempts to catch the problem I saw in unit tests failed, so I'm afraid I wont be able to update the unit tests for this PR.

fyi: I can confirm the current regex works for both use cases you mentioned here (https://github.com/voxpupuli/puppet-rabbitmq/pull/872#discussion_r561397049)

wyardley (approved):

PR #876, opened 2021-03-06
CLI Environment Fixes

bishopbm1:
Fixes issue where the local encoding is not correct by adding the correct LC_ALL values to the environment.
Error:
```shell
warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+
+## [v11.0.2](https://github.com/voxpupuli/puppet-rabbitmq/tree/v11.0.2) (2021-03-6)
+
+ - Updated the environment to unclude LANG and LC_ALL vairables to fix issues with trying to run with the incorrect encoding
+ - Fixed issues with when managing repos the Erlang repo was not managed and the required version was not available in EPEL
+
 ## [v11.0.0](https://github.com/voxpupuli/puppet-rabbitmq/tree/v11.0.0) (2021-01-16)
 
 [Full Changelog](https://github.com/voxpupuli/puppet-rabbitmq/compare/v10.3.0...v11.0.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,6 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
-
-## [v11.0.2](https://github.com/voxpupuli/puppet-rabbitmq/tree/v11.0.2) (2021-03-6)
-
- - Updated the environment to unclude LANG and LC_ALL vairables to fix issues with trying to run with the incorrect encoding
- - Fixed issues with when managing repos the Erlang repo was not managed and the required version was not available in EPEL
-
 ## [v11.0.0](https://github.com/voxpupuli/puppet-rabbitmq/tree/v11.0.0) (2021-01-16)
 
 [Full Changelog](https://github.com/voxpupuli/puppet-rabbitmq/compare/v10.3.0...v11.0.0)
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -40,6 +40,7 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
     output = rabbitmqctl('-q', 'status')
     version = output.match(%r{RabbitMQ.*?([\d\.]+)})
     @rabbitmq_version = version[1] if version
+    @rabbitmq_version
   end
 
   def self.rabbitmqctl_list(resource, *opts)
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -25,7 +25,7 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
 
   def self.home_tmp_command(name, path)
     has_command name, path do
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+
+## [v11.0.2](https://github.com/voxpupuli/puppet-rabbitmq/tree/v11.0.2) (2021-03-6)
+
+ - Updated the environment to unclude LANG and LC_ALL vairables to fix issues with trying to run with the incorrect encoding
+ - Fixed issues with when managing repos the Erlang repo was not managed and the required version was not available in EPEL
+
```
wyardley, line 12: This shouldn’t be edited in a PR - it would be generated with a release (same with metadata)

manifests/repo/rhel.pp
```diff
@@ -14,6 +15,17 @@
     gpgcheck => 1,
   }
 
+  # This is required because when using the latest version of rabbitmq because the latest version in EPEL
+  # for Erlang is 22.0.7 which is not compatible: https://www.rabbitmq.com/which-erlang.html
+  yumrepo { 'erlang':
```
wyardley, line 20: Historically, the module is setup in such a way that Erlang configuration is out of the module’s scope. There have been some proposals to fix puppet-erlang, but I think any fix along these lines would have to be made in that module and then update that dependency here.

bishopbm1: Normally I would agree except this is a version of Erlang packaged and maintained by RabbitMQ https://www.rabbitmq.com/which-erlang.html#erlang-repositories so i think that it makes sense to have it in here only when managing the repos and downloading from their package cloud.

CHANGELOG.md
bishopbm1: Ah i didn't realize. I reverted the changes!

manifests/repo/rhel.pp
wyardley: The issue would be if users also need a competing (but compatible) erlang version on the same machine. Also, adding this here only would be inconsistent with how the module works on other OSes / distros. However, if https://github.com/voxpupuli/puppet-erlang gets released finally, it could be made a dependency or a recommended install, and the issue could be solved that way.

wyardley: With the current behavior, the module is also not (by default) using the official rabbitmq packages either, tho there have been some proposals to switch that to being the default.

most of them have gotten stuck on similar issues around the Erlang dependency ssues.

bishopbm1: I can add the same code for the other OS's to make it consistent, that definitely makes sense. We could also add second variable to manage this repo if you think that may be a better option? 

When using RHEL 8 with out this addition made the module unusable. So I definitely think having this in here even if it is optional would be a good add. Especially since https://github.com/voxpupuli/puppet-erlang isn't released yet.

wyardley: It's pretty much _always_ been the case with this module that the user has to satisfy the Erlang dependency themself, I think. It could potentially be documented better, but IMO, it's better if that's done through a separate module vs. a parameter in this one.

Keep in mind that the module doesn't list support for RHEL 8 yet. https://github.com/voxpupuli/puppet-rabbitmq/pull/842 was @nmaludy's WIP working on adding support for it, which has some good discussion about these points.

I think if we
a) make the module default to using the rabbitmq package vs. OS vendor's default as the default again (as a breaking / major change; essentially toggling the value for `repos_ensure`)
b) get @nmaludy's puppet-erlang actually building and released and use it in the integration tests + recommend that users use it to manage erlang installation

that can work.

That said, so far, haven't seen anyone follow through with getting all this done in a way that works, passes tests, etc.

wyardley: Also, could there be issues with the newer erlang if older versions of rabbitmq were chosen?

lib/puppet/provider/rabbitmq_cli.rb
```diff
@@ -25,7 +25,7 @@ def self.append_to_path(dir)
 
   def self.home_tmp_command(name, path)
     has_command name, path do
-      environment HOME: '/tmp'
+      environment HOME: '/tmp', LANG: 'en_US.UTF-8', LC_ALL: 'en_US.UTF-8'
```
wyardley, line 28: IIRC, this should already be set for most use cases via
https://github.com/voxpupuli/puppet-rabbitmq/blob/dfd7607753b1fe4a4ad1de41eb31183d1a23dca8/data/common.yaml#L82-L83

I'm not 100% sure that that applies to commands run in the provider, but I _think_ it should; I know there have been some previous fixes around this error.

Since `environment_variables` is a parameter already, maybe the right fix here is to somehow either add `HOME` to defaults or merge the defaults with the environment for the CLI here?

Either way, I think the environment fix here should probably be a separate PR, as it's a narrower fix that we should be able to merge quickly, and likely affects other OSes besides RHEL.

Did you confirm that both `LANG` and `LC_ALL` were necessary? `LC_ALL` has worked in the past and is the somewhat narrower fix.

bishopbm1:
I made the requested changes to the variables and the erlang repo.

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppet-rabbitmq",
-  "version": "11.0.1-rc0",
+  "version": "11.0.2-rc0",
```
wyardley, line 3: this shouldn't be adjusted in here. it will get bumped with a release.

wyardley (commented):
Can you update the PR title / commit message to clarify that it's an environment fix vs. a RHEL fix?

bishopbm1:
Removed metadata tag and fix the title and message.

wyardley (approved):

PR #877, opened 2021-04-15
Add support for oom_score_adj

jlutran:

Add support for OOM score in init script and systemd limits.
Default value is 0, acceptable values range from -1000 to +1000.

Documentation : Chapter 3.1 of https://www.kernel.org/doc/Documentation/filesystems/proc.txt

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -140,6 +140,8 @@
 #   to 'False' and set 'erlang_cookie'.
 # @param file_limit
 #   Set rabbitmq file ulimit. Defaults to 16384. Only available on systems with `$::osfamily == 'Debian'` or `$::osfamily == 'RedHat'`.
+# @param oom_score_adj
+#   Set rabbitmq-server process OOM score. Defaults to 0.
 # @param heartbeat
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -394,6 +396,7 @@
   Boolean $wipe_db_on_cookie_change                                                                = false,
   String $cluster_partition_handling                                                               = 'ignore',
   Variant[Integer[-1],Enum['unlimited'],Pattern[/^(infinity|\d+(:(infinity|\d+))?)$/]] $file_limit = 16384,
+  Optional[Integer] $oom_score_adj                                                                 = 0,
```
bastelfreak, line 399: if the default is 0, you cannot set it to undef so `Optional` doesn't make sense here. I'm not sure if the default here should be undef or 0 :thinking:

jlutran: You're right, but since the current code write a value in ```templates/default.erb```, I will remove the ```Optional``` and keep the default value :)

bastelfreak:
thanks for the PR. Can you take a look at the failing spec tests?

bastelfreak (approved):

PR #878, opened 2021-04-15
Drop Puppet 5 support; require 6.1.0 / Drop RedHat 6 support

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -53,7 +53,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 8.0.0"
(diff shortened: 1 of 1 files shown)


llowder (approved):

wyardley (approved):
I’m fine with this; from what I’ve seen in other PRs and issues, I’m not sure it will actually work with Debian 8 yet, though.

see #842 #845 etc.

that said, may still be better to ship a narrower one supporting it in metadata if we can sort out any major issues and then fix anything else incrementally.

wyardley:
@bastelfreak can you rebase this one / fix conflicts?

wyardley (approved):

PR #881, opened 2021-04-16
Enable Puppet 7 support (where applicable)

bastelfreak:


bastelfreak:
This is expected to fail on Debian8/Ubuntu16.04 on Puppet 7 because Puppet 7 isn't supported on those platforms.

foxxx0 (approved):

PR #884, opened 2021-04-17
camptocamp/systemd: Allow 3.x

bastelfreak:
also contains:
* https://github.com/voxpupuli/puppet-rabbitmq/pull/880
* https://github.com/voxpupuli/puppet-rabbitmq/pull/882

Closes #875

PR #885, opened 2021-04-22
Remove superfluous daemon-reload code and raise minimum required puppet version to 6.1.0

### LINKED ISSUE #875: Not compatible with latest puppet-systemd

- Puppet: 6.21.1
- Ruby: 2.5
- Distribution: CentOS8
- Module version: Latest puppet-rabbitmq commit

with latest puppet-systemd commit issue should be seen as the removed class is being used in puppet-rabbitmq https://github.com/voxpupuli/puppet-rabbitmq/blob/7f806ac6983ec8d087a2c9e86b4e1467198d6981/manifests/service.pp#L26-L28

## What are you seeing
Could not find resource 'Class[Systemd::Systemctl::Daemon_reload]' for relationship on 'Service[rabbitmq-server]' 

It should be made compatible with latest puppet-systemd(no tag released yet). We test latest commits and hit the issue for now we have pinned puppet-systemd on our side https://review.rdoproject.org/r/#/c/32088/.

## Output log
https://logserver.rdoproject.org/ci.centos.org/weirdo-generic-packstack-scenario001/12896/weirdo-project/logs/latest/manifests/192.168.1.213_controller.pp.log.txt.gz

https://github.com/camptocamp/puppet-systemd/commit/97dd16fa32886b5b0f77a6f38a4953d4c1511c0e dropped support for puppet 4 and 5 as they are EOL, and removed Systemd::Systemctl::Daemon_reload with reasoning that it's not needed with puppet6.

TwizzyDizzy:
This removes the superfluous daemon-reload code in this module.

The upstream module (https://github.com/camptocamp/puppet-systemd) removed the support for puppet 4 and 5 that made this reload construct necessary.

They removed their code in this PR: https://github.com/camptocamp/puppet-systemd/pull/171

On the other hand, https://github.com/voxpupuli/puppet-rabbitmq/pull/884 doesn't go far enough, as it simply raises the possible version of the systemd module without taking into account the EOL of Puppet 4 and 5.

Fixes #875

Cheers
Thomas

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -22,9 +22,5 @@ class rabbitmq::service (
       hasrestart => true,
       name       => $service_name,
     }
-
-    if $facts['systemd'] {
-      Class['systemd::systemctl::daemon_reload'] -> Service['rabbitmq-server']
-    }
   }
 }
--- a/metadata.json
+++ b/metadata.json
@@ -54,7 +54,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 8.0.0"
+      "version_requirement": ">= 6.1.0 < 8.0.0"
     }
   ],
   "dependencies": [


wyardley:
May want to wait for #878 for that - I’d like to get 11.1 out and then we can ship the breaking release. I think #884 was an incremental step in that direction.

TwizzyDizzy:
Agreed, thanks for pointing this out!

Cheers
Thomas

wyardley:
It does mean that we probably should continue without #884 either for now. Let’s keep this open and rebase / merge after 11.1.

ekohl (approved):

wyardley:
@TwizzyDizzy can you resolve the conflicts from #886, which will ship first in a non-breaking release?

wyardley:
After #878 also won't need to raise min puppet version anymore, because it includes that,

TwizzyDizzy:
Hey @wyardley

Looking at the changes in master, I think my PR is obsolete

The daemon-reload code has not been removed but has an additional condition the catches the case of the upstream systemd-module version being 3.0.0 and above.

Hence, I don't think there is any need to raise the minimum required version either.

Do you agree?

Cheers
Thomas

wyardley:
> Hence, I don't think there is any need to raise the minimum required puppet version either.

> On the other hand, one might still vote for raising the minimum required version on the basis of the upstream module not giving the guarantee of puppet 4+5 support anymore, so breaking changes to the rabbitmq module might occur at any given point which is something one might want to prevent categorically.

I don't have a strong opinion on it, personally. I think if 4 / 5 are already EOL, since we've just cut 11.1, there's no reason not to drop support for it in the next release (which is already a breaking one), as well as simplify the code here further.

wyardley:
@TwizzyDizzy if you want to proceed with this one, could you rebase + resolve the conflicts?

wyardley:
Closing based on comments in #898

PR #886, opened 2021-05-06
Compatibility with camptocamp/systemd 3.x

TheMeier:
Remove `systemd::systemctl::daemon_reload`

Fails with puppet-systemd >= 3.0.0
See https://github.com/camptocamp/puppet-systemd/compare/2.12.0...3.0.0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R224-R229
`systemd::systemctl::daemon_reload` was removed.

@ekohl

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -23,8 +23,5 @@ class rabbitmq::service (
       name       => $service_name,
     }
 
-    if $facts['systemd'] {
-      Class['systemd::systemctl::daemon_reload'] -> Service['rabbitmq-server']
-    }
   }
 }
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -22,6 +22,8 @@ class rabbitmq::service (
       hasrestart => true,
       name       => $service_name,
     }
-
+    if $facts['systemd'] and defined(Class['systemd::systemctl::daemon_reload']) {
+      Class['systemd::systemctl::daemon_reload'] -> Service['rabbitmq-server']
+    }
   }
 }
--- a/metadata.json
+++ b/metadata.json
@@ -68,7 +68,7 @@
     },
     {
       "name": "camptocamp/systemd",
-      "version_requirement": ">= 2.10.0 < 3.0.0"
+      "version_requirement": ">= 2.10.0 < 4.0.0"
     }
   ],
   "tags": [


manifests/service.pp
```diff
@@ -23,8 +23,5 @@
       name       => $service_name,
     }
```
ekohl, line 25: CI will also want this line removed

ekohl (changes requested):
The daemon reload is needed on Puppet < 6.1.0 which according to `metadata.json` is still supported. That means this patch will break those systems.

I think the best way is to change the conditional to `if defined(Class['systemd::systemctl::daemon_reload'])`. Due to the inclusion of `systemd::service_limits` in `rabbitmq::config` it should be guaranteed to be loaded and the best way to support both `camptocamp/systemd` < 3 and >= 3.

Then it should also allow version 3.x in `metadata.json`.

manifests/service.pp
TheMeier: Thanks for the pointers. Changed it accordingly.

TheMeier:
The fainling test on puppet 7 debian 8 is not from the changes:
```debian8-64.example.com 09:44:54$ wget -O /tmp/puppet.deb http://apt.puppet.com/puppet7-release-jessie.deb
  --2021-05-06 09:44:54--  http://apt.puppet.com/puppet7-release-jessie.deb
  Resolving apt.puppet.com (apt.puppet.com)...   2600:9000:2191:8800:1d:fc37:1cc0:93a1, 2600:9000:2191:ac00:1d:fc37:1cc0:93a1, 2600:9000:2191:2e00:1d:fc37:1cc0:93a1, ...
  Connecting to apt.puppet.com (apt.puppet.com)|2600:9000:2191:8800:1d:fc37:1cc0:93a1|:80...   failed: Network is unreachable.```

ekohl:
I think we can ignore that failure on merge. A separate PR to drop Debian 8 would be fine for me.

ekohl (approved):

TheMeier:
> I think we can ignore that failure on merge. A separate PR to drop Debian 8 would be fine for me.

Ah right debian 8 is gone for some time already. I will prepare a  PR

ekohl:
I don't use RabbitMQ myself, I just reasoned from my Puppet knowledge. I also made a suggestion for the current implementation so perhaps it's good if someone else also took a look. If nobody does and you can verify this solves your problem then I don't mind merging it myself.

wyardley:
If we’re already cutting a breaking release, what about doing #885?

ekohl:
I'd be ok with that.

wyardley:
May as well ship this one with 11.1.0 as well, and then merge #885 as part of the breaking release.

PR #897, opened 2021-09-24
Fix Bug #804 'rabbitmqadmin will not be upgraded'

### LINKED ISSUE #804: rabbitmqadmin will not be upgraded

- Puppet: 5.5.14
- Ruby: 2.4.5
- Distribution: Ubuntu
- Module version: 9.0.0

Is there any way of forcing an update of rabbitmqadmin installed via `archive`?
Right now, if you've upgraded rabbitmq-server, rabbitmqadmin will stay the same as the file `/var/lib/rabbitmq/rabbitmqadmin` exists and won't be downloaded again.

Manually removing this file will download the newest rabbitmqadmin version.

s-johansson:

Add removal of rabbitmqadmin script and trigger refresh upon package upgrade or downgrade

Fixes #804

### DIFF
--- a/manifests/install/rabbitmqadmin.pp
+++ b/manifests/install/rabbitmqadmin.pp
@@ -40,9 +40,24 @@ class rabbitmq::install::rabbitmqadmin {
     }
 
     if !($rabbitmq::use_config_file_for_plugins) {
-      $rabbitmqadmin_archive_require = [Class['rabbitmq::service'], Rabbitmq_plugin['rabbitmq_management']]
+      $rabbitmqadmin_archive_require = [
+        Class['rabbitmq::service'],
+        Rabbitmq_plugin['rabbitmq_management'],
+        Exec['remove_old_rabbitmqadmin_on_upgrade']
+      ]
     } else {
-      $rabbitmqadmin_archive_require = [Class['rabbitmq::service'], File['enabled_plugins']]
+      $rabbitmqadmin_archive_require = [
+        Class['rabbitmq::service'],
+        File['enabled_plugins'],
+        Exec['remove_old_rabbitmqadmin_on_upgrade']
+      ]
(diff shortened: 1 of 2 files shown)


wyardley:
Hi - thanks for the contribution

Can you look at the formatting errors? There may be some expectations that also need to be updated.

Also, can you make the subject / commit message not include the bug number, and instead explain the fix, maybe something along the lines of
"Reinstall rabbitmqadmin on upgrade"?

s-johansson:
@wyardley I was finally able to figure out the needed adjustments. I think, that's it. Happy new year :four_leaf_clover: !

manifests/install/rabbitmqadmin.pp
```diff
@@ -40,9 +40,24 @@
...
+      $rabbitmqadmin_archive_require = [
+        Class['rabbitmq::service'],
+        File['enabled_plugins'],
+        Exec['remove_old_rabbitmqadmin_on_upgrade']
+      ]
+    }
+
+    Exec { 'remove_old_rabbitmqadmin_on_upgrade':
+      path        => ['/bin','/usr/bin','/sbin','/usr/sbin'],
+      command     => "rm ${rabbitmq::rabbitmq_home}/rabbitmqadmin",
```
wyardley, line 58: Could we do this with a file resource with `ensure => absent` vs. using a shell exec (but still using the same `onlyif`)?

s-johansson: I am aware, that exec resources should be avoided, if there is any possibilty to fullfil the task as well with other basically idempotent resources. In order to provide idempotence for this exec resource, `onlyif` and `refreshonly` parameter are set. 

Managing a file resource presence dependent on the refresh signal seems not to be possible imho.

PR #899, opened 2021-12-15
modulesync 5.2.0

bastelfreak:
modulesync 5.2.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,26 +9,14 @@ on:
     tags:
       - '*'
 
-env:
-  BUNDLE_WITHOUT: development:test:system_tests
-
 jobs:
-  deploy:
-    name: 'deploy to forge'
-    runs-on: ubuntu-latest
-    if: github.repository_owner == 'voxpupuli'
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-      - name: Setup Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: '2.7'
-          bundler-cache: true
-      - name: Build and Deploy
-        env:
-          # Configure secrets here:
-          #  https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets
-          BLACKSMITH_FORGE_USERNAME: '${{ secrets.PUPPET_FORGE_USERNAME }}'
-          BLACKSMITH_FORGE_API_KEY: '${{ secrets.PUPPET_FORGE_API_KEY }}'
-        run: bundle exec rake module:push
+  release:
+    name: Release
+    uses: voxpupuli/gha-puppet/.github/workflows/release.yml@v1
+    with:
+      allowed_owner: 'voxpupuli'
+    secrets:
+      # Configure secrets here:
+      #  https://docs.github.com/en/actions/security-guides/encrypted-secrets
+      username: ${{ secrets.PUPPET_FORGE_USERNAME }}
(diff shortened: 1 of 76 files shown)


.github/workflows/release.yml
```diff
@@ -9,26 +9,14 @@ on:
...
-        run: bundle exec rake module:push
+  release:
+    name: Release
+    uses: voxpupuli/gha-puppet/.github/workflows/release.yml@v1
+    with:
+      allowed_owner: 'voxpupuli'
+    secrets:
+      # Configure secrets here:
+      #  https://docs.github.com/en/actions/security-guides/encrypted-secrets
+      username: ${{ secrets.PUPPET_FORGE_USERNAME }}
```
wyardley, line 21: @bastelfreak are these secrets set on this repo, and is / has been the release automagic for this project yet? i.e., should I not run the release task locally anymore on the release PR?

bastelfreak: no. you still need to run the travis_release task (which is now also available as release`). The rake task pushes the git tag. The GitHub action builds a puppet module and pushes it to the forge (when a tag is pushed). The secrets are defined on our org level to authenticate against the forge API.

wyardley:
@bastelfreak feel free to recreate this entirely, or to take as much or as little of the changes already there if they help.

lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin.rb
```diff
@@ -34,11 +32,11 @@ def self.instances
           name = ''
         end
         # Convert output of arguments from the rabbitmqctl command to a json string.
-        if !arguments.nil?
+        if arguments.nil?
+          arguments = '{}'
+        else
           arguments = arguments.gsub(%r{^\[(.*)\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
```
wyardley, line 38: @root-expert I know I had one case where the frozen string literals caused issues with reassigning a variable like this...I don't think it was in this case, but this one will be Ok, right?

wyardley (approved):
LGTM. One quick question inline. Since I also pushed to this PR, would be great if you could flag someone else from Voxpup from review.

I think we can ship this without arch passing, but not sure if @bastelfreak knows what's going on and / or if there's anyone else who can look at the Archlinux failures?

lib/puppet/provider/rabbitmq_exchange/rabbitmqadmin.rb
smortex: `String#gsub` return a new String so this should be safe. `String#gsub!` would not.

spec/acceptance/class_spec.rb
```diff
@@ -157,13 +151,14 @@ class { 'rabbitmq':
       it { is_expected.to be_running }
     end
 
-    describe port(5672) do
+    describe port(5672), :port5671 do
```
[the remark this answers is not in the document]
wyardley: Feel free to rework this better. This was my lame / quick way of squelching the duplicate description error IIRC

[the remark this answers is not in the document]
root-expert: Yea I stole this from the cop's description.
Tagging it will fix the duplicate error. I don't find a better solution for this kinds of `describe`

wyardley:
@smortex are you good with this as-is?
@root-expert is this one ready to go if approved?

PR #900, opened 2022-02-07
Add official support for CentOS 8

kajinamit:

This change adds official support for CentOS 8. The current logic is
almost compatible except for the package name of Python 2, which is
python2 in CentOS 8 while python in CentOS 7.

N/A

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -12,12 +12,16 @@ describe 'rabbitmq class:' do
     service_name = 'rabbitmq'
   end
 
+  repos_ensure = (fact('os.family') == 'RedHat')
+
   context 'default class inclusion' do
     let(:pp) do
       <<-EOS
-      class { 'rabbitmq': }
+      class { 'rabbitmq':
+        repos_ensure => #{repos_ensure},
+      }
       if $facts['os']['family'] == 'RedHat' {
-        class { 'erlang': epel_enable => true}
+        class { 'erlang': repo_source => 'packagecloud' }
         Class['erlang'] -> Class['rabbitmq']
       }
       EOS
@@ -52,10 +56,11 @@ describe 'rabbitmq class:' do
     let(:pp) do
       <<-EOS
         class { 'rabbitmq':
+          repos_ensure   => #{repos_ensure},
           service_ensure => 'stopped',
         }
         if $facts['os']['family'] == 'RedHat' {
-          class { 'erlang': epel_enable => true}
+          class { 'erlang': repo_source => 'packagecloud' }
           Class['erlang'] -> Class['rabbitmq']
         }
       EOS
@@ -72,20 +77,23 @@ describe 'rabbitmq class:' do
   context 'service is unmanaged' do
     it 'runs successfully' do
       pp_pre = <<-EOS
-        class { 'rabbitmq': }
+        class { 'rabbitmq':
+          repos_ensure => #{repos_ensure},
+        }
         if $facts['os']['family'] == 'RedHat' {
-          class { 'erlang': epel_enable => true}
+          class { 'erlang': repo_source => 'packagecloud' }
           Class['erlang'] -> Class['rabbitmq']
         }
       EOS
 
       pp = <<-EOS
         class { 'rabbitmq':
+          repos_ensure   => #{repos_ensure},
           service_manage => false,
-          service_ensure  => 'stopped',
+          service_ensure => 'stopped',
         }
         if $facts['os']['family'] == 'RedHat' {
-          class { 'erlang': epel_enable => true}
+          class { 'erlang': repo_source => 'packagecloud' }
           Class['erlang'] -> Class['rabbitmq']
         }
       EOS
@@ -104,6 +112,7 @@ describe 'rabbitmq class:' do
     let(:pp) do
       <<-EOS
       class { 'rabbitmq':
+        repos_ensure      => #{repos_ensure},
         service_manage    => true,
         port              => 5672,
         admin_enable      => true,
@@ -137,6 +146,7 @@ describe 'rabbitmq class:' do
     let(:pp) do
       <<-EOS
         class { 'rabbitmq':
+          repos_ensure      => #{repos_ensure},
           service_manage    => true,
           port              => 5672,
           admin_enable      => true,
@@ -172,6 +182,7 @@ describe 'rabbitmq class:' do
     let(:pp) do
       <<-EOS
         class { 'rabbitmq':
+          repos_ensure    => #{repos_ensure},
           service_manage  => true,
           admin_enable    => true,
           node_ip_address => '0.0.0.0',
@@ -203,6 +214,7 @@ describe 'rabbitmq class:' do
     let(:pp) do
       <<-EOS
         class { 'rabbitmq':
+          repos_ensure          => #{repos_ensure},
           service_manage        => true,
           port                  => 5672,
           admin_enable          => true,
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -7,7 +7,7 @@ configure_beaker do |host|
   when 'Debian'
     install_module_from_forge_on(host, 'puppetlabs-apt', '>= 4.1.0 < 9.0.0')
   when 'RedHat'
-    install_module_from_forge_on(host, 'garethr-erlang', '>= 0.3.0 < 1.0.0')
+    install_module_from_forge_on(host, 'puppet-erlang', '>= 1.0.0 < 2.0.0')
     if fact_on(host, 'os.selinux.enabled')
       # Make sure selinux is disabled so the tests work.
       on host, puppet('resource', 'exec', 'setenforce 0', 'path=/bin:/sbin:/usr/bin:/usr/sbin', 'onlyif=which setenforce && getenforce | grep Enforcing')
--- a/data/family/RedHat.yaml
+++ b/data/family/RedHat.yaml
@@ -1,4 +1,5 @@
 ---
+rabbitmq::python_package: 'python3'
 rabbitmq::package_name: 'rabbitmq-server'
 rabbitmq::service_name: 'rabbitmq-server'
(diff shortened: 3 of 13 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -5,7 +5,18 @@
   when 'Debian'
     install_module_from_forge_on(host, 'puppetlabs-apt', '>= 4.1.0 < 9.0.0')
   when 'RedHat'
-    install_module_from_forge_on(host, 'garethr-erlang', '>= 0.3.0 < 1.0.0')
+    if fact_on(host, 'os.release.major').to_i < 8
+      # In Cent OS <8, rabbitmq-server is available in EPEL.
+      install_module_from_forge_on(host, 'puppet/epel', '>= 3.0.0')
+    else
+      # In Cent OS >=8, rabbitmq-server is no longer available in EPEL
+      on host, 'curl -s https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | sudo bash'
```
bastelfreak, line 13: that's quite hacky. can't we implement it into the puppet module to configure the repository?

spec/acceptance/class_spec.rb
```diff
@@ -16,13 +16,16 @@
     service_name = 'rabbitmq'
   end
 
+  repos_ensure = case fact('os.family')
+                 when 'RedHat'
+                   true
+                 end
```
wyardley, line 22: So the current behavior of the module IIRC is that `repos_ensure` defaults to `false`. Changing it for the acceptance test here means (I think) that this will now not test the module’s default behavior.

there were some discussions of changing that default again in the module itself, but I think that would be more involved, and that’s where previous attempts at this have kind of stalled in the absence of an active maintainer.

wyardley: #845 has some good discussion and I think is a good starting point.

bastelfreak: I think the module should provide sane and working defaults for the different operating systems and the tests should at least cover the defaults. If we test non-default values as well that's a nice addition.

kajinamit: I understand what you mentioned but even for CentOS7 currently this module can't provide complete configuration. This still relies on puppet-eralang to enable epel, now.

kajinamit: Also, AFAIK Ubuntu doesn't require the additional repo and can install rabbitmq from usual OS repos. If we enable repos then default repos will be replaced, which I don't think would be ideal setup.

manifests/repo/rhel/erlang.pp
```diff
@@ -0,0 +1,23 @@
+# Makes sure that the Packagecloud repo for erlang is installed
+#
+# @api private
+class rabbitmq::repo::rhel::erlang (
+  $location          = "https://packagecloud.io/rabbitmq/erlang/el/${facts['os'][release][major]}/\$basearch",
```
bastelfreak, line 5: can you use Stdlib::HTTPSUrl here?
```suggestion
  Stdlib::HTTPSUrl $location = "https://packagecloud.io/rabbitmq/erlang/el/${facts['os'][release][major]}/\$basearch",
```

manifests/repo/rhel/erlang.pp
```diff
@@ -0,0 +1,23 @@
...
+  String $key_source = $rabbitmq::erlang_package_gpg_key,
```
bastelfreak, line 6:
```suggestion
  Stdlib::HTTPSUrl $key_source = $rabbitmq::erlang_package_gpg_key,
```

manifests/init.pp
```diff
@@ -355,6 +359,7 @@
   Optional[Variant[Numeric, String]] $package_apt_pin                                              = undef,
   String $package_ensure                                                                           = 'installed',
   Optional[String] $package_gpg_key                                                                = undef,
+  Optional[String] $erlang_package_gpg_key                                                         = undef,
```
bastelfreak, line 362:
```suggestion
  Optional[Stdlib::HTTPSUrl] $erlang_package_gpg_key                                           = undef,
```

manifests/repo/rhel/erlang.pp
kajinamit: This is used to determine baseurl in yum repos and it is not always https url but can be different format like http url or file:// or ftp:// . So we can't use HTTPSUrl and I'm not aware of any good alternatives.

manifests/repo/rhel/erlang.pp
kajinamit: this is passed to --import option of the rpm command which can take various types like https url, http url or file path so HTTPSUrl doesn't work.

spec/acceptance/class_spec.rb
root-expert: Why not enable `repos_ensure` only for RedHat? Since as it seems only RedHat needs extra repos to work out of the box.
That's a common pattern IMO

spec/spec_helper_acceptance.rb
kajinamit: This is now taken care by the puppet module.

manifests/repo/rhel/erlang.pp
kajinamit: I'll close this conversation according to my previous comment.

manifests/repo/rhel/erlang.pp
kajinamit: The same as the previous comment.

manifests/init.pp
kajinamit: The same as the previous comment.

spec/acceptance/class_spec.rb
kajinamit: This module has never set up the required repositories by default. The current implementation relies on epel repository enabled externally. This change still keeps the same policy and let the users to enable the repository setup by their own(any external modules or setting this parameter.)
I initially hesitated to enable this because it can break existing usage(for example we use this module to deploy OpenStack but in that case we use rabbitmq packages provided by RDO, not packagecloud). However if all agrees we should enable this and make the all users to handle this change properly then I can enable it.

kajinamit:
It seems archi linux job is broken. I see it's failing with the same errors in https://github.com/voxpupuli/puppet-rabbitmq/pull/924 and I'm not sure how I can fix the job, honestly.

wyardley:
> It seems archi linux job is broken.
Hi  - yes, don't worry about that. It's an issue bootstrapping rabbitmqadmin, and has been there for a while now.

kajinamit:
thanks @wyardley .

I've rebased this on the latest master. it seems merge conflict tag is not purged automatically.

I'm wondering what I can do to get this PR merged. This has been floating for a while but I'm really eager to get this(and the subsequent work to support CentOS9) because we are extensively using this for OpenStack deployment.

As I mentioned in the rely I'm not too sure whether I should enable the repository management which likely cause problems during upgrade. If people think that is the only blocker then I can fix that.

wyardley:
@kajinamit maybe worth reading this comment in another PR that I made the other day too
https://github.com/voxpupuli/puppet-rabbitmq/issues/816#issuecomment-1454294332

High level, while I don't want to bikeshed this to death, this project doesn't have a real active maintainer right now, and so it has been hard to find someone who's able to get some of the things mentioned there working in a safe and testable way.

I do really disagree that we should be installing / managing erlang in this project, at least by default, so at a minimum, we cannot force that behavior without making it configurable and, ideally, set to false by default. That's because people may already have differing needs for which erlang version they need to manage, and / or may already be installing it, and it's important that this module first not break stuff. I'd be Ok with some more junk code to set it up in the integration test, and continue to document that the user needs to manage setting up erlang themself.

I think getting the integration tests running locally would be a good start, if you don't have those working yet; I have had struggles getting that to work recently, but it's honestly going to be very hard to do much on that if you don't have the test suite working (and exec-able into) locally.

You may be able to find some folks from vox who are able to help (esp. w/r/t getting integration tests working properly, as well as some ideas about the best way to handle the erlang dep) on their IRC channel. You could also see if there's any willingness to bump the metadata without appropriate and full working integration tests, but that's not a call I'd make on my own.

spec/acceptance/class_spec.rb
wyardley: @kajinamit this module did, at one point, when Puppet still maintained it, setup the RabbitMQ repo by default. It also had an option to include the simple distribution of erlang that rabbitmq uses.

For better or worse, I am responsible for making that change back then (#493). At the time, it seemed like a sane deafult. I think with hindsight, going back to the old behavior (with `repos_ensure` being default again, as a breaking change) is probably where things should go, it's just a matter of getting there when this project doesn't have an active maintainer.

wyardley:
If https://github.com/voxpupuli/puppet-erlang still is semi-maintained, and works(ish), one option would be to use that in the integration tests if it's not already used there, and just have it be a docs fix or optional dependency for users to use it.

wyardley:
Sorry, I’m using the two terms interchangeably. Acceptance tests.

kajinamit:
> If https://github.com/voxpupuli/puppet-erlang still is semi-maintained, and works(ish), one option would be to use that in the integration tests if it's not already used there, and just have it be a docs fix or optional dependency for users to use it.

This is a really good idea, because the module support installing earlang from packagecloud. Let me explore this option.

kajinamit:
ok it seems we have to get the new release of puppet-erlang which contains https://github.com/voxpupuli/puppet-erlang/pull/15 .

wyardley:
> ok it seems we have to get the new release of puppet-erlang which contains https://github.com/voxpupuli/puppet-erlang/pull/15 .

Sounds good. Key thing is, let's just document it vs. making it a module dependency (since it's not needed for people who bootstrap erlang on their own), or make including it optional vs. some kind of bool, but use it for setting up erlang in the acceptance test if needed.

Other thought: to keep stuff small, maybe we should release the `repos_ensure: true` breaking change as its own change first (to keep the PRs smaller), and then work on centos 8 / debian 10 etc. support after that?

kajinamit:
Thanks. I just got the job passing. I'll try to find time to attempt to enable repose_ensure by default later this week.

wyardley:
Seems like CentOS 8 actually was EOL before this PR was even created?
@bastelfreak do you think we should provide support for it in one release, or just close?

PR #907, opened 2022-04-27
Fixing Yum repo

bishopbm1:
We have notice some errors coming up with this module and it seems as so packagecloud has change the way they want their repos set up. We have had to update other repos as well.
Package cloud documentation:
https://packagecloud.io/rabbitmq/rabbitmq-server/install#bash-rpm
https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/config_file.repo?os=rhel&dist=8&source=script

Error:
```
Notice: /Stage[main]/Rabbitmq::Repo::Rhel/Exec[rpm --import https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey]/returns: executed successfully
Error: Execution of '/usr/bin/dnf -d 0 -e 1 -y install rabbitmq-server' returned 1: Importing GPG key 0xDF309A0B:
 Userid     : "https://packagecloud.io/rabbitmq/erlang (https://packagecloud.io/docs#gpg_signing) <support@packagecloud.io>"
 Fingerprint: 2EBD E413 D3CE 5D35 BCD1 5B7C 71C6 3471 DF30 9A0B
 From       : https://packagecloud.io/rabbitmq/erlang/gpgkey
The GPG keys listed for the "rabbitmq_rabbitmq-server" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: rabbitmq-server-3.9.15-1.el8.noarch
 GPG Keys are configured as: https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey
Error: GPG check FAILED
Error: /Stage[main]/Rabbitmq::Install/Package[rabbitmq-server]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/dnf -d 0 -e 1 -y install rabbitmq-server' returned 1: Importing GPG key 0xDF309A0B:
 Userid     : "https://packagecloud.io/rabbitmq/erlang (https://packagecloud.io/docs#gpg_signing) <support@packagecloud.io>"
 Fingerprint: 2EBD E413 D3CE 5D35 BCD1 5B7C 71C6 3471 DF30 9A0B
 From       : https://packagecloud.io/rabbitmq/erlang/gpgkey
The GPG keys listed for the "rabbitmq_rabbitmq-server" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: rabbitmq-server-3.9.15-1.el8.noarch
 GPG Keys are configured as: https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey
Error: GPG check FAILED
```

### DIFF
--- a/data/family/RedHat.yaml
+++ b/data/family/RedHat.yaml
@@ -1,4 +1,5 @@
 ---
 rabbitmq::package_name: 'rabbitmq-server'
 rabbitmq::service_name: 'rabbitmq-server'
-rabbitmq::package_gpg_key: 'https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey'
+rabbitmq::package_gpg_key: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
+rabbitmq::repo_gpg_key: 'https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey'
--- a/manifests/repo/rhel.pp
+++ b/manifests/repo/rhel.pp
@@ -2,22 +2,34 @@
 #
 # @api private
 class rabbitmq::repo::rhel (
-  $location          = "https://packagecloud.io/rabbitmq/rabbitmq-server/el/${facts['os'][release][major]}/\$basearch",
-  String $key_source = $rabbitmq::package_gpg_key,
+  $location                  = "https://packagecloud.io/rabbitmq/rabbitmq-server/el/${facts['os'][release][major]}/\$basearch",
+  String $repo_key_source    = $rabbitmq::repo_gpg_key,
+  String $package_key_source = $rabbitmq::package_gpg_key,
 ) {
+  # Import package key from rabbitmq to be able to
+  # sign the package and the repo.
+  # rabbitmq key is gpg-pubkey-6026dfca-573adfde
+  exec { "rpm --import ${package_key_source}":
+    path   => ['/bin','/usr/bin','/sbin','/usr/sbin'],
+    unless => 'rpm -q gpg-pubkey-6026dfca-573adfde 2>/dev/null',
+    before => YumRepo['rabbitmq'],
+  }
+
   yumrepo { 'rabbitmq':
-    ensure   => present,
-    name     => 'rabbitmq_rabbitmq-server',
-    baseurl  => $location,
-    gpgkey   => $key_source,
-    enabled  => 1,
-    gpgcheck => 1,
+    ensure        => present,
+    name          => 'rabbitmq_rabbitmq-server',
+    baseurl       => $location,
+    gpgkey        => $repo_key_source,
+    enabled       => 1,
+    gpgcheck      => 1,
+    repo_gpgcheck => 1,
   }
 
   # This may still be needed to prevent warnings
-  # packagecloud key is gpg-pubkey-d59097ab-52d46e88
-  exec { "rpm --import ${key_source}":
-    path   => ['/bin','/usr/bin','/sbin','/usr/sbin'],
-    unless => 'rpm -q gpg-pubkey-6026dfca-573adfde 2>/dev/null',
+  # packagecloud key is gpg-pubkey-4d206f89-5bbb8d59
+  exec { "rpm --import ${repo_key_source}":
+    path    => ['/bin','/usr/bin','/sbin','/usr/sbin'],
+    unless  => 'rpm -q gpg-pubkey-4d206f89-5bbb8d59 2>/dev/null',
+    require => YumRepo['rabbitmq'],
   }
 }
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+**Fixed bugs:**
```
bastelfreak, line 7: please do not update the changelog. we autogenerate it during the release.

bishopbm1: Apologies! It has been fixed.

manifests/repo/rhel.pp
```diff
@@ -6,18 +6,19 @@
...
-    baseurl  => $location,
-    gpgkey   => $key_source,
-    enabled  => 1,
-    gpgcheck => 1,
+    ensure        => present,
+    name          => 'rabbitmq_rabbitmq-server',
+    baseurl       => $location,
+    gpgkey        => $key_source,
+    enabled       => 1,
+    gpgcheck      => 0,
```
wyardley, line 14: are we disabling this because the package is no longer signed?

wyardley: @bishopbm1 while I see that the packagecloud builtin script does set `gpgcheck` to disabled, the RMQ docs do say the package itself should be signed.

https://www.rabbitmq.com/signatures.html
The bit under "Using a Key Server" mentions some other keys.
If at all possible, IMO it would be probably be best / safest to check both the signatures, even if it means we have to separately grab / import some additional keys.

bishopbm1: @wyardley I added the additional key that you referenced. It is working with both check enabled now.

manifests/repo/rhel.pp
```diff
@@ -4,20 +4,32 @@
 class rabbitmq::repo::rhel (
   $location          = "https://packagecloud.io/rabbitmq/rabbitmq-server/el/${facts['os'][release][major]}/\$basearch",
   String $key_source = $rabbitmq::package_gpg_key,
+  String $rabbitmq_key_source = 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc',
```
wyardley, line 7: nit, but looking in other classes, I _think_ these are usually lined up (though it seems like CI is happy).

manifests/repo/rhel.pp
```diff
@@ -4,20 +4,32 @@
 class rabbitmq::repo::rhel (
   $location          = "https://packagecloud.io/rabbitmq/rabbitmq-server/el/${facts['os'][release][major]}/\$basearch",
   String $key_source = $rabbitmq::package_gpg_key,
```
wyardley, line 6: should we make this less ambiguous to match `$rabbitmq_key_source`, like `$repo_key_source`, which I think is what this key is now?

wyardley (commented):
Looks good overall - were you able to see if the same issue exists for Debian as well (when packagecloud is used)? Unfortunately, there's not currently an acceptance test (for either) with `repos_enable` set to `true` IIRC, so hard to know if it's working just because of tests passing in this case.

Not sure if we should include both `$package_gpg_key` and `$repo_gpg_key` at top level to make it less confusing... for sure if we end up needing the same fix for Debian, we should include them both in the main class as overridable parameters.

Couple minor comments inline; also, I can squash merge at merge time if you want, but otherwise, please squash down to one commit once we get everything ready to go (also, please check that your author email is tied to your GitHub account in the commit itself).

manifests/repo/rhel.pp
bishopbm1: @wyardley i will make the variable name changes was just trying to keep the same as possible to avoid confusion. 

I do not have a way to test Debian.

bishopbm1:
@wyardley I believe i addressed all your comments. I did test this with Ubuntu 18.04 and the issue was not present. I added the option as a top level param just incase its every needed.

data/family/RedHat.yaml
```diff
@@ -1,4 +1,5 @@
 ---
 rabbitmq::package_name: 'rabbitmq-server'
 rabbitmq::service_name: 'rabbitmq-server'
-rabbitmq::package_gpg_key: 'https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey'
+rabbitmq::package_gpg_key: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
+rabbitmq::repo_gpg_key: 'https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey'
```
wyardley, line 5: nit: wrong line-ending style here, but we can probably ship it as is.

wyardley (approved):
Thanks! Let's :shipit:

Issue #909, opened 2022-06-29
rabbitmq_plugin not working properly with RabbitMQ 3.10.x


kvisle:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.27.0
- Ruby: 2.5.9p229
- Distribution: Rocky Linux 8
- Module version: 12.0.1
- RabbitMQ Server 3.10.5

## How to reproduce (e.g Puppet code you use)

```bash
puppet resource rabbitmq_plugin
```

## What are you seeing

Abovementioned command returns the following:
```
Error: Could not run: Cannot parse invalid plugins line: Listing plugins with pattern ".*" ...
```

## What behaviour did you expect instead

A list of plugin resources and their ensured status.

PR #912, opened 2022-08-12
Make rabbitmq_plugin resource functional on RabbitMQ 3.10.x

kvisle:
The rabbitmq_plugin resource use the "rabbitmq-plugins"-command when
enumerating enabled plugins.  Starting with rabbitmq 3.10, this command
will output a informational message:
'Listing plugins with pattern ".*" ...' which can be hidden with the -s
or -q parameter.  However, this approach will break rabbitmq_plugin on
older versions of RabbitMQ.

This patch implements a workaround that simply ignores the outputted line
if it exists.

Fixes: #909

### DIFF
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -12,6 +12,7 @@ Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Pr
     end
 
     plugin_list.split(%r{\n}).map do |line|
+      next if line.start_with?('Listing plugins')
       raise Puppet::Error, "Cannot parse invalid plugins line: #{line}" unless line =~ %r{^(\S+)$}
 
       new(name: Regexp.last_match(1))
--- a/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
+++ b/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
@@ -22,6 +22,16 @@ describe provider_class do
     provider.create
   end
 
+  context 'with RabbitMQ version >=3.10.0' do
+    it 'matches plugins' do
+      provider.expects(:rabbitmqplugins).with('list', '-E', '-m').returns <<~EOT
+        Listing plugins with pattern ".*" ...
+        foo
+      EOT
+      expect(provider.exists?).to eq(true)
+    end
+  end
+
   context 'with RabbitMQ version >=3.4.0' do
     it 'calls rabbitmqplugins to enable' do
       provider.class.expects(:rabbitmq_version).returns '3.4.0'


lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
```diff
@@ -11,7 +11,7 @@ def self.instances
       rabbitmqplugins('list', '-E', '-m')
     end
 
-    plugin_list.split(%r{\n}).map do |line|
+    plugin_list.split(%r{\n}).grep_v(/^Listing plugins.*$/).map do |line|
```
wyardley, line 14: Do you think it would work to do something similar to this to conditionally based on the version fact?
https://github.com/voxpupuli/puppet-rabbitmq/blob/00e1cc0d9b1075b786f40fff963fb3314a7c3a82/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb#L24

wyardley: I'm taking a look at this option locally (either way, I think this will need a unit test)

wyardley:
#913 is failing a test, but going to see if that's a viable approach.

wyardley:
Ah, the solution in #913 won't work, because old versions have to be running before the version can be obtained, and it can't start before the plugins work, so there's a chicken / egg problem.

Would you be able to add a unit test for this? Basically, should be 

https://github.com/voxpupuli/puppet-rabbitmq/blob/277e3e8b783e9810b5101e9f7b2cb0073e1b86c7/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb#L14-L17

I think something like this should work

```diff
diff --git a/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb b/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
index ec2e7a4..e7be870 100644
--- a/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
+++ b/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
@@ -22,6 +22,16 @@ describe provider_class do
     provider.create
   end
 
+  context 'with RabbitMQ version >=3.10.0' do
+    it 'matches plugins' do
+      provider.expects(:rabbitmqplugins).with('list', '-E', '-m').returns <<-EOT
+        Listing plugins with pattern ".*" ...
+        rabbitmq_prometheus
+      EOT
+      expect(provider.exists?).to eq(true)
+    end
+  end
+
   context 'with RabbitMQ version >=3.4.0' do
     it 'calls rabbitmqplugins to enable' do
       provider.class.expects(:rabbitmq_version).returns '3.4.0'
```

wyardley (commented):
How would you feel about something like this?

```diff
diff --git a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
index 78c68e2..3c92192 100644
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -17,6 +17,7 @@ Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Pr
     end
 
     plugin_list.split(%r{\n}).map do |line|
+      next if line.start_with?('Listing plugins')
       raise Puppet::Error, "Cannot parse invalid plugins line: #{line}" unless line =~ %r{^(\S+)$}
 
       new(name: Regexp.last_match(1))
diff --git a/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb b/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
index b8d6f7a..2e770d0 100644
--- a/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
+++ b/spec/unit/puppet/provider/rabbitmq_plugin/rabbitmqctl_spec.rb
@@ -12,7 +12,9 @@ describe provider_class do
   let(:provider) { provider_class.new(resource) }
 
   it 'matches plugins' do
-    provider.expects(:rabbitmqplugins).with('list', '-E', '-m').returns("foo\n")
+    provider.expects(:rabbitmqplugins).with('list', '-E', '-m').returns <<~EOT
+      foo
+    EOT
     expect(provider.exists?).to eq(true)
   end
 
@@ -22,6 +24,16 @@ describe provider_class do
     provider.create
   end
 
+  context 'with RabbitMQ version >=3.10.0' do
+    it 'matches plugins' do
+      provider.expects(:rabbitmqplugins).with('list', '-E', '-m').returns <<~EOT
+        Listing plugins with pattern ".*" ...
+        foo
+      EOT
+      expect(provider.exists?).to eq(true)
+    end
+  end
+
   context 'with RabbitMQ version >=3.4.0' do
     it 'calls rabbitmqplugins to enable' do
       provider.class.expects(:rabbitmq_version).returns '3.4.0'
```

kvisle:
Thank you - pushed with your suggestions now.

The `next if`-approach is to be preferred, I'm just not that familiar with Ruby 🙂

kvisle:
I suspect the feiling tests on arch is unrelated to the changes in the pull request.

wyardley (approved):

PR #913, opened 2022-08-12
Fix rabbitmq_plugin with RabbitMQ >= 3.10.x

### LINKED ISSUE #909: rabbitmq_plugin not working properly with RabbitMQ 3.10.x

wyardley:
Resolve an issue with extra output from `rabbitmq_plugin` in versions >= 3.10.x

Fixes #909
Closes #912

### DIFF
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -8,7 +8,12 @@ Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Pr
 
   def self.instances
     plugin_list = run_with_retries do
-      rabbitmqplugins('list', '-E', '-m')
+      cmd = ['list', '-E', '-m']
+
+      # This argument silences some extra output that's new in 3.10, but the
+      # argument is not available in some older versions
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
```diff
@@ -35,6 +40,10 @@ def destroy
   end
 
   def exists?
-    run_with_retries { rabbitmqplugins('list', '-E', '-m') }.split(%r{\n}).include? resource[:name]
+    cmd = ['list', '-E', '-m']
+    # This argument silences some extra output that's new in 3.10, but the
+    # argument is not available in some older versions
+    cmd << '-q' if Puppet::Util::Package.versioncmp(self.class.rabbitmq_version, '3.10') >= 0
```
wyardley, line 46: should this command really be running separately on both `exists?` / `self.instances?` I think there's a cleaner way based on some other providers, but not totally sure the easiest way to do that.

If not, is there another way to DRY up this logic a little?

PR #923, opened 2023-02-17
cluster: add local_node settings

fatpat:
When overriding local node name (using NODENAME environment variable), clustering does not work as tinit_node is tested again hostname or fqdn. By adding the optional local_node to specify which node is the current node we can fix this issue.

Use case: a server with a public and a private interface, rabbitmq cluster nodes communicates through the private interface while hostname and fqdn resolves to the public interface (and that can't be changed because of other dependencies on the server). The only solution is to use dedictated name for the private interface.


lib/puppet/type/rabbitmq_cluster.rb
```diff
@@ -40,6 +40,11 @@
     end
   end
 
+  newparam(:local_node) do
+    desc 'Name of the local node'
+    defaultto('')
```
wyardley, line 45: I'm a bit rusty on Puppet types, but could this default to `nil` or be unset, or does it have to be an empty string?

wyardley (commented):
Hi - thanks for the PR
Sorry for the delay in reviewing it. Overall, I think this makes sense.

Few minor comments. Also, can you reset the changes to `REFERENCE.md`? We can rebuild it when cutting the next release.

Question: your commit seems to be from @openiobot  - can you amend the author to be your own GH username?

fatpat:
> Hi - thanks for the PR Sorry for the delay in reviewing it. Overall, I think this makes sense.
> 
> Few minor comments. Also, can you reset the changes to `REFERENCE.md`? We can rebuild it when cutting the next release.
> 
> Question: your commit seems to be from @openiobot - can you amend the author to be your own GH username?

- REFERENCE.md changes have been reverted
- authors has been fixed
- empty default value has been fixed

lib/puppet/type/rabbitmq_cluster.rb
fatpat: fixed

wyardley:
@fatpat I must have been speaking on outdated information :/ Sorry about that. Looks like the current CI checks want `REFERENCE.md` to be up to date.

fatpat:
> 

fixed

wyardley (approved):

PR #924, opened 2023-03-08
Fix detection of management_ip_address for rabbitmqadmin

kajinamit:
This fixes inconsistent logic to determine management ip address if the rabbitmq::management_ip_address parameter is not set, and makes sure all logic tries to look up the rabbitmq::node_ip_address parameter.

N/A


wyardley (approved):
LGTM; thanks for the contribution.

PR #915, opened 2022-10-02
modulesync 7.5.0

bastelfreak:
modulesync 5.3.0

PR #927, opened 2023-04-28
Drop Puppet 6 support

bastelfreak:


wyardley:
@bastelfreak are the non-preexisting CI failures expected along with this?
Also, do you know what's going on with #915 failures?

wyardley:
> I'm not sure what you mean with "non-preexisting CI failures".

@bastelfreak while the arch integration test failure was preexisting (though IMO, needs to get fixed or dropped), this PR also had failures on

Puppet / 7 - Debian 9
https://github.com/voxpupuli/puppet-rabbitmq/actions/runs/4831215814/jobs/8608377691

Puppet / 7 - Ubuntu 16.04
https://github.com/voxpupuli/puppet-rabbitmq/actions/runs/4831215814/jobs/8608377999

which I don't think were preexisting. So unless they're transient, I think this PR was merged with failing tests?

wyardley:
```
BeakerHostGenerator::Exceptions::InvalidNodeSpecError:
  Invalid node_info token: ubuntu1604
```

```
BeakerHostGenerator::Exceptions::InvalidNodeSpecError:
  Invalid node_info token: debian9
```

It's possible that this is just because the project needs a modulesync? I don't have enough context on that error to know, but since the open modulesync PR is also broken, not sure what the right fix is here?

bastelfreak:
ah right. we had a few CI errors in some of the PRs. I had to cleanup ~150 modules and having puppet 6 in metadata.json blocked the whole CI pipeline. We need to drop the EoL debian 9 and Ubuntu 16.04, but I didn't want to do that in the same PR.

wyardley:
Got it! Ok, makes sense! So there will be a separate one inbound for that, right?

bastelfreak:
when I find time I will clean it up in all modules. I also need to do a modulesync. Both require a lot of time. If anybody wants to help please join or IRC channel or slack channel so we can coordinate this.

wyardley:
I'm not doing much of anything Puppet related these days, and just help out here because no one else is really working on maintaining this module. So one issue is that until #926 and some related stuff is fixed, I don't think this module yet will work on the _non_ deprecated versions of these OSes 😞 
While the module will technically work on newer versions of things, assuming Erlang itself is installed / bootstrapped properly, I don't think we can properly integration test it on them yet until some other problems get sorted out by someone.

I could drop Ubuntu 16.04, but In particular, 9 is the only version of Debian currently marked as supported.

PR #928, opened 2023-05-10
Remove support for Debian 9 and Ubuntu 16.04; Add Debian 10 support

### LINKED ISSUE #887: Add support for debian 10
Debian 10 ("buster") ist the stable version of debian since 2019 but not listed in metadata.json and no testing is done with this version.

### LINKED ISSUE #930: Idempotency issue with implicitly enabled plugins
rabbitmq_plugin` was retrieving a list of existing plugins by querying for "explicitly" enabled plugins using the `rabbitmq-plugin -E` command. This caused us issues when adding support for CentOS/RHEL8 where we noticed that in this mode, ordering of plugins matters. 

Example of the scenario we saw:

In this spec test: https://github.com/voxpupuli/puppet-rabbitmq/blob/master/spec/acceptance/parameter_spec.rb#L18

The following plugins are declared:
```puppet
rabbitmq_plugin { [ 'rabbitmq_federation_management', 'rabbitmq_federation' ]:
  ensure => present
}
```

The `rabbitmq_federation_management` plugin depends on `rabbitmq_federation`, so when our code enables `rabbitmq_federation_management` it will automatically enable `rabbitmq_federation`, but mark it as being "implicitly" enabled because it was enabled as a result of enabling `rabbitmq_federation_management`.

Then, when go to query for the list of enabled plugins using the `-E` option, this only lists our "explicitly" enabled plugins, so `rabbitmq_federation` is never listed and every time we run Puppet it attempts to enable the plugin, breaking idempotency.

Copied this text from @nmaludy's #844, but creating an issue as well, just for tracking purposes.

This also implements a different fix for what was originally fixed in #909

wyardley:
- Deprecate Debian 9 and Ubuntu 16.04
- Add support for Debian 10
- Minor test updates (remove `RABBITMQ_USE_LONGNAME` in clustering acceptance tests)
- Fix rabbitmq_plugin to correctly detect implicitly enabled plugins (Cherry-picked from #844)

Closes #844
Fixes #930
Fixes #887

### DIFF
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -6,13 +6,30 @@ require File.expand_path(File.join(File.dirname(__FILE__), '..', 'rabbitmq_cli')
 Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Provider::RabbitmqCli) do
   confine feature: :posix
 
-  def self.instances
-    plugin_list = run_with_retries do
-      rabbitmqplugins('list', '-E', '-m')
+  def self.plugin_list
+    list_str = run_with_retries do
+      # Pass in -e to list both implicitly and explicitly enabled plugins.
+      # If you pass in -E instead, then only explicitly enabled plugins are listed.
+      # Implicitly enabled plugins are those that were enabled as a dependency of another plugin/
+      # If we do not pass in -e then the order if plugin installation matters within the puppet
+      # code. Example, if Plugin A depends on Plugin B and we install Plugin B first it will
+      # implicitly enable Plugin A. Then when we go to run Puppet a second time without the
+      # -e parameter, we won't see Plugin A as being enabled so we'll try to install it again.
+      # To preserve idempotency we should get all enabled plugins regardless of implicitly or
+      # explicitly enabled.
+      rabbitmqplugins('list', '-e', '-m')
     end
+    # Split by newline.
+    lines = list_str.split(%r{\n})
+    # Return only lines that are single words because sometimes RabbitMQ likes to output
+    # information messages. Suppressing those messages via CLI flags is inconsistent between
+    # versions, so this this regex removes those message without having to use painful
+    # version switches.
+    lines.grep(%r{^(\S+)$})
+  end
 
-    plugin_list.split(%r{\n}).map do |line|
-      next if line.start_with?('Listing plugins')
+  def self.instances
+    plugin_list.map do |line|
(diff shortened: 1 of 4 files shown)


wyardley:
@bastelfreak actually, this is not as bad as I thought. acceptance tests basically pass - I can see if I can figure out an explanation / fix for the one failing test.
Let me know if there's a better way to handle the title / description for both removing and adding support for versions in the same PR, or if it needs to be done in two phases.

```
  1) rabbitmq parameter on a vhost: create parameter resource runs successfully
     Failure/Error: apply_manifest(pp, catch_changes: true)
     Beaker::Host::CommandFailure:
       Host 'debian10-64-puppet7.example.com' exited with 2 running:
        puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest_195729764.pp.4YjFap
       Last 10 lines of output were:
       	Info: Loading facts
       	Info: Loading facts
       	Notice: Compiled catalog for debian10-64-puppet7.example.com in environment production in 0.29 seconds
       	Info: Using environment 'production'
       	Info: Applying configuration version '1683748656'
       	Notice: /Stage[main]/Main/Rabbitmq_plugin[rabbitmq_federation]/ensure: created
       	Info: /Stage[main]/Main/Rabbitmq_plugin[rabbitmq_federation]: Scheduling refresh of Service[rabbitmq-server]
       	Notice: /Stage[main]/Rabbitmq::Service/Service[rabbitmq-server]: Triggered 'refresh' from 1 event
       	Info: Calling prefetch: documentumFed@fedhost
       	Notice: Applied catalog in 18.74 seconds
```

bastelfreak:
ah nice, thanks! I think the title is fine.

bastelfreak (approved):

wyardley:
Worse, it doesn't seem to be totally consistent what fails 🤦

wyardley:
Ubuntu 22 has a lot more failures; going to leave that out for now.
If anyone on IRC (or any folks who have been more involved recently with Vox / Puppet stuff) has time to try and fix modulesync, that would also be super helpful, because I am wondering if that might help make sure all the integration test scaffolding is up to date.

wyardley:
Ok, pushed a test fix for some clustering related stuff.

The other stubborn one is the "create parameter resource" one, where the test itself works fine, but the idempotency check (no changes on second run) fails. Going to look at that one.
```
  Info: Applying configuration version '1683775408'
  Notice: /Stage[main]/Main/Rabbitmq_plugin[rabbitmq_federation]/ensure: created
```

The issue w/ Ubuntu 20.04 seems to be related to Python installation seemingly needing to run each time.

wyardley:
It looks like the main issue is due to the newer RabbitMQ version and the issue outlined in #844

wyardley:
I gave up on Debian 11 / Ubuntu 20 for today. Someone can sort those out after some of the rest of this gets out.

bastelfreak (approved):

PR #932, opened 2023-05-11
Support FreeBSD 12 and 13

wyardley:
Update tests and metadata to support FreeBSD 12 and 13 This also helps us unit test the (otherwise untested) conditionals related to systems without systemd

We may still need to tweak the current python package name, but this is a start

Per @bastelfreak's suggestion, use `os_facts` to avoid confusion with 
`:facts`

### DIFF
--- a/spec/classes/rabbitmq_spec.rb
+++ b/spec/classes/rabbitmq_spec.rb
@@ -6,18 +6,25 @@
 require 'spec_helper'
 
 describe 'rabbitmq' do
-  on_supported_os.each do |os, facts|
+  on_supported_os.each do |os, os_facts|
     context "on #{os}" do
       let :facts do
-        facts
+        os_facts
       end
 
-      name = case facts[:osfamily]
+      name = case os_facts[:osfamily]
              when 'Archlinux', 'OpenBSD', 'FreeBSD'
                'rabbitmq'
              else
                'rabbitmq-server'
              end
+
+      rabbitmq_home = case os_facts[:osfamily]
+                      when 'FreeBSD'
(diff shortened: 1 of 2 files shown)


spec/classes/rabbitmq_spec.rb
```diff
@@ -227,8 +234,8 @@
             is_expected.to contain_archive('rabbitmqadmin').with_source('http://1.1.1.1:15672/cli/rabbitmqadmin')
           end
 
-          it { is_expected.to contain_package('python') } if %w[RedHat Debian SUSE Archlinux].include?(facts[:os]['family'])
-          it { is_expected.to contain_package('python2') } if %w[FreeBSD OpenBSD].include?(facts[:os]['family'])
+          it { is_expected.to contain_package('python') } if %w[RedHat Debian SUSE Archlinux].include?(os_facts[:os]['family'])
+          it { is_expected.to contain_package('python38') } if %w[FreeBSD].include?(os_facts[:os]['family'])
```
wyardley, line 238: note: this may have changed by now - not sure. but at least it matches the current stuff in data, and can be adjusted in tests later if needed

spec/classes/rabbitmq_spec.rb
```diff
@@ -6,27 +6,34 @@
 require 'spec_helper'
 
 describe 'rabbitmq' do
-  on_supported_os.each do |os, facts|
+  on_supported_os.each do |os, os_facts|
     context "on #{os}" do
       let :facts do
-        facts
+        os_facts
```
wyardley, line 12: @bastelfreak this (and associated changes) is what you had in mind, right?

PR #933, opened 2023-05-11
BREAKING: Drop official Archlinux support

wyardley:
- Drop official metadata.json support for Archlinux until the acceptance tests can be fixed
- Replace legacy :osfamily in spec tests

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -47,9 +47,6 @@
     },
     {
       "operatingsystem": "OpenBSD"
-    },
-    {
-      "operatingsystem": "Archlinux"
     }
   ],
   "requirements": [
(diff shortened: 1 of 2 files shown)


wyardley (commented):
I'm thinking to leave the spec tests in place for now vs. ripping them out? There's some risk of drift or problems from them being there but not run, but on the other hand, would be a pain to add them back later.

Not sure if there's a way with the current Vox tooling to not run the acceptance tests but still run the spec tests or if that would be worth doing...

bastelfreak (approved):
I think we can leave the tests in and hope that I will find the time in the future to fix it.

PR #939, opened 2023-07-12
RabbitMQ 3.8.2+ compatibility for rabbitmqctl list_users

mika:
rabbitmq-server's rabbitmqctl as of v3.7.8-4 (as shipped with Debian/buster) used to behave as follows:

```
| % sudo rabbitmqctl -q list_users
| foo        [bar]
```

As of rabbitmq-server v3.8.2-1+deb10u1 (as present in buster-security, see https://packages.qa.debian.org/r/rabbitmq-server/news/20230711T122021Z.html), it instead behaves as follows:

```
| % sudo rabbitmqctl -q list_users
| user    tags
| foo        [bar]
```

This results in:

```
| Error: Could not prefetch rabbitmq_user provider 'rabbitmqctl': Cannot parse invalid user line: user    tags
```

There's a new `--no-table-headers` option available in more recent versions of RabbitMQ (which would also be available for rabbitmq-server v3.8.2-1+deb10u1), also see https://github.com/rabbitmq/rabbitmq-cli/issues/273.

But since don't want to rely on having this option available (and therefore breaking compatibility with e.g. v3.7.8-4), let's instead skip lines matching this header output, to remaing both backwards as well as forwards compatible.

Related to https://github.com/voxpupuli/puppet-rabbitmq/issues/740

FTR: tested/verified with rabbitmq-server v3.7.8-4 + v3.8.2-1+deb10u1

### DIFF
--- a/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
+++ b/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
@@ -17,7 +17,7 @@ Puppet::Type.type(:rabbitmq_user).provide(
       rabbitmqctl_list('users')
     end
 
-    user_list.split(%r{\n}).map do |line|
+    user_list.split(%r{\n}).reject { |line| line.strip =~ %r{^user\s+tags$} }.map do |line|
       raise Puppet::Error, "Cannot parse invalid user line: #{line}" unless line =~ %r{^(\S+)\s+\[(.*?)\]$}
 
       user = Regexp.last_match(1)


wyardley (commented):
Can you add a test case for this?

If that regex `^` is anchored to the start of any line, may be smart to also only ignore it if that pattern is found on the first line vs. on any line? I am guessing it would be rare for someone to create a user with the name "user" and the tag "tags", but.... better safe than sorry?

Probably it makes sense soon to drop support for older versions of rabbitmq, which would let the module make use of newer features (or, better, if someone had time, rewrite the module to use the API directly).

mika:
Gna sorry, I missed that new `rabbitmqctl_list` as of 42d45ed1667 already checks for the according package version and uses `--no-table-headers` as needed, we used an older version yet. Sorry for the noise and unneeded PR.

wyardley:
Thanks! Got it, yeah, thought I'd seen a PR for this, but hadn't come across it.

mika:
> Thanks! Got it, yeah, thought I'd seen a PR for this, but hadn't come across it.

Yeah sorry, while looking into adding a test case I noticed that we were missing latest changes of puppet-rabbitmq, latest master works perfectly fine for both v3.7.8-4 and v3.8.2-1+deb10u1. :partying_face: 

Incredibly fast reply from your side, kudos! Thanks and sorry again :)

PR #944, opened 2023-10-26
fix purge rabbitmq_parameter

fatpat:
purging shovels was not working since #832.

I don't know why, but `resource[:component_name]` is an empty string when calling `clear_parameter`. Using the instance variable `component name` works.

I don't know why the `resource object` is used to get instance variable in an instance function.

I just change component name to discuss about this bug, but for consistency, other variables in this function should be changed:
- `resource[:name]` -> `name`
- `resource[:value]` -> `value`


fatpat:
nervermind, wrong fix, wrong analyze

PR #937, opened 2023-07-01
puppetlabs/stdlib: Allow 9.x

bastelfreak:

PR #952, opened 2023-10-31
puppet/archive: Allow 7.x

bastelfreak:
contains #937


wyardley:
Cherry picked the last commit from #954 (26dc83b) to see if it works

wyardley:
Ok, yeah, that doesn't work either (well, acceptance tests fail), because of the `puppet-erlang` version (see #955) barfing on methods that are no longer in stdlib.

wyardley (approved):

PR #953, opened 2023-10-31
puppet/systemd: Alow 5.x & 6.x

bastelfreak:
also contains #952 and #937

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -58,7 +58,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.25.0 < 9.0.0"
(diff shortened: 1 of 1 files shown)


maureen-lawless:
@bastelfreak, any idea when this pull request will be released?
The older versions of stdlib, and systemd supported by puppet-rabbitmq do not officially support puppet 8.
So this looks like an issue if one wants to move to puppet 8 and use the puppet-rabbmiq module.

bastelfreak:
see the other response and https://github.com/voxpupuli/puppet-rabbitmq/pull/955 . I curently don't use the module and don't have the capacity to work on it. Feel free to pick it up.

wyardley:
Closing for #977

PR #962, opened 2023-11-09
add a workaround for running with --noop or --tags

### LINKED ISSUE #961: resources fail to prefetch when rabbitmq is not intended to be installed (via --noop or --tags)

- Puppet: 7.14.0, 7.23.0, and probably other versions
- Ruby: vendored ruby included with puppet
- Distribution: CentOS 7.9, AlmaLinux 8.7, Debian Sid
- Module version: current git

```
mkdir modules
git clone https://github.com/voxpupuli/puppet-rabbitmq modules/rabbitmq
git clone https://github.com/puppetlabs/puppetlabs-stdlib modules/stdlib
git clone https://github.com/voxpupuli/puppet-systemd.git modules/systemd
git clone https://github.com/voxpupuli/puppet-archive modules/archive

cat > rabbit.pp << EOF
class { 'rabbitmq':
    erlang_cookie => 'asdf',
}

rabbitmq_user { 'foo':
    password => 'bar',
}

notify { 'foo':
    message => 'bar',
    require => Rabbitmq_user['foo'],
    tag     => 'message',
}
EOF

sudo puppet apply --modulepath modules rabbit.pp --tags message
```

## What are you seeing

```
Error: Could not prefetch rabbitmq_user provider 'rabbitmqctl': Command rabbitmqctl is missing
```

Since rabbitmq resources do not have the `message`, tag, they should neither be applied nor fail.

## Output log

```
Warning: This function is deprecated, please use stdlib::ensure_packages instead. at ["/home/chickey/modules/rabbitmq/manifests/install/rabbitmqadmin.pp", 14]:["/home/chickey/modules/rabbitmq/manifests/init.pp", 567]
   (location: /home/chickey/modules/stdlib/lib/puppet/functions/deprecation.rb:35:in `deprecation')
Notice: Compiled catalog for <redacted> in environment production in 0.41 seconds
Error: Could not prefetch rabbitmq_user provider 'rabbitmqctl': Command rabbitmqctl is missing
Notice: /Stage[main]/Main/Notify[foo]: Dependency Rabbitmq_user[foo] has failures: true
Warning: /Stage[main]/Main/Notify[foo]: Skipping because of failed dependencies
Notice: Applied catalog in 0.55 seconds
```

I will submit a draft PR with a POC workaround for this. The commit will include a more technical explanation. I don't really know what a proper fix would be.

Original discussion in puppet-community slack:
https://puppetcommunity.slack.com/archives/C0W298S9G/p1699468482674569

Thanks,
Corey

bugfood:

This is a POC workaround for #961. See the commit message for technical details.

I don't know if this is the best fix or not. There's probably a better way.

Fixes #961

### DIFF
--- a/lib/puppet/provider/rabbitmq_cli.rb
+++ b/lib/puppet/provider/rabbitmq_cli.rb
@@ -54,6 +54,14 @@ class Puppet::Provider::RabbitmqCli < Puppet::Provider
         ['-q']
       end
     rabbitmqctl("list_#{resource}", *list_opts, *opts)
+  rescue Puppet::MissingCommand
+    # rabbitmqctl is not present. Normally we would have installed a package
+    # that provides rabbitmqctl by now, but if we're running under --noop or
+    # with a restrictive set of tags, the package may not have been installed.
+    # Return an empty string to avoid error. This may give false positives for
+    # resources under --noop!
+    Puppet.debug('rabbitmqctl command not found; assuming rabbitmq is not installed')
+    ''
   end
 
   def self.rabbitmq_running
--- a/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
+++ b/lib/puppet/provider/rabbitmq_plugin/rabbitmqplugins.rb
@@ -18,6 +18,11 @@ Puppet::Type.type(:rabbitmq_plugin).provide(:rabbitmqplugins, parent: Puppet::Pr
       # To preserve idempotency we should get all enabled plugins regardless of implicitly or
       # explicitly enabled.
       rabbitmqplugins('list', '-e', '-m')
+    rescue Puppet::MissingCommand
+      # See note about Puppet::MissingCommand in:
+      # lib/puppet/provider/rabbitmq_cli.rb
+      Puppet.debug('rabbitmqplugins command not found; assuming rabbitmq is not installed')
+      ''
     end
     # Split by newline.
     lines = list_str.split(%r{\n})


wyardley (commented):
Won’t this have the effect of suppressing failures when _not_ running as a noop as well?

bugfood:
> Won’t this have the effect of suppressing failures when _not_ running as a noop as well?

If it does, I don't know how to make that happen. My patch only applies to the listing operations, not to the actual entity-creation operations.

I can do the following, with my patch in place:
1. Set up rabbitmq normally.
2. Manually remove all instances of `rabbitmqctl` from the path.
3. Run puppet again.

I observe that this fails as expected. Here's an excerpt from the debug output:
```
Debug: Prefetching rabbitmqctl resources for rabbitmq_user
Debug: rabbitmqctl command not found; assuming rabbitmq is not installed
Debug: Command succeeded
[...]
Error: Command rabbitmqctl is missing
Error: /Stage[main]/[redacted]/Rabbitmq_user[admin]/ensure: change from 'absent' to 'present' failed: Command rabbitmqctl is missing (corrective)
```

-Corey

bugfood:
Thank you. I amended the PR and got checks to pass. Changes are:
* use single-quoted strings for rubocop
* add missing explicit empty-string return for rabbitmqplugins

bugfood:
> Yeah, I'm good with this solution, but do try to get some other folks to approve / weigh in from the Slack discussions.

I asked in slack a couple times; @bastelfreak is deferring to you, and I didn't get any other responses.

bugfood:
Thank you.

PR #974, opened 2024-03-29
Harden codebase and add documentation stubs

zilchms:
Might be backwards incompatible due to codebase hardening. Hardening occurs in:
- repo::apt
- repo::rhel
- service
All of those are either changing String to String[1] (disallowing empty Strings) or add a String datatype to variables already having a default String value

### DIFF
--- a/.puppet-lint.rc
+++ b/.puppet-lint.rc
@@ -1,3 +1,1 @@
 --fail-on-warnings
---no-parameter_documentation-check
---no-parameter_types-check
--- a/.sync.yml
+++ b/.sync.yml
@@ -2,3 +2,7 @@
 spec/spec_helper.rb:
   spec_overrides: "require 'spec_helper_local'"
   mock_with: ':mocha'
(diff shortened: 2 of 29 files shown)


wyardley (approved):
Hi @zilchms 
Can you clarify what is being hardened and what the likely breaking changes might be? Are they related to features not available in certain Puppet versions, or something else?

wyardley:
Also, I'm thinking this should probably be flagged as an enhancement as well as a docs change?

zilchms:
Sure let me edit the description real quick :)

wyardley (approved):
LGTM 🙏 

Not sure if this will cause big conflicts with #926 -- @kajinamit - you may have to make some adjustments when you rebase.

wyardley:
I'm going to leave off the "backwards incompatible" label, since I don't think this should interfere with normal / expected use of the module; let me know if that's untrue, though.

PR #975, opened 2024-04-02
Use a more expressive method of rewriting values

ekohl:
This achieves almost the same, except that this version doesn't change the original value. If that's really desired, `transform_values!` can be used. It makes it clearer what's going on.

### DIFF
--- a/lib/puppet/type/rabbitmq_parameter.rb
+++ b/lib/puppet/type/rabbitmq_parameter.rb
@@ -106,9 +106,12 @@ Puppet::Type.newtype(:rabbitmq_parameter) do
   def munge_value(value)
     return value if value(:autoconvert) == :false
 
-    value.each do |k, v|
-      value[k] = v.to_i if v =~ %r{\A[-+]?[0-9]+\z}
+    value.transform_values do |v|
+      if v.is_a?(String) && v.match?(%r{\A[-+]?[0-9]+\z})
+        v.to_i
+      else
+        v
+      end
     end
-    value
   end
 end
--- a/spec/unit/puppet/type/rabbitmq_parameter_spec.rb
+++ b/spec/unit/puppet/type/rabbitmq_parameter_spec.rb
@@ -44,9 +44,8 @@ describe Puppet::Type.type(:rabbitmq_parameter) do
   end
 
   it 'accepts a valid hash for value' do
-    value = { 'message-ttl' => '1800000' }
-    parameter[:value] = value
-    expect(parameter[:value]).to eq(value)
+    parameter[:value] = { 'message-ttl' => '1800000' }
+    expect(parameter[:value]).to eq({ 'message-ttl' => 1_800_000 })
   end
 
   it 'does not accept an empty string for definition' do


lib/puppet/type/rabbitmq_parameter.rb
```diff
@@ -106,9 +106,6 @@ def validate_value(value)
   def munge_value(value)
     return value if value(:autoconvert) == :false
 
-    value.each do |k, v|
-      value[k] = v.to_i if v =~ %r{\A[-+]?[0-9]+\z}
-    end
-    value
+    value.transform_values { |v| v.match?(%r{\A[-+]?[0-9]+\z}) ? v.to_i : v }
```
ekohl, line 109: Or is this better
```suggestion
    value.transform_values do |v|
      if v.match?(%r{\A[-+]?[0-9]+\z})
        v.to_i
      else
        v
      end
    end
```

lib/puppet/type/rabbitmq_parameter.rb
```diff
@@ -106,9 +106,6 @@ def validate_value(value)
   def munge_value(value)
     return value if value(:autoconvert) == :false
 
-    value.each do |k, v|
-      value[k] = v.to_i if v =~ %r{\A[-+]?[0-9]+\z}
```
ekohl, line 110: This would fail on Ruby 3.2 where `[] =~ //` is no longer defined, breaking Puppet 8. That's why I switched over to the string check. I also use `match?` because it's faster if you don't need the resulting match.

Valantin: but now

Failure/Error: expect(parameter[:value]).to eq(value)

  expected: {"message-ttl"=>"1800000"}
       got: {"message-ttl"=>1800000}


as said I've add next if array on the value iteration and ti "fix" the test on puppet 8

ekohl: > expected: {"message-ttl"=>"1800000"}
> got: {"message-ttl"=>1800000}

That's a side effect of the change. Previously the passed in value was modified in place and the tests (incorrectly) relied on that. I've modified that.

> as said I've add next if array on the value iteration and ti "fix" the test on puppet 8

I think that's wrong. Your code can still fail on other non-string and non-array types because https://bugs.ruby-lang.org/issues/15231 changed the behavior.

```text
[1] pry(main)> 123 =~ /regex/
NoMethodError: undefined method `=~' for 123:Integer
from (pry):1:in `__pry__'
[2] pry(main)> {} =~ /regex/
NoMethodError: undefined method `=~' for {}:Hash
from (pry):2:in `__pry__'
```

That's why I explicitly chose `is_a?(String)`.

You could write `/regex/.match?(v)` but you'd get a TypeError if you pass in a non-string:
```
[3] pry(main)> /1234/.match?('123')
=> false
[4] pry(main)> /1234/.match?([])
TypeError: no implicit conversion of Array into String
from (pry):4:in `match?'
```

wyardley (approved):

lib/puppet/type/rabbitmq_parameter.rb
wyardley: Personally I think that’s a little more readable than the terser syntax but not a strong preference, and don’t claim to be a Ruby expert.

ekohl: Updated

PR #976, opened 2024-04-02
Fix Test

### LINKED ISSUE #965: puppet-archive version
Is it possible to bump the puppet-archive version to 7.1.0. Should i create a pull request?

- Puppet:
- Ruby:
- Distribution:
- Module version:

Valantin:
initial refactoring of acceptance
fix test failing on puppet 8
remove unused erlang module from fixture
bump dependency version

Fixes #965


wyardley:
Thanks for taking a look at this.

Just a heads up before you get too deep into this -- I'm not sure if this will be easy to get working without #926 or something like it.

Also, the issue in some of the failing tests with the cluster being reported as `rabbitmq` vs `rabbitmq_cluster` is odd (and showed up in some other recent PRs as well) - I took a quick look at it and seems like the test explicitly specifies the cluster name as `rabbitmq_cluster`, so the most likely reason is that since we're still using the Linux distribution's rmq version in the tests that this behavior may have changed in a newer version.

I haven't had luck getting the acceptance tests for puppet to work locally in a while (can get them to start and then certain OSes just crash outright on my machine using the Docker version of the beaker tests), so it's been hard to figure out some of the issues that are ongoing.

As far as the erlang module, there was a goal of switching to the vox module in #955 and that's been also stalled by some of these continuing / circular issues.

I don't use puppet or rabbitmq and haven't for quite a few years now, but I'm happy to provide any help that I can in terms of helping move any of these ongoing efforts forward.

Valantin:
Thanks for the resources to start.
Idealy OS package was self consistent, so the erlang version shipped was the correct one for the rabbimq-server version.

The erlang module was used only on acceptance test, I've already use it on on ubuntu 22.04 in my production cloud, and I hate the repository because in one year the have dropped support for erlang version I use, and I've updated all my cluster to add a new node. Not so happy in a production environment.

My idea was to fix acceptance using the os package (if possible) to add new OS support
For example no "official" external repos was available for debian 12 at this moment, but i'm pretty sure you can run correctly with the system version.

manifests/config.pp
```diff
@@ -234,7 +234,7 @@
     default: {}
   }
 
-  if $facts['systemd'] { # systemd fact provided by systemd module
+  if $facts['service_provider'] == 'systemd' { # systemd fact provided by systemd module
```
ekohl, line 237: This shouldn't be needed since the module depends on the systemd module. If anything, the comment should be updated.

Valantin: Ok, try to revert to facts[systemd], if don't work remove the comment.

Valantin:
@bastelfreak I've rebased on modulesync branch, hope Is a better solution to use the modulesync 7.3 test

wyardley:
Hi @Valantin 
Are you still working on this?

Valantin:
Yes and no, I'm switched to other things.
But in my todolist, I need more time to check OS related things, for example the EL approch in 8 and 9 release.

wyardley:
Probably not needed with #985 etc

PR #981, opened 2024-05-19
Remove leftovers from Linux systems without systemd

bastelfreak:
to implement puppet 8 support I need to update the lower version of the puppet/systemd module because systemd::service_limits is deprecated and we need to use systemd::manage_dropin. And that change is easier when we drop the dead code paths.


wyardley (approved):

wyardley:
I assume we're considering this a non-breaking change, right? Do you want this to be highlighted in the changelog?

bastelfreak:
we already dropped support for those OSes, so IMO it's not a breaking change.

PR #982, opened 2024-05-19
systemd: migrate from service_limits->manage_dropin

bastelfreak:


wyardley (approved):

wyardley:
I assume we're not flagging this one as breaking / backwards-incompatible since the low limit of systemd is so old?

PR #979, opened 2024-05-18
Release 13.3.0

wyardley:
Includes

- #978 
- #975 
- #977 
- #974 
- #969 
- #980
- #915 
- #981 
- #982

PR #983, opened 2024-05-19
CentOS7: default to EPEL7 as source

bastelfreak:
Historically we used the garethr/erlang module as soft dependency on CentOS 7. This also configured the EPEL7 repository. I always always a bit unsure where the actual rabbitmq packages come from. I assumed somehow from the erlang repo. My impression is that the erlan repo isn't used at all. rabbitmq packages where always pulled from EPEL7. The problem with the garethr/erlang is that it depends on the deprecated `stahnma/epel` (which is now puppet/epel) and uses topscope facts and
variables. The module doesn't work anymore with Puppet 8.

This PR replaces the unused garethr/erlang directly with the EPEL module.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -492,6 +492,8 @@ class rabbitmq (
     }
   }
 
+  # when repos_ensure is true, we configure externel repos
+  # CentOS 7 doesn't contain rabbitmq. It's only in EPEL.
   if $repos_ensure {
     case $facts['os']['family'] {
       'RedHat': {
@@ -505,6 +507,8 @@ class rabbitmq (
       default: {
       }
     }
+  } elsif ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '7') {
+    require epel
   }
 
   contain rabbitmq::install
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,8 +1,9 @@
+---
 fixtures:
   repositories:
     stdlib: 'https://github.com/puppetlabs/puppetlabs-stdlib'
     apt: 'https://github.com/puppetlabs/puppetlabs-apt'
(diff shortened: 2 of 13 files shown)


manifests/init.pp
```diff
@@ -505,6 +507,8 @@
       default: {
       }
     }
+  } elsif ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '7') {
+    require epel
```
bastelfreak, line 511: I don't have a strong opinion on how we should call the epel module. We could also just add it to the acceptance tests or introduce another parameter for it.

wyardley:
The actual rmq packages used to come from packagecloud, and then we switched to them coming from the OS vendor repo. They did not ever come from the Erlang repo, but I believe they had a dependency on some version of Erlang being installed that perhaps the package itself did not satisfy? This was ancient history, though.

this makes sense to me if it works, though may make sense to ship the other, safer changes first since there are a bunch.

wyardley (approved):

TheMeier (approved):

bastelfreak:
@wyardley you want to release 13.3.0 first, then merge #983 and do another release?

wyardley:
If #979 looks good to you as-is, let's do that; also, is it possible that CentOS 8 (see #900) will now work, though based on https://en.wikipedia.org/wiki/CentOS, seems like CentOS 8 actually was already EOLed a long time ago, and 7 will be next month?

bastelfreak:
We already started dropping CentOS 7 in some repos (https://voxpupuli.org/blog/2024/05/08/CentOS-7-CentOS-Stream-8-deprecation/) but it would be nice to have one release where it's actually fixed. And the updated dependencies will make it easier for users to upgrade.

PR #986, opened 2024-05-19
puppetlabs/stdlib: Allow 9.x

bastelfreak:


wyardley (approved):

wyardley (commented):
Note: from my experiments yesterday, I think there might be some previously deprecated methods that are fully dropped? So we might want to do this as a breaking change and update any removed or deprecated methods at the same time?

bastelfreak:
that's why I want to get the puppet 8 PR in. CI will fail on puppet 8 when we use deprecated functions.

wyardley:
Seems like this one passes now, though? With or without Puppet 8, does it make sense to drop support for older stdlib as a breaking change (with removal of those methods)? I'm good either way - if you think this one is good, let's :shipit:

wyardley:
I noticed that the GH UI isn't setup to suggest / warn when the base needs to be updated if there are no conflicts - would it make sense to add that in the branch protection settings if someone's got that power?

TheMeier (approved):

PR #990, opened 2024-05-19
Use ensure_packages from stdlib

wyardley:
Resolve deprecation errors by using `ensure_packages()` from stdlib vs. from Puppet

```
  Warning: This function is deprecated, please use stdlib::ensure_packages instead. at ["/etc/puppetlabs/code/environments/production/modules/rabbitmq/manifests/install/rabbitmqadmin.pp", 14]:["/etc/puppetlabs/code/environments/production/modules/rabbitmq/manifests/init.pp", 574]
```


wyardley:
@bastelfreak does this make sense? seems like `ensure_packages` existed even in the oldest version we support?

So far, I didn't see any other deprecated stdlib methods? Not sure if you'd like to increase the min version for it anyway as a breaking change?

TheMeier:
replacing ensure_packages by stdlib::ensure_packages makes perfectly sense

TheMeier (approved):

PR #996, opened 2024-05-20
Revert hard require on epel for CentOS 7

### LINKED ISSUE #995: could not find class ::epel when upgrading to 13.4.0

- Puppet: 7
- Distribution: CentOS 7
- Module version: 13.4.0

```
class { 'rabbitmq': }
```

## What are you seeing
We're not setting `repos_ensure` and defaults to `false` still , so we were not expecting any repo management by the module. When upgrading to the latest version it seems to have a hard requirement on the `epel` module when we're already managing repositories outside of this. Is it possible to make this optional still so that `repos_ensure` or possibly some new option like` manage_repos` supersedes it entirely, and does not perform any repo management when it is not necessary?  For example, we are running on CentOS 7 but pulling packages from packagecloud

No errors when not expecting to have any repos managed by the rabbitmq module, or managing repos externally

## Output log
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Could not find class ::epel for testrmq.example.com (file: /etc/puppetlabs/code/puppet-control/production/modules/rabbitmq/manifests/init.pp, line: 511, column: 5) on node testrmq.example.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

wyardley:
This leaves some of the documentation about epel, and installs it for acceptance tests, but leaves it up to users on CentOS to include this separately if needed.

Fixes #995

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ These are now documented via [Puppet Strings](https://github.com/puppetlabs/pupp
 
 You can view example usage in [REFERENCE](REFERENCE.md).
 
-**[puppet/epel](https://forge.puppet.com/modules/puppet/epel) is a soft dependency. The module requires it if you're on CentOS 7**
+**[puppet/epel](https://forge.puppet.com/modules/puppet/epel) is a soft dependency. On CentOS, with `$repos_ensure` set to `false` (current default), it may be required.
 
 Version v13.2.0 and older also added an erlang repository on CentOS 7. That isn't used and can be safely removed.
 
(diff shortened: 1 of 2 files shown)


wyardley:
Note: acceptance tests will probably fail now - I will need to figure out how to include or require it in the tests, but not in the code, if we go this route vs. using a flag.

wyardley:
Assuming #997 is preferred then

PR #997, opened 2024-05-20
Add require_epel parameter, defaulting to `true`

### LINKED ISSUE #995: could not find class ::epel when upgrading to 13.4.0

wyardley:
Allow an option to restore epel as a true soft dependency. Defaulting to `true` so that the tests will work with default behavior.

IMO, the safer / better / not technically breaking behavior would be to default this to `false`, however I don't think it's possible / trivial to do this while keeping the default class inclusion test in 
https://github.com/voxpupuli/puppet-rabbitmq/blob/137edeedf1f15e04d239a026e119c114eb2021ba/spec/acceptance/class_spec.rb#L18
as-is? If we want to try instantiating it differently there, I can rework this or open a new one.

Since there wasn't already a unit test for this, I didn't add one here, but can do so if needed depending on which option we go with.

Fixes #995
Closes #998

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ These are now documented via [Puppet Strings](https://github.com/puppetlabs/pupp
 
 You can view example usage in [REFERENCE](REFERENCE.md).
 
-**[puppet/epel](https://forge.puppet.com/modules/puppet/epel) is a soft dependency. The module requires it if you're on CentOS 7**
+**[puppet/epel](https://forge.puppet.com/modules/puppet/epel) is a soft dependency. If you're on CentOS 7 and don't want to require it, set `$require_epel` to `false`**
 
 Version v13.2.0 and older also added an erlang repository on CentOS 7. That isn't used and can be safely removed.
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -243,6 +243,8 @@
 #   Defaults to false (use system packages). This does not ensure that soft dependencies (like EPEL on RHEL systems) are present.
 #   It also does not solve the erlang dependency.  See https://www.rabbitmq.com/which-erlang.html for a good breakdown of the
 #   different ways of handling the erlang deps.  See also https://github.com/voxpupuli/puppet-rabbitmq/issues/788
+# @param require_epel
+#   If this parameter is set, On CentOS / RHEL 7 systems, require the "puppet/epel" module
 # @param service_ensure
 #   The state of the service.
 # @param service_manage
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -49,9 +49,9 @@ These are now documented via [Puppet Strings](https://github.com/puppetlabs/pupp
 
 You can view example usage in [REFERENCE](REFERENCE.md).
 
-**[puppet/epel](https://forge.puppet.com/modules/puppet/epel) is a soft dependency. The module requires it if you're on CentOS 7**
+**[puppet/epel](https://forge.puppet.com/modules/puppet/epel) is a soft dependency. If you're on CentOS 7 and don't want to require it, set `$require_epel` to `false`**
 
-Version v13.2.0 and older also added an erlang repository on CentOS 7. That isn't used and can be safely removed.
+Version v13.2.0 and older also added an erlang module on CentOS 7. That isn't used and can be safely removed.
```
bastelfreak, line 54: I used `repository` to reference the file in `/etc/yum.repos.d/`

manifests/init.pp
```diff
@@ -243,6 +243,8 @@
 #   Defaults to false (use system packages). This does not ensure that soft dependencies (like EPEL on RHEL systems) are present.
 #   It also does not solve the erlang dependency.  See https://www.rabbitmq.com/which-erlang.html for a good breakdown of the
 #   different ways of handling the erlang deps.  See also https://github.com/voxpupuli/puppet-rabbitmq/issues/788
+# @param require_epel
+#   If this parameter is set, On CentOS / RHEL 7 systems, require the "epel" module
```
bastelfreak, line 247: maybe we should be a bit more precise which epel repo we use?
```suggestion
#   If this parameter is set, On CentOS / RHEL 7 systems, require the "puppet/epel" module
```

PR #1001, opened 2024-05-24
require puppet/systemd >= 6.0.0

saz:
With the changes made in https://github.com/voxpupuli/puppet-rabbitmq/pull/982, this module needs `puppet-systemd` >= 6.0.0, as this is the first release with the `Limits` moved to the `Systemd::Unit::Service` data type. Otherwise, it's not possible to set limits, as validation will fail

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -66,7 +66,7 @@
     },
     {
       "name": "puppet/systemd",
-      "version_requirement": ">= 4.0.0 < 8.0.0"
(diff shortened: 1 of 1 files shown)


wyardley (approved):
Approved. @bastelfreak should we flag this as backwards compatible? In a bit of a spot since the change went out with the last version.

wyardley:
Maybe we can add a release note specifying the last release that supports earlier systemd versions?

PR #1007, opened 2024-06-12
replace systemd fact with core fact

bastelfreak:

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -27,8 +27,5 @@ class rabbitmq::service (
       hasrestart => true,
       name       => $service_name,
     }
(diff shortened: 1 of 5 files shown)


manifests/service.pp
```diff
@@ -27,8 +27,5 @@
       hasrestart => true,
       name       => $service_name,
     }
-    if $facts['systemd'] and defined(Class['systemd::systemctl::daemon_reload']) {
-      Class['systemd::systemctl::daemon_reload'] -> Service['rabbitmq-server']
```
bastelfreak, line 31: the class doesn't exist anymore and also isn't needed.

wyardley (approved):

PR #1008, opened 2024-06-12
Add Debian 11

bastelfreak:


wyardley (approved):
I may have tried this already and it didn't work, but LGTM if tests pass 🚢

wyardley:
I think I sorted out the issue - I ended up opening a new one vs. editing / changing yours.

PR #1010, opened 2024-06-27
Add support for Debian 11

wyardley:
Add metadata support for Debian 11, and use the "python3" package for the entire Debian family. This appears to be valid for both currently supported versions

Closes #1008

### DIFF
--- a/data/family/Debian.yaml
+++ b/data/family/Debian.yaml
@@ -1,4 +1,5 @@
 ---
+rabbitmq::python_package: 'python3'
 rabbitmq::package_name: 'rabbitmq-server'
 rabbitmq::service_name: 'rabbitmq-server'
 rabbitmq::package_gpg_key: 'https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey'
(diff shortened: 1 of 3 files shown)


spec/classes/rabbitmq_spec.rb
```diff
@@ -190,7 +190,8 @@
             is_expected.to contain_archive('rabbitmqadmin').with_source('http://1.1.1.1:15672/cli/rabbitmqadmin')
           end
 
-          it { is_expected.to contain_package('python') } if %w[RedHat Debian SUSE Archlinux].include?(os_facts[:os]['family'])
+          it { is_expected.to contain_package('python') } if %w[RedHat SUSE Archlinux].include?(os_facts[:os]['family'])
+          it { is_expected.to contain_package('python3') } if %w[Debian].include?(os_facts[:os]['family'])
```
wyardley, line 194: I played with single quoting this, but decided to keep the existing style, even for ones where we only have 1 right now.

data/family/Debian.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+rabbitmq::python_package: 'python3'
```
wyardley, line 2: note: these are not alpha sorted; I tried to follow the convention / lack of convention from other files, but happy to reorder some or all of these if necessary.

PR #1013, opened 2024-09-04
Add support for cowboy_opts config in rabbitmq.config

SpinEternel:
Adds the possibility  to use cowboy_opts configuration in the rabbitmq.config.

Unfortunately, i can't test the ssl/cowboy_opts part on my side.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -10,6 +10,7 @@ class rabbitmq::config {
   $cluster_node_type                   = $rabbitmq::cluster_node_type
   $cluster_nodes                       = $rabbitmq::cluster_nodes
   $config                              = $rabbitmq::config
+  $config_cowboy_opts                  = $rabbitmq::config_cowboy_opts
   $config_cluster                      = $rabbitmq::config_cluster
   $config_path                         = $rabbitmq::config_path
   $config_ranch                        = $rabbitmq::config_ranch
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -10,8 +10,8 @@ class rabbitmq::config {
   $cluster_node_type                   = $rabbitmq::cluster_node_type
   $cluster_nodes                       = $rabbitmq::cluster_nodes
   $config                              = $rabbitmq::config
-  $config_cowboy_opts                  = $rabbitmq::config_cowboy_opts
   $config_cluster                      = $rabbitmq::config_cluster
+  $config_cowboy_opts                  = $rabbitmq::config_cowboy_opts
   $config_path                         = $rabbitmq::config_path
   $config_ranch                        = $rabbitmq::config_ranch
   $config_stomp                        = $rabbitmq::config_stomp
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -137,6 +137,8 @@
 #   Additional config variables in rabbitmq.config
 # @param config_cluster
 #   Enable or disable clustering support.
+# @param config_cowboy_opts
+#   Additional config variables for cowboy_opts in rabbitmq.config.
 # @param config_kernel_variables
 #   Hash of Erlang kernel configuration variables to set (see [Variables Configurable in rabbitmq.config](#variables-configurable-in-rabbitmq.config)).
 # @param config_path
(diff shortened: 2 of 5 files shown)


manifests/init.pp
```diff
@@ -137,6 +137,8 @@
 #   Additional config variables in rabbitmq.config
 # @param config_cluster
 #   Enable or disable clustering support.
+# @param config_cowboy_opts
+#   Additional config variables for cowboy_opts in rabbitmq.config.
```
wyardley, line 141: probably note that this is a hash of values, similar to `$config_kernel_variables` etc.

templates/rabbitmq.config.epp
```diff
@@ -164,7 +171,10 @@
   <%- if $rabbitmq::config::management_ip_address {-%>
       {ip, "<%= $rabbitmq::config::management_ip_address %>"},
   <%- } -%>
-      {port, <%= $rabbitmq::config::management_port %>}
+      {port, <%= $rabbitmq::config::management_port %>}<% if !$rabbitmq::config::config_cowboy_opts.empty {%>,
```
wyardley, line 174: Can this match the other one, with a newline after but then suppressing the newline in the template? Or does the open curly brace here need to be on the same line?

I think this will look wrong, but could be wrong.

Probably need to test both cases here, not sure if you'll need to specify any other params to be able to test both the SSL and non-SSL variant.

wyardley (commented):
Hi @SpinEternel 
Thanks for the contribution!

A few minor comments, also, this should add unit tests for the change you're adding (this will also at least _somewhat_ help with the issue you mentioned around testing the SSL variant).

I believe the tests should be something along the lines of the ones for, e.g., `config_kernel_variables`.
https://github.com/voxpupuli/puppet-rabbitmq/blob/4eed5e4eee563f64e6a22716eeaefc34a7acca8a/spec/classes/rabbitmq_spec.rb#L1587-L1600

[here](https://github.com/voxpupuli/puppet-rabbitmq/blob/4eed5e4eee563f64e6a22716eeaefc34a7acca8a/spec/classes/rabbitmq_spec.rb#L522-L523) is a rough example of how you can set other parameters to test different cases.

The regexes or string tests in the tests can be somewhat loose, but should be restrictive enough that they are useful.

wyardley (commented):
Also, you'll need the CI checks to pass. For the case of the outdated reference file, you'll need to run the rake task to update the strings. I'd also make sure the other checks (unit tests etc.) pass locally.

templates/rabbitmq.config.epp
SpinEternel: Still looking for a solution to make this more readable. For now this is the only way i found to not get this result :
```
-      {port, 15672},
+      {port, 15672}
+      ,
```

wyardley: got it... I'll take a look. may be possible to suppress the newline entirely with dashes, though I had missed that comma - this may actually be the best / most readable way to do it.

manifests/config.pp
```diff
@@ -10,6 +10,7 @@
   $cluster_node_type                   = $rabbitmq::cluster_node_type
   $cluster_nodes                       = $rabbitmq::cluster_nodes
   $config                              = $rabbitmq::config
+  $config_cowboy_opts                  = $rabbitmq::config_cowboy_opts
```
wyardley, line 13: nit: I'd move this down so that cl is alphabetical (before co)

templates/rabbitmq.config.epp
wyardley: I think it's good to leave this as-is. I can make a PR against your branch (side note: ideally, make your PR off a branch other than the default one, i.e., `main` or `master`, in your fork).

wyardley:
See https://github.com/SpinEternel/puppet-rabbitmq/pull/1

SpinEternel:
Thx a lot for the help ! I didn't have a lot of time to check all the tests. All have passed, sounds good ?

wyardley (approved):

wyardley:
Ack, sorry, meant to squish merge this one.

PR #1014, opened 2024-09-06
Switch to string vs. symbols for facts in spec tests

wyardley:
Set `facterdb_string_keys` to `true` and update to string versions of facts in our spec tests.

Update `.sync.yml` so that modulesync doesn't override this while the default is still `false`

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -10,7 +10,7 @@ ENV['COVERAGE'] ||= 'yes' if Dir.exist?(File.expand_path('../lib', __dir__))
 require 'voxpupuli/test/spec_helper'
 
 RSpec.configure do |c|
-  c.facterdb_string_keys = false
+  c.facterdb_string_keys = true
   c.mock_with :mocha
 end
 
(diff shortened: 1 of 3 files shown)


TheMeier (approved):

spec/spec_helper.rb
```diff
@@ -10,7 +10,6 @@
 require 'voxpupuli/test/spec_helper'
 
 RSpec.configure do |c|
-  c.facterdb_string_keys = false
```
bastelfreak, line 13: This Optik is coming from modulesync_config : https://github.com/voxpupuli/puppet-autofs/blob/master/.sync.yml#L13 https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/spec/spec_helper.rb.erb#L13

wyardley: Or just that this would get wiped out by a future modulesync and should just close this PR and leave it as it is?

wyardley: https://github.com/voxpupuli/modulesync_config/blob/78c09c9954deb7b2510a381a1dc750a9feba1855/config_defaults.yml#L115

Ah, default from here. Yeah, will just leave this as is

PR #1024, opened 2024-11-14
Add client_renegotiation ssl option, use more ssl options in management plugin

wyardley:
Rework of #906, credit to @Infraded

1. Adds the option for `client_renegotiation` to rabbitmq config
2. Copies multiple options from rabbitmq config to management plugin config similar to existing ssl settings
    - `client_renegotiation`
    - `secure_renegotiate`
    - `reuse_sessions`
    - `honor_cipher_order`
3. Wraps `client_renegotiation` and `secure_renegotiate` in a conditional to not include them when enabling TLSv1.3 as they are incompatible
4. Updates/add tests for `client_renegotation` setting and TLSv1.3 conditional

@Infraded I reworked your PR, since the conflicts still hadn't been resolved. This necessitated rewriting the template changes in epp.

Probably should be squash merged

Closes #906


data/common.yaml
```diff
@@ -62,6 +62,7 @@ rabbitmq::ssl_fail_if_no_peer_cert: false
 rabbitmq::ssl_management_verify: 'verify_none'
 rabbitmq::ssl_management_fail_if_no_peer_cert: false
 rabbitmq::ssl_versions: ~
+rabbitmq::ssl_client_renegotiation: ~
```
bastelfreak, line 65: We don't need this default value, because we've the same value in init.pp (which is also the preferred place, because puppt-strings will pick it up and document it in the REFERENCE.md)

wyardley: updated in second commit.

bastelfreak (approved):

PR #1032, opened 2024-12-10
add params rhel_satellite

flepoutre:
Hello,
The purpose of this PR is to not install the `centos-release-rabbitmq-38` package if we used RHEL satellite.
Can you please take it into account ?
Thanks.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -245,6 +245,8 @@
 #   Defaults to false (use system packages). This does not ensure that soft dependencies are present.
 #   It also does not solve the erlang dependency. See https://www.rabbitmq.com/which-erlang.html for a good breakdown of the
 #   different ways of handling the erlang deps. See also https://github.com/voxpupuli/puppet-rabbitmq/issues/788
+# @param rhel_satellite
+#   True if we used RedHat satellite
 # @param service_ensure
 #   The state of the service.
 # @param service_manage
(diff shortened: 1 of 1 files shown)


wyardley (commented):
@flepoutre: Thanks for this. I had an idea this might cause a problem, but because there's no working CI setup for acceptance tests that works for this module on RHEL (there is an image, but IIRC, there are non-available required packages that make practically using it for Puppet tests difficult), it was hard to know for sure, and also didn't know what the desired behavior would be in the case of RHEL.

High level, I think this is about the right fix, but I think we should do a combination of:
* A bit more magic internally
and / or
* Rename the parameter to something more consistent and generic (like `enable_centos_release` (bool)).

I think we're trying to do less of "data in modules", so the fix of setting the param default to `true` when the exact OS is CentOS, and false otherwise might not work.

But we can probably use a combination of facts and defaults to get this to a somewhat better place.

You'll also need to update the generated docs, and possibly adjust one or two spec tests.

jay7x:
Please update REFERENCE.md with `bundle exec rake strings:generate:reference`.

Also, parameter name is confusing. I'd say the whole logic there is confusing. From my point of view, it should be under `$repos_ensure` (which, in turn, should be named `$manage_repos` for consistency, but it's different topic). I'm not that good in RHELs, but I *guess* that package in question installs the repo actually. Forget this, if I'm wrong. Though the parameter should be renamed in any case.

wyardley:
This is a kind of wonky one. I implemented something that I think will resolve this in #1034 – hope that’s ok?

flepoutre:
@wyardley Perfect many thanks for => https://github.com/voxpupuli/puppet-rabbitmq/pull/1034  :)

PR #1054, opened 2025-06-19
Fix binding arguments parsing and uniqueness in rabbitmq_binding provider

### LINKED ISSUE #1053: rabbitmq_binding: arguments ignored due to regexp and uniqueness issues

- Puppet: 6.24.0
- Ruby: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]
- Distribution: Debian 12
- Module version: 14.3.0

```
  rabbitmq_vhost { '/myvhost':
    ensure => present,
  }
  rabbitmq_queue { 'myqueue@/myvhost':
    ensure  => present,
    durable => true,
  }
  rabbitmq_exchange { 'myexchange@/myvhost':
    ensure    => present,
    durable   => true,
    type      => 'direct',
    arguments => {},
  }
  rabbitmq_binding { 'myexchange@myqueue_value1@/myvhost':
    ensure           => present,
    source           => 'myexchange',
    destination      => 'myqueue',
    destination_type => 'queue',
    routing_key      => '',
    vhost            => '/myvhost',
    arguments        => { 'x-match' => 'all', 'header1' => 'value1' },
  }
  rabbitmq_binding { 'myexchange@myqueue_value11@/myvhost':
    ensure           => present,
    source           => 'myexchange',
    destination      => 'myqueue',
    destination_type => 'queue',
    routing_key      => '',
    vhost            => '/myvhost',
    arguments        => { 'x-match' => 'all', 'header1' => 'value11' },
  }

```

## What are you seeing

**Issue** : Existing arguments are not correctly retrieved from RabbitMQ due to a faulty regular expression in the provider code, which leads to incorrect parsing and comparison of binding arguments.
**Potential root cause** : The call of `gsub(%r{^\[(.*)\]$}, '')` removes the entire arguments string instead of extracting its content, resulting in arguments always being empty `{}` when parsed. This prevents Puppet from comparing or managing bindings with arguments properly. Replacing this with `gsub(%r{^\[|\]$}, '')` correctly strips only the square brackets and preserves the arguments content.

<details open>

<summary>Faulty regexp result logs</summary>

```
Notice: /Stage[main]/Main/Node[######]/Rabbitmq_binding[myexchange@myqueue_value1@/myvhost]/arguments: arguments changed {

} to {
  'x-match' => 'all',
  'header1' => 'value1'
}
```

</details>

**Issue** : Puppet also does not correctly manage the presence of multiple bindings with the same source, destination, vhost, and routing_key but different arguments.
**Potential root cause** : This second issue with arguments seems to come from the fact that the hash used to identify bindings `hashed_name` and the prefetch method do not take arguments into account.

<details open>

<summary>Hash and Prefetch without arguments</summary>

```
Notice: /Stage[main]/Main/Node[######]/Rabbitmq_binding[myexchange@myqueue_value1@/myvhost]/arguments: arguments changed {

} to {
  'x-match' => 'all',
  'header1' => 'value1'
} (corrective)
Notice: /Stage[main]/Main/Node[######]/Rabbitmq_binding[myexchange@myqueue_value11@/myvhost]/arguments: arguments changed {
  'x-match' => 'all',
  'header1' => 'value1'
} to {
  'x-match' => 'all',
  'header1' => 'value11'
} (corrective)
```

</details>

Both of these issues generate a perpetual diff at every Puppet run.

I expect to manage each binding uniquely, even if only the arguments differ.
Each binding with different arguments should be created and updated independently.

## Output log

See above for the Puppet notice logs showing the incorrect handling of arguments and bindings.

J-Mx:
See #1053 for a clear description of the issue. This PR fix arguments parsing and uniqueness while managing rabbitmq binding with puppet

Fixes #1053

### DIFF
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -43,7 +43,7 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin, parent: Puppet::Pro
           arguments = arguments.gsub(%r{^\[|\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
           arguments = '{}' if arguments == ''
         end
-        hashed_name = Digest::SHA256.hexdigest format('%s@%s@%s@%s', source_name, destination_name, vhost, routing_key)
+        hashed_name = Digest::SHA256.hexdigest format('%s@%s@%s@%s@%s', source_name, destination_name, vhost, routing_key, arguments)
         next if source_name.empty?
 
         binding = {
@@ -67,7 +67,7 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin, parent: Puppet::Pro
   def self.prefetch(resources)
     bindings = instances
     resources.each do |name, res|
-      if (provider = bindings.find { |binding| binding.source == res[:source] && binding.destination == res[:destination] && binding.vhost == res[:vhost] && binding.routing_key == res[:routing_key] })
+      if (provider = bindings.find { |binding| binding.source == res[:source] && binding.destination == res[:destination] && binding.vhost == res[:vhost] && binding.routing_key == res[:routing_key] && binding.arguments == res[:arguments] })
         resources[name].provider = provider
       end
     end
--- a/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
@@ -37,7 +37,10 @@ Puppet::Type.type(:rabbitmq_binding).provide(:rabbitmqadmin, parent: Puppet::Pro
         if arguments.nil?
           arguments = '{}'
         else
-          arguments = arguments.gsub(%r{^\[(.*)\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
+          # Substitution : Removes the opening '[' and closing ']' brackets from the arguments string
+          # Substitution 2 : Converts JSON-style "key", format to "key": format by replacing commas after quoted keys with colons
+          # Substitution 3 : Merges multiple object definitions by replacing "},{" with "," to create a single valid JSON object
+          arguments = arguments.gsub(%r{^\[|\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
           arguments = '{}' if arguments == ''
         end
         hashed_name = Digest::SHA256.hexdigest format('%s@%s@%s@%s', source_name, destination_name, vhost, routing_key)
--- a/lib/puppet/provider/rabbitmq_queue/rabbitmqadmin.rb
+++ b/lib/puppet/provider/rabbitmq_queue/rabbitmqadmin.rb
@@ -31,7 +31,10 @@ Puppet::Type.type(:rabbitmq_queue).provide(:rabbitmqadmin, parent: Puppet::Provi
         if arguments.nil?
           arguments = '{}'
         else
-          arguments = arguments.gsub(%r{^\[(.*)\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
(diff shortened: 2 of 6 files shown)


lib/puppet/provider/rabbitmq_queue/rabbitmqadmin.rb
```diff
@@ -31,7 +31,7 @@ def self.instances
         if arguments.nil?
           arguments = '{}'
         else
-          arguments = arguments.gsub(%r{^\[(.*)\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
+          arguments = arguments.gsub(%r{^\[|\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
```
wyardley, line 34: The unit test for this one defines a queue with no arguments, and then has the stubbed shell call return a second instance _with_ arguments

Maybe this could be adjusted slightly, to make sure that the arguments are parsed properly, for example, by adding a check similar to the one you added in the new unit test?
https://github.com/voxpupuli/puppet-rabbitmq/blob/83aa697583224d5761f922b5fe710d9faa3f7a46/spec/unit/puppet/provider/rabbitmq_queue/rabbitmqadmin_spec.rb#L23

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
```diff
@@ -37,10 +37,10 @@ def self.instances
         if arguments.nil?
           arguments = '{}'
         else
-          arguments = arguments.gsub(%r{^\[(.*)\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
+          arguments = arguments.gsub(%r{^\[|\]$}, '').gsub(%r{\{("(?:.|\\")*?"),}, '{\1:').gsub(%r{\},\{}, ',')
```
wyardley, line 40: Might be good to add a comment here explaining the logic, since this is pretty obtuse / hard to read... Where is `\1` in `{\1:` coming from now that the capturing parenthesis are gone? is it still needed?

wyardley (commented):
Thanks for this! Looks good overall. Have you tested this locally? Any corner cases with escaping or quoting or security risks? In the case where the hash changes, will that trigger any changes for existing users of the module that have existing bindings?

As with a lot of stuff in this module, the real fix would probably be to deprecate older versions, get some of the other needed changes across the line, and switch to either using JSON output from the CLI, or using the API directly, but this makes sense, I think.

Just thinking aloud, but I wonder if it would be feasible to JSON parse the entire arguments string vs. trying to use regex to parse it?

lib/puppet/provider/rabbitmq_binding/rabbitmqadmin.rb
wyardley: The second substitution has `?:`, so should be non-capturing, right? What is getting pulled into `\1` exactly, especially if it used to be coming from the first capturing parens that are now gone?

wyardley (approved):

bastelfreak (approved):

J-Mx:
Thank you for your review @wyardley. I have updated my branch to address your feedbacks.

> Thanks for this! Looks good overall. Have you tested this locally? Any corner cases with escaping or quoting or security risks? In the case where the hash changes, will that trigger any changes for existing users of the module that have existing bindings?

There are no particular edge cases or potential security risks with these changes. Including 'arguments' in the binding hash name calculation prevents perpetual binding modifications. Previously, bindings with identical source, destination, vhost and routing_key but different arguments would generate the same hash, causing Puppet to continuously try to update them thinking they were the same binding. This change ensures that bindings with different arguments are correctly identified as distinct resources.


> As with a lot of stuff in this module, the real fix would probably be to deprecate older versions, get some of the other needed changes across the line, and switch to either using JSON output from the CLI, or using the API directly, but this makes sense, I think.

As for parsing the arguments as JSON instead of using regex: it’s definitely possible, but it would require much more time and a significant refactor of the codebase compared to the fix proposed in this pull request. That said, it would likely help avoid these kinds of regex issues in the future.


> Just thinking aloud, but I wonder if it would be feasible to JSON parse the entire arguments string vs. trying to use regex to parse it?

The format returned by rabbitmqctl is not parsable JSON code, but rather follows Erlang's tuple list syntax (e.g., [{"key","value"}]). This is why substitutions with regular expressions are present in the code to transform this Erlang format into a JSON-parsable format.

PR #1058, opened 2025-07-25
puppet/archive Allow 8.x

TheMeier:
Ref: https://github.com/voxpupuli/community-triage/issues/44

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -63,7 +63,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 2.0.0 < 8.0.0"
+      "version_requirement": ">= 2.0.0 < 9.0.0"
     },
     {
       "name": "puppet/systemd",


wyardley (approved):
LGTM, not sure if we want to drop some version support in the lower end of these at some point too?

TheMeier:
raising min versions would be a breaking change

wyardley:
> raising min versions would be a breaking change

Yes, I know, that's why I said "at some time"; was not suggesting raising them as part of this.

wyardley:
Do you know if that CI failure is just a transient issue? I can try re-running once the rest of the run finishes.

TheMeier:
puppetlabs-apt is not compatible with puppet/openvox 7. We will drop it soon. In the meantime we could bin the fixture.

See https://github.com/search?q=org%3Avoxpupuli%20%22tag%3A%20v10.0.1%22&type=code for examples

TheMeier:
also removing openvox/puppet 7 is being worked on currently, so we could wait and rebase

PR #1060, opened 2025-08-05
Correctly handle rabbitmqctl list_users output

thomasgoirand:
The rabbitmq_user provider seemed broken to me, at least with a recent version of rabbitmq (I'm talking about 4.0.5 in Debian 13 aka Trixie).

This small patch fixes the provider.

### DIFF
--- a/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
+++ b/lib/puppet/provider/rabbitmq_user/rabbitmqctl.rb
@@ -18,16 +18,20 @@ Puppet::Type.type(:rabbitmq_user).provide(
       rabbitmqctl_list('users')
     end
 
-    user_list.split(%r{\n}).map do |line|
-      raise Puppet::Error, "Cannot parse invalid user line: #{line}" unless line =~ %r{^(\S+)\s+\[(.*?)\]$}
+    if user_list.split(%r{\n}).size > 0
+      user_list.split(%r{\n}).map do |line|
+        raise Puppet::Error, "Cannot parse invalid user line: #{line}" unless line =~ %r{^(\S+)\s+\[(.*?)\]$}
 
-      user = Regexp.last_match(1)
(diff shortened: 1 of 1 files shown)


wyardley:
https://github.com/voxpupuli/puppet-rabbitmq?tab=readme-ov-file#module-description

The module has been tested against 3.6, and has some minor fixes for later versions IIRC. I doubt it's been tested with 4.x, and does not currently have any active maintainers.

I would personally like to see something like #926, combined with dropping support for older versions, and making sure we're acceptance testing with a recent RMQ version, before we get too deep into supporting 4.x at all. (Once older version support is dropped, most of the providers could also use the json flag for the CLI, which would resolve a lot of the parsing issues that exist now).

We can happily look at small fixes that improve compatibility with newer versions of RMQ, but for sure we'd need additional spec tests added for how the new output looks. It might also make some more sense to key on the version, for example, like this:
https://github.com/voxpupuli/puppet-rabbitmq/blob/5216e37504a99ca5aec4c2c45d793fcf6630ce49/lib/puppet/provider/rabbitmq_policy/rabbitmqctl.rb#L24
or like this:
https://github.com/voxpupuli/puppet-rabbitmq/blob/5216e37504a99ca5aec4c2c45d793fcf6630ce49/lib/puppet/provider/rabbitmq_vhost/rabbitmqctl.rb#L19-L42

I would worry that even if you fix this issue, you may turn up other issues.

thomasgoirand:
Hi.

Well, I'm using this module in production under Debian 13 (aka: Trixie) with the patch I just sent plus a bit more (like puppetserver 8.4.0 support, ie: fixes to stop using toplevel facts that are now gone). Full list of patches here:

https://salsa.debian.org/openstack-team/puppet/puppet-module-puppetlabs-rabbitmq/-/tree/debian/epoxy/debian/patches?ref_type=heads

I can do the work to add all of my patches, but if you start refusing them it's going to be complicated...

As for the validation tests, I'm not sure how to fix them.

PR #1064, opened 2025-08-26
refactor: change from mocha to rspec-mocks #514

TheMeier:
Ref: https://github.com/voxpupuli/plumbing/issues/78


wyardley (approved):
Works for me. Should we update the rubocop config vs adding / updating more ignores?

TheMeier:
@wyardley the rubocop config is handled by [modulesync](https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.rubocop.yml.erb#L6) 
which pulls it from voxpupuli-test. 
I guess changing `RSpec/MessageSpies` and `RSpec/StubbedMock` should be discussed threre.

wyardley:
Sorry, yeah, that’s what I was getting at, since we manage the rubocop config, if we’ve moved away from it, we should probably adjust those cops in the modulesync config.

PR #1081, opened 2026-05-13
Switch to INI-style config; require rabbitmq 3.7+

### LINKED ISSUE #829: support for sysctl config file format
Hi, Would it be possible to add sysctl config file format as describe in the link below?
https://www.rabbitmq.com/configure.html#config-file

Thank you.

d1nuc0m:

Switch to the new INI-style config file.
Supersedes #1075.
Plan is to fully modernize module for 3.x with subsequents PRs and then add other templates etc for 4.x

~Problem is I'm not a RabbitMQ expert so I have some doubts:~
~* can I switch everything to the new format? For example LDAP options are still reported in advanced.conf in docs, but also in the new format, can I put everything in the ini-like file?~
~* What for options that are not in example confs, like ssl_cert_password?~
~* Ranch and Cowboy - found nothing in the docs, what should I do about that?~
~* Additional conf keys - should they go in the advanced file?~

Fixes #829


d1nuc0m:
Ready for review - @bastelfreak @corporate-gadfly could you review this one instead? @jay7x might be interested too

templates/advanced_3.config.epp
```diff
@@ -0,0 +1,110 @@
+[
+%% Erlang Kernel Application
```
bastelfreak, line 4: can you please add a header like `# THIS FILE IS MANAGED BY PUPPET`?

manifests/config.pp
```diff
@@ -171,7 +173,16 @@
     mode   => '2750',
   }
 
-  file { 'rabbitmq.config':
+  # file { 'rabbitmq.config':
```
bastelfreak: please don't commit commented out code.

bastelfreak:
@wyardley could you take a look here please?

templates/advanced_3.config.epp
d1nuc0m: Done

d1nuc0m:
EDIT - just realized that part of the module was still using the old EPP template due to duplicate data being in Hiera... Will have to update some things

manifests/config.pp
d1nuc0m: Will be replaced with options to purge old config files

d1nuc0m:
Ruby tests are now passing

manifests/init.pp
```diff
@@ -146,17 +146,25 @@
...
 #   Hash of Erlang kernel configuration variables to set (see [Variables Configurable in rabbitmq.config](#variables-configurable-in-rabbitmq.config)).
 # @param config_path
-#   The path to write the RabbitMQ configuration file to.
+#   The path to write the RabbitMQ configuration (in sysctl format) file to.
+# @param advanced_config_path
+#   The path to write the RabbitMQ advanced configuration (in Erlang format) file to.
+# @param purge_legacy_config_files
+#   Purge old config file (Erlang format).
+# @param legacy_config_path
+#   Path of the legacy config file path.
```
greatflyingsteve: Not a big deal, obviously, but this sentence both starts and ends with "path"

templates/advanced_3.config.epp
```diff
@@ -0,0 +1,119 @@
+%% THIS FILE IS MANAGED BY PUPPET
+%% DO NOT EDIT MANUALLY!
+[
+%% Erlang Kernel Application
+%% See https://www.erlang.org/doc/apps/kernel/kernel_app.html
+%%
+<% if $rabbitmq::config::config_kernel_variables {-%>,
+  {kernel, [
+    <%- $rabbitmq::config::config_kernel_variables.keys.sort.each |$k| { -%>
+    {<%= $k %>, <%= $rabbitmq::config::config_kernel_variables[$k] %>}<%- if $k != $rabbitmq::config::config_kernel_variables.keys.sort[-1] { -%>,<%- } %>
```
greatflyingsteve: This is incredibly awkward to read.  Remember, you can join things with strings that have newlines in them!  Consider using a map to walk through the Hash first, then sort the resulting Array, and join it with a string that has a newline.  Since the keys are at the head of each string that the map() outputs, you should still get the same sort order.
```suggestion
    <%= $rabbitmq::config::config_kernel_variables
        .map |$k, $v| { "{${k}, ${v}}" }
        .sort
        .join(",\n    ") %>
```
(Note that this suggestion is meant to replace all of lines 9-11, not just line 10)

manifests/init.pp
greatflyingsteve: Using stringified true/false is very legacy-flavored Puppet, and them ol' days weren't no _good_ ol' days.  Consider using real Boolean values, instead:
```suggestion
  Variant[Boolean, Enum['peer', 'best_effort']] $ssl_crl_check                                      = 'false',
```

manifests/init.pp
```diff
@@ -146,17 +146,25 @@
...
+#   Hash of additional configs (key / value) for `cowboy_opts` in advanced.config.
 # @param config_kernel_variables
 #   Hash of Erlang kernel configuration variables to set (see [Variables Configurable in rabbitmq.config](#variables-configurable-in-rabbitmq.config)).
 # @param config_path
-#   The path to write the RabbitMQ configuration file to.
+#   The path to write the RabbitMQ configuration (in sysctl format) file to.
+# @param advanced_config_path
+#   The path to write the RabbitMQ advanced configuration (in Erlang format) file to.
+# @param purge_legacy_config_files
+#   Purge old config file (Erlang format).
```
greatflyingsteve, line 165: Does this mean that the advanced config and the legacy config were both in Erlang format?

manifests/config.pp
```diff
@@ -252,7 +272,7 @@
       rabbitmq_group => $rabbitmq_group,
       rabbitmq_home  => $rabbitmq_home,
       service_name   => $service_name,
-      before         => File['rabbitmq.config'],
+      before         => File['rabbitmq.conf'],
```
greatflyingsteve: Consider whether this relationship is actually worth preserving.  Unless these files are re-read periodically, the relative order in which they're written is unlikely to be important.  When this relationship was originally put here, the rabbitmq.config file also had a `notify` to the service class, which it no longer has.

templates/advanced_3.config.epp
```diff
@@ -0,0 +1,119 @@
...
+<% if $rabbitmq::config::config_variables {-%>
+<%- $rabbitmq::config::config_variables.keys.sort.each |$key| {-%>
+    {<%= $key %>, <%= $rabbitmq::config::config_variables[$key] %>},
+<%- } -%>
+<%- } -%>
+
+%% Advanced RabbitMQ SSL Options
+<%- if $rabbitmq::config::ssl {-%>
+  {ssl_options, [
+    <%- if $rabbitmq::config::ssl_crl_check != 'false' {-%>
```
greatflyingsteve: If you use real booleans for this param, this can be simplified.  In the Puppet DSL, there are only two "falsey" values: `false` and `undef`.  You only need to explicitly check for `false` or `Undef` if the difference between them actually matters (like it would for, say, `$rabbitmq::config::quorum_membership_reconciliation_auto_remove` on line 44 below).  The other three values you allow for this param would all evaluate as `true`, and stringify cleanly with no additional work (Boolean `true` stringifies literally to `true` when interpolated into strings).
```suggestion
    <%- if $rabbitmq::config::ssl_crl_check {-%>
```

greatflyingsteve (changes requested):
It looks like there's a lot of cleanup here, which is awesome and good.  I think there's  only one question I want to ask, as someone who knows very little about RabbitMQ: would people still be using versions of RabbitMQ that don't support the new format yet?  Is it worth keeping the ability to create both (with the new one as default), then either let people switch or just use the correct format for the version installed?  Leaving the legacy file in place but forcing it go stale still seems a little heavy-handed if people would still be using this module with RabbitMQ versions that don't support the new one, but again, I don't know how likely that would be.

Also, though it's waaaaaay too late for it to matter for this PR, consider making discrete, related changes in smaller, more focused commits.  It makes reviewing easier, since you can step through the individual changes, and I find in my own work that it also helps me keep my focus on what I'm doing better, and I'm less likely to drop or forget something I needed to change like one giant commit does.  If you're worried about losing access to past changes if you find something you'd missed later on, you can always `git rebase -i HEAD~3`, for example, to go back and edit any one of the last three commits you've made.

templates/advanced_3.config.epp
greatflyingsteve: I didn't read all the way through every template on this first pass, but I did see this pattern used in a few other places; consider the map/sort/join approach for those, too.

README.md
```diff
@@ -57,6 +53,13 @@ Version v13.2.0 and older also added an erlang repository on CentOS 7. That isn'
 
 See [REFERENCE](REFERENCE.md).
 
+## Limitations
+
+Supported OSes and dependencies are given into metadata.json file.
+
+This module is tested with the last RabbitMQ 3.x release (3.13) and 3.x release from CentOS Messaging SIG (3.8).
+Priorly it was tested against 3.5.x and 3.6.x and known to not support all features against versions prior to 2.7.1.
```
greatflyingsteve: You might consider using "previously" instead of "priorly."

d1nuc0m:
> would people still be using versions of RabbitMQ that don't support the new format yet?

I don't think so, the new format has been introduced in [RabbitMQ 3.7.0](https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.7.0) in 2017, and 3.6 version support [has ended in 2018](https://endoflife.date/rabbitmq).

Minimum version I'd try to support is 3.8 because of CentOS Messaging SIG release.

> consider making discrete, related changes in smaller, more focused commits.

Usually I do, the issue in this case is doing a commit that groups a single change (moving to the new config file) instead of splitting it in dozens of related commits that leave the project in an "undefined" state, where nor the new nor the old format are fully supported

templates/advanced_3.config.epp
d1nuc0m: I just copied the content of the old template for this :D
I'll try to use map/sort/join, I'm still at the beginning with templating

manifests/init.pp
d1nuc0m: Exactly, quoting docs:

```
All supported RabbitMQ versions use an ini-like, sysctl configuration file format for the main configuration file. The file is typically named rabbitmq.conf.

The new config format is much simpler, easier for humans to read and machines to generate. It is also relatively limited compared to the classic config format used prior to RabbitMQ 3.7.0. For example, when configuring LDAP support, it may be necessary to use deeply nested data structures to express desired configuration.

To accommodate this need, modern RabbitMQ versions allow for both formats to be used at the same time in separate files: rabbitmq.conf uses the new style format and is recommended for most settings, and advanced.config covers more advanced settings that the ini-style configuration cannot express. 
```

[Link](https://www.rabbitmq.com/docs/configure#config-file-formats)

manifests/init.pp
d1nuc0m: ~~Can we postpone data types verification to a subsequent PR?~~ Done

manifests/config.pp
d1nuc0m: I think it should have a notify to the service, it should be reloaded in case of changes, I'll check

wyardley:
> I don't think so, the new format has been introduced in [RabbitMQ 3.7.0](https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.7.0) in 2017, and 3.6 version support [has ended in 2018](https://endoflife.date/rabbitmq).

Yeah, and the supported / tested in this module are pretty crusty so if they're working in this module's acceptance tests, I do think it's probably safe(ish).

It would be great if @d1nuc0m can confirm that they've tested this with newer RabbitMQ versions somewhat well /carefully, since there are a lot of gaps in testing still with newer versions, and would love to see something like #1075 or various earlier attempts at that ship as well. But overall, this seems good to me directionally.

templates/rabbitmq_3.conf.epp
```diff
@@ -0,0 +1,237 @@
...
+<%- } -%>
+<%- } -%>
+
+###
+### TLS Configuration
+###
+### Relevant doc guide: https://www.rabbitmq.com/docs/3.13/ssl.
+###
+<%- if $rabbitmq::config::ssl { -%>
+<%-   if $rabbitmq::config::ssl_interface {-%>
```
wyardley: IIRC it's possible to do the indenting outside the `<%-` tags vs. inside? IMO that would read easier

```suggestion
  <%- if $rabbitmq::config::ssl_interface {-%>
```

REFERENCE.md
```diff
@@ -421,15 +425,23 @@ Default value: `undef`
 
 Data type: `String`
 
-The file to use as the rabbitmq.config template.
+The file to use as the rabbitmq.conf template.
+
+Default value: `'rabbitmq/rabbitmq_3.conf.epp'`
```
wyardley, line 430: have we tested the behavior with applying the new config _on top of_ an instance run with the original / old config style? Even if this is flagged as a backwards incompatible change, I do think we should make sure that works properly.

wyardley (approved):
Overall, I'm Ok with this as best I understand, though I am not actively using RabbitMQ and haven't been for a while.

Few minor comments inline.

templates/rabbitmq_3.conf.epp
d1nuc0m: Yes, I'm too used to ERB where it's not possible, will fix

templates/advanced_3.config.epp
d1nuc0m: Added you as co-author too

manifests/config.pp
d1nuc0m: Wasn't necessary apparently

REFERENCE.md
d1nuc0m: It works when purging the old file or when the two have the same values, strange things can happen if the user also change the parameters, so there are conflicting values. I tried to clarify in the docs that it is a breaking change and users should test it.

d1nuc0m:
> It would be great if @d1nuc0m can confirm that they've tested this with newer RabbitMQ versions somewhat well /carefully

I tried it with my setup, but I guess it's not an extensive test. I'd rely more on a lot of warnings to users



> there are a lot of gaps in testing still with newer versions

I know, one of the goals is to also expand/refactor them



> would love to see something like #1075 or various earlier attempts at that ship as well

I'm the author of that PR too :D  
I'd like to bring in the necessary changes with further PRs and then get a 16.x release that supports the new RabbitMQ major (basically I have started this process because I want to upgrade to RabbitMQ 4.x in my setup)

d1nuc0m:
@bastelfreak @wyardley @greatflyingsteve could you review it again?

spec/acceptance/class_spec.rb
```diff
@@ -106,14 +106,12 @@ class { 'rabbitmq':
       it { is_expected.to be_listening.on('0.0.0.0').with('tcp') }
     end
 
-    describe port(15_672) do
+    describe port(15_672), :port15672 do
```
wyardley, line 109: nit: is the `, :port15672` significant? I noticed line 113 still doesn't have something like that? I'd lean towards leaving it out unless it's needed?

spec/acceptance/class_spec.rb
```diff
@@ -106,14 +106,12 @@ class { 'rabbitmq':
...
       it { is_expected.to be_listening.on('0.0.0.0').with('tcp') }
     end
 
     describe port(25_672) do
-      xit 'Is on 55672 instead on older rmq versions' do
-        is_expected.to be_listening.on('0.0.0.0').with('tcp')
-      end
+      it { is_expected.to be_listening.on('0.0.0.0').with('tcp') }
```
wyardley, line 114: thanks for unskipping these! 🙏

spec/classes/rabbitmq_spec.rb
```diff
@@ -632,20 +644,21 @@
...
-                           '    {servers, ["ldap.example.com"]},',
-                           '    {user_dn_pattern, "ou=users,dc=example,dc=com"},', '    {use_ssl, false},',
-                           '    {port, 389},', '    {foo, bar},', '    {log, true}',])
+          is_expected.to contain_file('rabbitmq.conf').with('content' => %r{auth_ldap.servers.1 = ldap.example.com})
+          is_expected.to contain_file('rabbitmq.conf').with('content' => %r{auth_ldap.user_dn_pattern = ou=users,dc=example,dc=com})
+          is_expected.to contain_file('rabbitmq.conf').with('content' => %r{auth_ldap.use_ssl = false})
+          is_expected.to contain_file('rabbitmq.conf').with('content' => %r{auth_ldap.port = 389})
+          is_expected.to contain_file('rabbitmq.conf').with('content' => %r{auth_ldap.log = true})
+          is_expected.to contain_file('advanced.config').with('content' => %r{\{other_bind, as_user\}})
+          is_expected.to contain_file('advanced.config').with('content' => %r{\{foo, bar\}})
```
wyardley, line 653: I'll just note here that while the old test here is much harder to read, and this is probably an improvement, there are some ways this might "match" some invalid configs? At the least, may want to anchor some of these regexes to line beginnings?

wyardley (commented):
Thanks!
The epp templates are much easier to read now IMO.

I think the REFERENCE.md needs to be regenerated vs. deleted, unless Vox has changed how they do things -- otherwise, overall LGTM; some minor non-blocking feedback.

Has the clustering stuff been tested? That's the other thing that may somewhat be under-tested in the acceptance testing? Similarly, has it been tested with _newer_ RMQ versions that are not tested by default in the acceptance tests, but are installed with the upstream (RMQ) repo vs. the distribution's package?

Only other suggestion is, if you don't already have the acceptance tests working locally, getting them working (including with `BEAKER_destroy=no`) is going to really make your future plans much easier to work with.... not just to be able to iterate locally on things when you are reworking the rmq version / install stuff, but also to be able to have a quick platform to do your own functional testing or iterative fixes on.

d1nuc0m:
> I think the REFERENCE.md needs to be regenerated vs. deleted

Human error, deleted it by mistake

> Has the clustering stuff been tested?

They pass the test as before, I don't have a cluster to test it

> That's the other thing that may somewhat be under-tested in the acceptance testing?

Probably yes



> Similarly, has it been tested with _newer_ RMQ versions that are not tested by default in the acceptance tests

No, because the module still points to the old repositories that contain up to RabbitMQ 3.13 - so tests run on 3.8 on CentOS and 3.13 on the rest. Will do another PR after this to create a test matrix that explicitly uses versions (and uses newer repos)



> if you don't already have the acceptance tests working locally, getting them working

Done, sometimes I still push to have a remote copy of the work

spec/classes/rabbitmq_spec.rb
d1nuc0m: Idk, as the file uses indentation I can't anchor to line beginning

spec/acceptance/class_spec.rb
d1nuc0m: Had to add it as otherwise test do not work as there is a duplicate-related error

spec/classes/rabbitmq_spec.rb
wyardley: Could match `^\s*` then maybe, which would not match if the string existed but was commented out or prefixed with something else; I’m fine with leaving it as is. We have seen some issues with tests like this matching but still having an invalid config, but may be less of an issue with the new config file format.

d1nuc0m: Just realized it was about rabbitmq.conf, that does _not_ have indentation (as it's ini-style), line beginning with ^ should be enough for it. Instead `^\s*` gives issues with advanced.config. Out of curiosity, is there a recommended way of writing tests to improve readability?

wyardley: Got it, yeah that's what I was thinking when I looked at this one quickly.
Mostly, I was saying that your style was _more_ readable than asserting the entire raw string, just potentially a little more open to issues related to ordering / indentation etc.
But overall, I think this is mostly in the right direction, just wanted to point that out.

d1nuc0m: Yes, sorry, I didn't explain myself well - tried to fix the possible issues with regex matches. Let me know if it can be made even more readable/there are other issues

d1nuc0m:
Ping @wyardley @greatflyingsteve

bastelfreak (approved):

PR #1088, opened 2026-09-18
chore: support puppet/systemd 10.x

d1nuc0m:

Sort and upgrade dependencies.

n/a.


wyardley (approved):
Thanks! I would frame the title it around what we’re updating (support Puppet 10 or whatever) vs the sorting thing if that’s the goal.

